Issue #26, opened 2026-04-07
Lowest-common-denominator problem with heterogeneous backends


negz:
### What problem are you facing?

ModelDeployment is backend-agnostic by design. The ML team says "deploy this model" without knowing whether it lands on KServe, Dynamo, or a SaaS backend. But backends are heterogeneous -- they don't all support the same features. Autoscaling (#5), model sources (HuggingFace vs S3 vs GCS), and scale-to-zero all work on some backends but not others.

This creates a lowest-common-denominator problem. If ModelDeployment can only express features that every backend supports, the API surface shrinks as Modelplane adds more backends. Features that are common but not universal (like concurrency-based autoscaling) can't be expressed on ModelDeployment without some way to handle backends that don't support them.

Modelplane already has implicit capability matching. The scheduler filters environments by backend match (a model with only a KServe serving profile can't run on a Dynamo IE) and by VRAM fit (a 70B model can't run on an IE with a single L4). But these checks are hardcoded in the scheduling logic, not part of an extensible model.

### How could Modelplane help solve your problem?

We've identified two possible approaches. We have low confidence in either at this point -- this issue is to track the problem and iterate on solutions.

**Approach 1: Discriminated unions + capability-based scheduling.** Modelplane already uses discriminated unions on InferenceEnvironment (`backend`: KServe vs Dynamo, cluster `source`: GKE vs Existing). The same pattern could extend to ModelDeployment for features with backend-specific semantics. Autoscaling (#5) is the motivating case:

```yaml
spec:
  scaling:
    signal: Concurrency           # Concurrency | RPS | Schedule | Latency
    concurrency:
      minReplicas: 1
      maxReplicas: 5
      target: 32
      utilization: 70
      scaleDownDelay: 900
```

The signal discriminator determines the config struct. The ML team declares intent ("scale on concurrency, target 32") without knowing how it's implemented. The scheduler filters environments using a lookup table of what each backend supports -- Modelplane already knows this because the placement function contains the composition logic. Backends that can't fulfill the request aren't scheduling candidates. No scaling block = fixed replicas, same as today.

[...]

PR #27, opened 2026-04-09
Add scaling support for Dynamo and KServe backends

### LINKED ISSUE #5: Support model server autoscaling

ModelPlacement currently composes an LLMInferenceService with a [fixed replica count](https://github.com/modelplaneai/modelplane/blob/bdb2ac892176142d4d16059aba64e422de1a84c4/functions/compose-model-placement/main.py). Without autoscaling, every deployed model holds its GPUs whether it's serving traffic or not. There's no burst handling — if traffic exceeds what the fixed replicas can serve, requests queue and latency degrades. And there's no scale-to-zero, so idle models are expensive.

For the KServe backend, autoscaling requires [KEDA and Prometheus](https://kserve.github.io/website/docs/model-serving/generative-inference/autoscaling) (or an OpenTelemetry collector). KServe creates KEDA ScaledObjects automatically when an InferenceService is annotated with `serving.kserve.io/autoscalerClass: "keda"`. Prometheus scrapes vLLM metrics like `vllm:num_requests_running` and KEDA scales based on a target threshold.

The infrastructure work is straightforward — install KEDA and Prometheus as additional Helm releases in the [KServeStack](https://github.com/modelplaneai/modelplane/blob/bdb2ac892176142d4d16059aba64e422de1a84c4/functions/compose-kserve-stack/main.py), and have the placement function add the right annotations and `autoScaling` block to the LLMInferenceService.

The harder question is where scaling intent lives in Modelplane's API. The ML team creating a ModelDeployment shouldn't need to think about replicas or instances — Modelplane deliberately abstracts away infrastructure concepts. But someone needs to express "this model should scale between 1 and 5 serving instances with a target concurrency of 2." Options include putting it on ClusterModel (platform team bundles it with model config), on InferenceEnvironment (environment-wide defaults), or on a policy resource. The design doc's [future work section](https://github.com/modelplaneai/modelplane/blob/bdb2ac892176142d4d16059aba64e422de1a84c4/design/design.md#L757-L759) describes typed policy resources (PlacementPolicy, ResourcePolicy, etc.) that could be a natural home for this — a ScalingPolicy or similar, namespace-scoped, following the LimitRange pattern. The right answer probably depends on how backend-specific the scaling semantics are — what KEDA + Prometheus does for KServe is completely different from how Dynamo or a managed API would scale.

negz:
## Approach

Fixes #5.
Relates to #26.

The scaling block on ModelDeployment uses a discriminated union on signal type, consistent with how InferenceEnvironment uses discriminated unions for backend and cluster config. Two signals are supported:

- **Fixed**: Static replica count, no autoscaling.
- **Concurrency**: KEDA-based autoscaling driven by in-flight request metrics from Envoy Gateway.

```yaml
spec:
  scaling:
    signal: Concurrency
    concurrency:
      minReplicas: 1
      maxReplicas: 5
      target: 32
      utilization: 70
      scaleDownDelay: 300
```

`target` is the number of concurrent requests each replica should handle. `utilization` is the percentage of the target at which scaling triggers -- with target 32 and utilization 70, KEDA scales up when per-replica concurrency exceeds 22 (32 * 70 / 100).

The ML team specifies scaling intent on the ModelDeployment. The scheduler filters environments whose backends don't support the requested signal (lookup table in the scheduling function). The placement function translates intent into backend-specific resources. See #26 for the broader discussion about how this capability model addresses the lowest-common-denominator problem with heterogeneous backends.

## Scaling metric

Both backends use Envoy Gateway for ingress. Envoy exposes `envoy_cluster_upstream_rq_active`, a gauge of in-flight upstream requests per route. This gives a consistent proxy-level concurrency signal regardless of backend or inference engine. KEDA queries this metric from Prometheus, which discovers and scrapes the Envoy proxy pod via `additionalScrapeConfigs` in kube-prometheus-stack.

The Envoy cluster name is derived from the HTTPRoute name on each backend:
- KServe: `httproute/default/{name}-kserve-route/rule/0` (created by the LLMInferenceService controller)
- Dynamo: `httproute/default/{name}/rule/0` (composed by the placement function)

## Backend infrastructure

Both backend composition functions install kube-prometheus-stack and KEDA. The Helm release configuration is shared via `lib/prometheus.py` and `lib/keda.py`. The Prometheus service name is pinned via `fullnameOverride` so the ScaledObject can reference it at a known address.

## Placement

For Concurrency scaling, the placement function composes a KEDA ScaledObject as a provider-kubernetes Object on the remote cluster. On KServe it targets the Deployment that LLMInferenceService creates. On Dynamo it targets the DynamoGraphDeploymentScalingAdapter. The KEDA threshold is `target * utilization / 100`.

For Fixed scaling, the placement function sets the replica count directly. No KEDA resources are composed.

## Known limitations

The UI shows the configured scaling range (e.g. "1-5 replicas") but not the actual current replica count. For Dynamo, the DGD status exposes per-service replica counts that could be surfaced. For KServe, the LLMInferenceService v0.16 status doesn't include replica counts, and the underlying resource type varies (Deployment for single-node, LeaderWorkerSet for multi-node), making observation fragile.


functions/compose-model-placement/main.py
```diff
@@ -233,6 +232,97 @@ def compose_kserve_llmis(self, gpus):
...
+        requests at the proxy level. KServe creates an HTTPRoute named
+        {llmis-name}-kserve-route, and Envoy uses that as the cluster
+        name in its metrics."""
+        llmis_name = naming.to_dns_label(self.xr.spec.modelRef.name)
+        scaling = self.xr.spec.scaling
+        concurrency = scaling.concurrency
+
+        target = concurrency.target or 1
+        utilization = concurrency.utilization or 70
+        threshold = str(max(1, target * utilization // 100))
```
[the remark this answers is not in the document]
negz: Will fix this at the API level in https://github.com/modelplaneai/modelplane/issues/28.

apis/modeldeployments/definition.yaml
```diff
@@ -67,6 +67,73 @@ spec:
...
+                        minimum: 1
+                        default: 1
+                        description: Number of replicas to run.
+                  concurrency:
+                    type: object
+                    description: >-
+                      Scale based on in-flight request concurrency per
+                      replica. Required when signal is Concurrency.
+                    properties:
+                      minReplicas:
```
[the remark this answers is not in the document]
negz: Will fix in https://github.com/modelplaneai/modelplane/issues/28

negz:
Creating an autoscaled MD:

Screenshot 2026-04-13 at 4 53 29 PM


Viewing an autoscaled MD (note the `1-5 replicas` annotation):

Screenshot 2026-04-13 at 4 55 15 PM

functions/compose-model-placement/main.py
```diff
@@ -233,6 +232,105 @@ def compose_kserve_llmis(self, gpus):
...
+        # Envoy names its upstream clusters after the HTTPRoute:
+        # httproute/{namespace}/{route-name}/rule/{index}. KServe's
+        # LLMInferenceService creates a route named {name}-kserve-route.
+        envoy_cluster = f"httproute/{metadata.NAMESPACE_REMOTE}/{llmis_name}-kserve-route/rule/0"
+
+        min_replicas = concurrency.minReplicas if concurrency.minReplicas is not None else 1
+        max_replicas = concurrency.maxReplicas if concurrency.maxReplicas is not None else 1
+        cooldown = concurrency.scaleDownDelay if concurrency.scaleDownDelay is not None else 300
+
+        if max_replicas < min_replicas:
```
dennis-upbound, line 262: is this / can this / should this be enforced also at the CRD schema level with a CEL admission rule?

functions/compose-model-deployment/scheduling.py
```diff
@@ -48,10 +67,11 @@ def schedule(
...
-        if (p.metadata.labels or {}).get(metadata.LABEL_KEY_DEPLOYMENT) == deployment.metadata.name:
-            existing_envs.add(p.spec.inferenceEnvironmentRef.name)
+    existing_envs = {
+        p.spec.inferenceEnvironmentRef.name
+        for p in all_placements
+        if (p.metadata.labels or {}).get(metadata.LABEL_KEY_DEPLOYMENT) == deployment.metadata.name
+    }
 
     candidates = []
     for env in envs:
```
dennis-upbound, line 77: do we want to collect and log any failures for envs epsecially in the case of 0 candidate envs where we want to surface failures like not supporting the requested scaling signal, so instead of the user seeing something like InsufficientCapacity / "0 of N matched" message as a GPU-capacity miss where a more specific condition reason (e.g., ScalingSignalUnsupported) and a warning message that names the signal — something like "0 of 1 environments matched: prod-gpu-east does not support scaling signal WVA" might help surface misconfigurations easier? or is that already validated elsewhere too?

dennis-upbound (approved):
shipit! a couple questions/nits

functions/compose-model-placement/main.py
negz: Yes definitely. I want to take a pass on CEL validation in general in a follow up PR. Noted this case here so I don't forget: https://github.com/modelplaneai/modelplane/issues/28#issuecomment-4247580440

functions/compose-model-deployment/scheduling.py
negz: It's not surfaced elsewhere but I agree it could be improved. Want to raise an issue with ideas?

The main tools we have are conditions and results (which appear as Kube events). We do need to make sure they're stable, and we can't emit events on every reconcile.

We could also look to the Kubernetes scheduler for inspiration. What does it tell you when it can't schedule a pod?

PR #37, opened 2026-04-16
Support multi-node inference for models exceeding single-node VRAM

### LINKED ISSUE #32: Support multi-node inference for models that exceed single-node VRAM

Modelplane's VRAM-based scheduler computes the number of GPUs a model needs by dividing declared VRAM by per-GPU VRAM ([`compute_gpus`](https://github.com/modelplaneai/modelplane/blob/6d7a1de5be1c63b365f37772676ffeb10f065c45/functions/compose-model-placement/main.py#L152-L162)), then requests all of them on a single pod. This works when the model fits on one node. It breaks when it doesn't.

A 405B parameter model at FP16 needs ~810 GiB of VRAM. An 8x H100 node has 640 GiB. The scheduler computes `ceil(810/80) = 11` GPUs, sets that on the pod, and Kubernetes fails to schedule it. There's no check for whether the computed GPU count exceeds a single node's capacity, no fallback to multi-node orchestration, and no way for the user to request it.

Both KServe and Dynamo support multi-node inference through several mechanisms:

1. **Tensor parallelism (TP) across nodes.** KServe supports this via LLMInferenceService + LeaderWorkerSet using `spec.parallelism.tensor` and `spec.worker`. Dynamo supports it via `DynamoGraphDeployment` with `multinode.nodeCount`, using Ray (vLLM), native dist-init (SGLang), or MPI (TensorRT-LLM).

2. **Pipeline parallelism (PP) across nodes.** Both backends support staging model layers across node groups. Can be combined with TP.

Modelplane installs LeaderWorkerSet on KServe clusters ([`compose_leader_worker_set`](https://github.com/modelplaneai/modelplane/blob/6d7a1de5be1c63b365f37772676ffeb10f065c45/functions/compose-kserve-backend/main.py#L379-L395)) but never uses it — the placement function composes a single-pod LLMInferenceService with no `worker` section.

Note that KServe's LLMInferenceService cross-node TP has a known bug ([kserve#4946](https://github.com/kserve/kserve/issues/4946)) and PP config presets are not yet merged ([kserve#5047](https://github.com/kserve/kserve/issues/5047)).

The placement function should be able to compose multi-node inference workloads when a model's VRAM requirements exceed what a single node can provide.

**Detection.** The scheduler already knows model VRAM and per-GPU VRAM. It could compare the computed GPU count against the node pool's per-node GPU count (available from the InferenceEnvironment's capacity or resolved config) and determine when multi-node is required.

**Composition.** For KServe, this means emitting an LLMInferenceService with a `spec.worker` section and appropriate `spec.parallelism` settings. For Dynamo, it means setting `multinode.nodeCount` on the DynamoGraphDeployment service.

dennis-upbound:
## Summary

Fixes #32.

Models requiring more VRAM than a single node's GPU capacity (e.g. Llama 405B at ~810GiB on 8×H100 nodes with 640GiB) could not be scheduled. The scheduler computed the correct total GPU count but the placement function requested all of them on a single pod, which Kubernetes rejected.

The fix threads per-node GPU count through the stack:

- **XRD**: new `countPerNode` field on `InferenceEnvironment.status.capacity.gpuPools[]`
- **IE composition**: populates `countPerNode` from each node pool's `acceleratorCount`
- **Placement function**: new `GpuAllocation` dataclass. When `total > per_node`:
  - KServe: emits `spec.parallelism.tensor` + `spec.worker` (LeaderWorkerSet)
  - Dynamo: emits `multinode.nodeCount` on the Worker service
- **Scheduler**: rejects pools without enough nodes for multi-node models

Backward compatible — old IEs without `countPerNode` fall back to single-node behavior.

## Why `countPerNode` on IE status (not derived from spec)

The per-node GPU count is available in the IE spec (`spec.kserve.cluster.gke.nodePools[].gpu.acceleratorCount`), but it's buried inside a discriminated union (backend × cluster source × GKE/Existing). The IE status `gpuPools` was specifically designed to flatten that complexity for consumers — the IE composition function traverses the union once, and the scheduler and placement function read a simple flat list.

Alternatives considered:

| Approach | Why not |
|---|---|
| **Derive from spec in the placement function** | Couples placement to the IE's discriminated union shape (backend × source × provider). Duplicates the pool-traversal logic that compose-inference-env already does. Fragile if the IE schema changes. |
| **Derive from `acceleratorType` lookup table** | Not deterministic — L4 can be 1, 2, 4, or 8 per node depending on machine type. Would be wrong for non-standard configs. |
| **Factor from existing `count` field** | Information is lost — `count=16` could be 8×2 nodes, 4×4 nodes, or 16×1 node. Can't reverse the multiplication. |

Adding `countPerNode` to status follows the same pattern as the existing `count` and `memory` fields: the IE composition function resolves it once from whichever spec shape applies, and consumers read a flat list.

## Test plan

- [x] All 23 composition tests pass (`nix run .#test-crossplane`)
- [x] `test-model-placement-kserve-multinode` — 810GiB model on 3×H100 KServe cluster. Asserts `parallelism.tensor: 11`, `worker.size: 1`, 8 GPUs per pod
- [x] `test-model-placement-dynamo-multinode` — same model on Dynamo. Asserts `multinode.nodeCount: 2`, 8 GPUs per pod
- [x] `test-model-deployment-insufficient-nodes` — 810GiB model on single-node H100. Asserts 0 placements
- [x] Existing single-node tests still pass (no regressions)
- [ ] Real cluster validation with a multi-node model (requires 2+ H100 nodes)


apis/inferenceenvironments/definition.yaml
```diff
@@ -413,6 +413,13 @@ spec:
                         count:
                           type: integer
                           description: Total GPUs in this pool.
```
negz: Can we drop `count` and always set `countPerNode`? I think we'd need to add `nodes`, which'd be derived from `nodes` / `maxNodes` in spec. I think this would simplify some of the scheduling logic a bit.

functions/compose-model-placement/main.py
```diff
@@ -64,6 +65,16 @@
...
 
 
+@dataclass
+class GpuAllocation:
+    """Result of compute_gpus() — describes how GPUs should be distributed."""
+
+    total: int  # Total GPUs needed (ceil(model_vram / gpu_vram))
+    per_node: int  # GPUs available per node in the selected pool
+    node_count: int  # Nodes needed (ceil(total / per_node))
+    multi_node: bool  # Whether total > per_node (multi-node required)
```
negz: Does this need to be stored, given it can be derived pretty easy? (Should it be a method?)

functions/compose-model-placement/main.py
```diff
@@ -150,16 +161,31 @@ def resolve_inputs(self):
...
             pool_memory = quantities.parse_quantity(pool.memory or "0Gi")
             if pool_memory > 0:
-                return max(
+                total = max(
                     1,
                     math.ceil(quantities.parse_quantity(self.model.spec.resources.vram) / pool_memory),
                 )
-        return 1
+                # countPerNode may be absent on old IE statuses that haven't
+                # been reconciled after the XRD upgrade.
```
negz: I don't think we need to consider this (backward compatibility) given this is unreleased software.

negz (commented):
I think this could be slightly simplified, but overall it looks great!

dennis-upbound:
Comments addressed and re-tested!

negz (approved):
Thanks @dennis-upbound!

PR #39, opened 2026-04-24
feat(apideps): switch usages to apiDependencies

haarchri:
use models for apiDependencies like Usage API from Crossplane CRDs


negz (approved):
Nice, thanks @haarchri!

PR #40, opened 2026-04-24
Add intent-based deployment design exploration

dennis-upbound:
Design doc exploring how ModelPlane could abstract engine configuration (TP/PP, quantization, KV cache, P/D split) behind user-stated intent — workload patterns and SLA targets — compiled into backend-specific config via inference profiles.

Key ideas:
- Intent is optional fields on ModelDeployment, not a new CRD
- Translation lives in the composition function (inference profiles), not in the schema — updatable independently
- Built-in profiles ship with the Configuration package; platform teams override via ConfigMap; commercial tier is hardware-validated, continuously-maintained profiles
- Profiles encode stable heuristics, not precise tuning — and can delegate to backend-native compilers (Dynamo DGDR) when available

Includes CDN evolution analogy, competitive positioning vs KServe/Dynamo, commercial model analysis with industry precedents (NIM, OpenShift, Databricks, Cast.ai), and sustainability analysis for profile maintenance.


design/intent-based-deployment.md
```diff
@@ -0,0 +1,216 @@
...
+
+### What no one does today
+
+**Workload intent** (user describes the pattern, system picks defaults) shows up partially at Together AI and Fireworks. **SLA intent** (user describes targets, system compiles config) shows up at Dynamo's DGDR. Nobody unifies both across multiple backends and engines.
+
+---
+
+## Part 2: How It Works
+
+Two new concepts. **`spec.intent`** — an optional block on ModelDeployment where users state workload pattern, SLA targets, and optimization priority instead of engine flags. **Inference profiles** — curated mappings from `{workload, priority, model tier}` to recommended engine configuration, shipped with the Configuration package and overridable via ConfigMap. The composition function resolves intent through profiles and emits backend-specific resources. Explicit user config always wins.
```
negz: > overridable via ConfigMap

I generally prefer structured configuration that's specific to Modelplane than a ConfigMap. This helps with RBAC (grant access to this specific type, not all ConfigMaps) and schema.

design/intent-based-deployment.md
```diff
@@ -0,0 +1,216 @@
...
+spec:
+  modelRef:
+    name: llama-3.1-70b-instruct
+  environments: 2
+  intent:                        # NEW — optional
+    workload: rag
+    sla:
+      ttft: 500ms
+      itl: 40ms
+    priority: latency
```
negz: What does the intent resolve to? Status fields? Does it mutate its own spec?

I think NVIDIA have it right here: intent should be a distinct CR that resolves to another CR containing the concrete, resolved spec.

design/intent-based-deployment.md
```diff
@@ -0,0 +1,216 @@
...
+
+```yaml
+status:
+  conditions:
+  - type: ProfileResolved
+    status: "True"
+    message: "Resolved: rag/latency/medium (2026.04)"
+  intent:
+    resolved:
+      engineArgs: ["--max-model-len=32768", "--quantization=fp8"]
```
negz: How does this interact with the engine flags currently on ClusterModel? Right now that's the only source of flags. Will we need to handle merging flags (and validating merged flag sets)?

design/intent-based-deployment.md
dennis-upbound: Good call. Updated — profile overrides are now an InferenceProfile CRD instead of a ConfigMap. Gets us schema validation, scoped RBAC and discoverability. The composition function reads them via required resources, same pattern as how the placement function reads InferenceEnvironments. Platform teams create InferenceProfile CRDs for their hardware; built-in profiles in lib/profiles.py are the fallback.

design/intent-based-deployment.md
dennis-upbound: You're right — updated to the DGDR → DGD pattern. Intent is now a separate ModelDeploymentRequest CRD that compiles into a ModelDeployment
                                    
  - Resolved config IS the ModelDeployment — visible via kubectl get md, not hidden in status
  - Flag merging happens at compile time in the Request's composition function, not at runtime in the placement function. The ModelDeployment output has no ambiguity.
  - Pin = delete the Request, keep the Deployment. No profileVersion machinery needed.                               
  - Break glass = create a ModelDeployment directly. No "explicit wins over intent" precedence rules.
                                                                                                                    
 
The CRD chain is now: ModelDeploymentRequest → ModelDeployment → ModelPlacement → LLMInferenceService/DGD. Users who want intent use the Request. Users who want control use ModelDeployment directly. Both first-class.

design/intent-based-deployment.md
dennis-upbound: Addressed by the two-CRD model. The Request's composition function merges intent-resolved args with ClusterModel serving profile args at compile time. Rule: serving profile args always win (they're the platform team's explicit choices). The resulting ModelDeployment has a complete, unambiguous spec — the placement function doesn't need to know about intent at all. It just sees a normal ModelDeployment.

dennis-upbound:
Refactored into a 2-CRD model

PR #41, opened 2026-04-24
Move packages to the modelplane registry org

negz:
The Crossplane configuration package was published under a personal org (`negzupboundio/modelplane-infra`). This PR moves it to a shared org (`modelplane/modelplane`) so that contributors can be granted access directly.

Three related changes:

**Registry org migration.** All package references in `upbound.yaml`, `docs/getting-started.md`, and the `crossplane-python-functions` skill are updated to `xpkg.upbound.io/modelplane/modelplane`.

**Remove committed pull secret.** The literal Upbound session credential in `docs/pull-secret.yaml` is replaced with instructions to generate one using `up ctp pull-secret create`. Contributors need access to the `modelplane` org on Upbound. The repos will become public at launch, at which point the pull secret and ImageConfig can be removed entirely.

**CI push on merge to main.** A `push-package` job builds and pushes the Configuration after all checks pass. The push-crossplane app now auto-generates a dev tag from git metadata (`v0.1.0-dev.<count>.g<hash>`) instead of requiring a manually bumped counter. Registry auth uses `docker/login-action` with a robot account because `up login --token` alone doesn't authenticate OCI push operations. Requires two repo secrets: `UP_ROBOT_ID` and `UP_ROBOT_TOKEN`.


dennis-upbound (approved):
lgtm

PR #44, opened 2026-04-28
Standardize on a single inference backend

### LINKED ISSUE #26: Lowest-common-denominator problem with heterogeneous backends

negz:
Closes #26.

Supporting multiple inference backends creates a lowest-common-denominator problem. ModelDeployment is backend-agnostic, but backends don't all support the same features. If the API can only express features that every backend supports, the surface shrinks as backends are added. Autoscaling (#5, #27) made this concrete: fixed replicas and concurrency-based scaling are portable, but richer strategies (SLA-driven, KV-cache-aware) are backend-specific.

This PR makes the inference backend an implementation detail. KServe is the current implementation, but users don't select or see it. Since it's not in the API, it can be replaced or evolved without breaking the user contract.

The InferenceEnvironment spec is flattened. Before:

```yaml
    spec:
      backend: KServe
      kserve:
        version: v0.16.0
        cluster:
          source: GKE
          gke: { ... }
```

After:

```yaml
    spec:
      cluster:
        source: GKE
        gke: { ... }
```

Serving profiles on ClusterModel and Model drop the backend discriminator. Before:

```yaml
    serving:
    - name: vllm-kserve
      backend: KServe
      engine:
        name: vLLM
        image: vllm/vllm-openai:v0.7.3
```

After:

```yaml
    serving:
    - name: vllm
      engine:
        name: vLLM
        image: vllm/vllm-openai:v0.7.3
```

The engine enum is restricted to [vLLM] until KServe ships multi-engine support (there's a draft PR adding SGLang via spec.runtime).

On the composition side, compose-inference-env always composes a KServeBackend (no dispatch), compose-model-placement loses the Dynamo codepath (DynamoGraphDeployment, Dynamo HTTPRoute, Dynamo KEDA ScaledObject), and the scheduler drops backend compatibility checks. lib/backends.py and compose-dynamo-backend are deleted.

The design doc is updated to describe the single-backend architecture as-is, with multi-backend covered in Alternatives Considered and external inference endpoints (BYO SaaS) noted as future work.


functions/compose-inference-env/main.py
```diff
@@ -26,11 +24,9 @@
...
 from .model.io.k8s.apimachinery.pkg.apis.meta import v1 as metav1
 
-# Maps backends to their XR kinds (for Usage resources).
-BACKEND_XR_KINDS = {
-    backends.KSERVE: "KServeBackend",
-    backends.DYNAMO: "DynamoBackend",
-}
+# KServe version installed on remote clusters. Hardcoded as an internal
+# implementation detail — users don't choose or see this.
+KSERVE_VERSION = "v0.16.0"
```
negz, line 29: Note to self: bump this to latest.

design/design.md
```diff
@@ -54,10 +50,35 @@ governance model on top of it.
...
+NVIDIA and AMD GPUs, and [llm-d][llmd] extends it with disaggregated serving.
+Even NVIDIA Dynamo is [adopting Gateway API Inference Extension][dynamo-gaie]
+rather than building a competing routing standard.
+
+Modelplane uses KServe as its inference stack. Users don't select or see the
+stack; it's an implementation detail. This lets Modelplane expose the full depth
+of KServe's capabilities without worrying about portability across stacks, and
+since the stack isn't in the API, it can be replaced or evolved without breaking
+the user contract. See [Alternatives: Multiple inference
+backends](#multiple-inference-backends) for the rationale.
```
negz: None of this is background. It doesn't explain why we're building Modelplane. It's a detail of the proposal.

design/design.md
```diff
@@ -188,14 +208,13 @@ spec:
   resources:
     vram: "140Gi"
 
-  # Serving profiles — how this model is served on each backend. Each profile
-  # is a complete, tested configuration for a specific (backend, engine) pair.
+  # Serving profiles — how this model is served. Each profile is a complete,
+  # tested configuration for a specific (engine, image, args) combination.
```
negz, line 186: Aren't engine and image the same thing?

design/design.md
```diff
@@ -747,9 +678,8 @@ keep others pinned. A shared profile upgrades every model simultaneously with
 no way to canary.
 
 Serving profiles on ClusterModel preserve the "complete deployable unit"
-property while handling multiple backends. Each profile is a tested
-combination of (model, backend, engine, image, args). The ML team still sees
-one ref — no engine concepts leak to consumers. See
+property. Each profile is a tested combination of (model, engine, image, args).
```
negz, line 679: Again, how are engine and image different?

design/design.md
```diff
@@ -795,14 +753,18 @@ placements automatically.
...
-OpenAI-compatible endpoints with cost tracking, rate limiting, and fallbacks.
-The InferenceGateway API's `backend` discriminator is designed for this
-evolution — adding `backend: LiteLLM` would mean ModelPlacement composes
-LiteLLM config entries instead of Gateway API HTTPRoutes. The routing
-concern stays in ModelPlacement either way.
+performance-aware routing: selecting environments based on GPU pricing, latency,
+throughput, and queue depth.
+
+**External inference endpoints.** Enterprises won't run all inference on their
+own infrastructure. SaaS providers like BaseTen, Together AI, and NVIDIA NIM are
```
negz: Is NVIDIA NIM a SaaS provider?

design/design.md
```diff
@@ -795,14 +753,18 @@ placements automatically.
...
+part of the picture. Rather than modeling SaaS providers as backends (which would
+reintroduce the multi-backend complexity we deliberately avoided), ModelDeployment
```
negz: From the POV of this design "reintroduce" doesn't make sense.

negz (commented):
Self-review of LLM assisted PR. Mostly fine but a few nits on the design.

dennis-upbound (approved):

dennis-upbound:
One thing I hit while testing multi-node on A100s: the GPU VRAM lookup table in `compose-inference-env` doesn't have `nvidia-tesla-a100` (that's what GCP calls the A100 40GB). The scheduler silently sets memory to `0Gi` and rejects everything with InsufficientCapacity — no error message explaining why.

I think the right fix is to let users declare VRAM on the node pool spec and use the hardcoded table as a fallback:

```yaml
nodePools:
  - name: gpu-a100
    role: GPU
    gpu:
      acceleratorType: nvidia-tesla-a100
      acceleratorCount: 8
      memory: "40Gi"    # new field — optional, falls back to lookup table
```

The composition prefers user-declared memory, falls back to the table, and emits a warning if neither resolves. This way unknown GPU types (new hardware, weird cloud names) don't silently break scheduling.

I've got this working on `dennis/kimi-k2.6` (branched off `back`) — XRD change to add `gpu.memory` on both existing and GKE node pool specs, and a `gpu_pool_memory()` helper in the composition function. Happy to open a PR if this looks right.

Also bumped KServe to v0.18.0 and fixed the multi-node worker spec (`spec.worker` is a flat PodSpec in v0.17+, no `size`/`template` wrapper). Needed both for Kimi K2.6 on 16x A100 40GB across 2 nodes.

PR #49, opened 2026-04-30
Require explicit per-GPU VRAM on InferenceEnvironment node pools

negz:
The `compose-inference-env` function derived per-GPU VRAM from a hardcoded lookup table keyed by `acceleratorType`. The table only covered seven GCP accelerator names, and any unknown type silently resolved to `"0Gi"`. The scheduler then rejected every environment with `InsufficientCapacity` without explaining why. Dennis hit this on #44 testing A100 40GB GPUs, which GCP calls `nvidia-tesla-a100` -- a string the table doesn't know.

Adding entries to the table would only delay the next instance of the same problem. The table is invisible to users writing YAML, has no way to signal which strings it knows, and would need ongoing maintenance as new hardware ships and as other clouds use different naming. Centralizing it in a ConfigMap has the same fundamental issue -- it's still a mapping that has to be maintained, and the function would need awkward plumbing to read it.

This PR drops the table and makes `memory` a required field on the `gpu` block of every node pool, on both `InferenceEnvironment.spec.cluster` (GKE and Existing) and the underlying `GKECluster.spec.nodePools`. The field is validated by the XRD, so missing it is rejected at admission time rather than producing 0Gi capacity at reconcile time.

`acceleratorType` keeps its existing role: it's still passed to GCP to provision the right hardware, and it's still surfaced on `status.capacity.gpuPools` as informational metadata. It's just no longer used to look up VRAM.

Before:

```yaml
nodePools:
- name: gpu
  role: GPU
  machineType: a3-highgpu-8g
  gpu:
    acceleratorType: nvidia-h100-80gb
    acceleratorCount: 8
```

After:

```yaml
nodePools:
- name: gpu
  role: GPU
  machineType: a3-highgpu-8g
  gpu:
    acceleratorType: nvidia-h100-80gb
    acceleratorCount: 8
    memory: 80Gi
```

This is a breaking change for anyone with existing `InferenceEnvironment` or `GKECluster` resources -- they'll need to add `memory` to each GPU node pool.


dennis-upbound:
Thanks!

Issue #56, opened 2026-05-04
Align hardware capabilities design with Kubernetes Dynamic Resource Allocation


bassam:
### What problem are you facing?

Today's InferenceEnvironment node pool spec mixes hardware capabilities and provisioning details in ways that make matching brittle and don't align with where Kubernetes is heading for accelerator management.

We've been discussing how to express hardware capabilities for matching against ClusterModel requirements (see #55). The core question is what vocabulary and structure to use. Designing this from scratch with our own enums (`interconnect: nvlink`, `precisionSupport: [fp8]`, etc.) means inventing a parallel taxonomy that we have to maintain ourselves, that diverges from vendor terminology, and that doesn't extend cleanly to multi-vendor scenarios.

There's likely a better answer available: Dynamic Resource Allocation (DRA) reached stable in Kubernetes 1.34 and is now the cornerstone of the Kubernetes AI Conformance program. Major drivers exist for NVIDIA (`gpu.nvidia.com`), AMD (`gpu.amd.com`), Google TPUs, and networking (DRANET). Vendor attribute schemas are stable and standardized within each vendor. Cross-vendor topology attributes (`resource.kubernetes.io/pcieRoot`, NUMA, etc.) are being standardized through KEP-4381, KEP-5316, and KEP-5491 by WG-Device-Management.

This design sketch explores adopting DRA conventions directly rather than invent parallel ones.

There's also a second concern. Node pool spec serves two purposes today, and they're conflated in a way that makes neither clean:

1. **Matching** — ClusterModel serving profiles match against pool capabilities. The deploy function decides which pool can host a deployment by comparing requirements to capabilities.
2. **Provisioning** — for Modelplane-provisioned environments, the node pool spec needs enough information to actually create the pool in the underlying CSP (GKE, EKS, etc.).

Cloud providers are evolving toward separating these concerns. GKE's ComputeClass is the provisioning-side abstraction; ResourceSlice (DRA) is the matching-side abstraction. EKS Karpenter's NodePool is similar to ComputeClass. We should follow the same separation rather than fight against it.

### How could Modelplane help solve your problem?

Restructure the InferenceEnvironment node pool spec to separate provisioning from matching, align matching attributes with DRA conventions, and structure the matching surface as three distinct levels — environment, node, and device.

#### Three levels of attributes

Different match concerns belong at different levels. Conflating them into a flat space loses the semantic structure that makes federation matching tractable.

[...]

PR #63, opened 2026-05-06
Federation scheduler + KServe renderer (managed-kai)

dennis-upbound:
## What this is

The federation scheduling algorithm + KServe rendering composition functions, scoped to **managed-kai** as the in-cluster scheduler. The plugin/dispatch system (Kueue, Volcano, none) and per-scheduler capacity adapters land in a follow-up MR — kept out of here so the algorithm + IR boundary are the focus of review.

API shape is owned by [#64](https://github.com/modelplaneai/modelplane/pull/64); this branch implements against that shape.

## Read order

1. [`design/proposed-modelplane-api/design.md`](design/proposed-modelplane-api/design.md) — architecture, what-lives-where, dependencies, scheduler properties, KAI integration, use-case traces.
2. [`functions/compose-model-deployment/scheduling.py`](functions/compose-model-deployment/scheduling.py) — federation scheduler. Plain Python, no Crossplane imports. `schedule(md, clusters, existing) → ScheduleResult`. Filter → Score → Bind.
3. [`functions/compose-model-deployment/main.py`](functions/compose-model-deployment/main.py) — composer. Required-resources → `scheduling.schedule()` → emit `ModelReplica` × `spec.replicas` + `ModelEndpoint` × `spec.replicas`.
4. [`functions/compose-model-placement/rendering.py`](functions/compose-model-placement/rendering.py) — pure builders for KServe LLM-IS, DRA `ResourceClaim`, KAI `PodGroup`.
5. [`functions/compose-model-placement/main.py`](functions/compose-model-placement/main.py) — renderer. `MR → LLMInferenceService + PodGroup + ResourceClaim(s)` on the target cluster.

## What's in vs out

**In:** federation scheduler (Filter → Score → Bind), `ModelReplica` IR, sticky placement, multi-replica capacity reservation, disaggregation (decode + prefill same-cluster), KServe v0.18 LLM-IS rendering, DRA selector CEL derivation, **managed-kai** wrap (schedulerName + PodGroup), 49 unit tests.

**Out (separate MRs):**
- Per-scheduler dispatch (Kueue/Volcano/none) + `IC.spec.scheduler.type` enum
- Per-scheduler capacity adapter controllers
- Cluster onboarding controller (auto-detect installed scheduler)
- Eviction controller (cluster-degraded re-placement)
- Per-version KServe adapters (v0.16/v0.17 dispatch)
- Real CEL evaluator (placeholder + monkeypatched in tests today)

## Tests

```bash
uv venv .venv-test
uv pip install --python .venv-test/bin/python pytest ruff pyright
.venv-test/bin/python -m pytest tests/unit -v
.venv-test/bin/ruff check functions/ lib/
```

49/49 tests pass in ~20ms. Ruff + pyright clean. Doesn't run end-to-end yet — adapters raise `NotImplementedError` until #64's protos are generated; algorithm and shape are testable in isolation.


design/scheduler-deliverables/examples/inferencecluster-prod-coreweave.yaml
```diff
@@ -0,0 +1,63 @@
...
+    mode: dra
+
+  attributes:
+    "modelplane.ai/region":       "us-east"
+    "modelplane.ai/provider":     "coreweave"
+    "modelplane.ai/tier":         "production"
+    "modelplane.ai/compliance":   ["soc2-type2", "iso-27001"]
+    "modelplane.ai/billingModel": "per-gpu-hour"
+
+  engine:
```
bassam: @dennis-upbound why is engine expressed here? I would have thought that InferenceCluster would not have engine or runtime concerns. essentially just hardware information and node pool configurations

design/proposed-modelplane-api/examples/clusters/byoc-coreweave-h200-dra.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - ngram-speculation
+      - multi-lora
+      - structured-output
+      - tool-calling
+      - chunked-prefill
+      - fp8-kv-cache
+      - mxfp4-quantization
+
+  nodePools:
+    - name: frontier-multinode
```
bassam, line 54: we'll likely need a node label to ensure that kubernetes schedules the pod to a node in this node pool.

design/scheduler-deliverables/examples/inferenceprovider-together.yaml
```diff
@@ -0,0 +1,54 @@
...
+        key: api-key
+
+  attributes:
+    "modelplane.ai/region":       "us-east"
+    "modelplane.ai/provider":     "togetherai"
+    "modelplane.ai/tier":         "production"
+    "modelplane.ai/billingModel": "per-token"
+    "modelplane.ai/networkAccess": "internet-routable"
+
+  engine:
```
bassam: @dennis-upbound whats the point of specifing these here if our concern is routing only to inference providers?

design/scheduler-deliverables/examples/inferencecluster-prod-coreweave.yaml
dennis-upbound: Agreed — engine on `InferenceCluster` mixes substrate with runtime, and locks each cluster to one engine. Pulled it out in 471fe0d.

Engine info now lives on the existing `KServeBackend` internal XR (`apis/kservebackends/`) — substrate / runtime split. The matcher reads features from the `KServeBackend` referenced by the cluster, not from the cluster itself. Multiple backends per cluster (vLLM + SGLang side-by-side) becomes natural.

Side benefit: `managed-kserve` vs `byo-kserve` is crisper — managed = Modelplane composes a `KServeBackend`; byo = operator points at an existing one.

design/proposed-modelplane-api/examples/clusters/byoc-coreweave-h200-dra.yaml
dennis-upbound: Real gap — fixed in 471fe0d. Added `nodeSelector` to `nodePools[]` and updated the example to set `modelplane.ai/pool: frontier-multinode`.

Convention: Modelplane-provisioned pools auto-apply the label; BYO pools require the operator to label their nodes themselves (or use cloud-provider native labels like `cloud.google.com/gke-nodepool`, `eks.amazonaws.com/nodegroup`). The composer uses this selector to constrain pods to the pool's nodes.

design/scheduler-deliverables/examples/inferenceprovider-together.yaml
dennis-upbound: Agreed — providers are opaque routes; declaring engine features pretends we know what's inside. Dropped `engine` from `InferenceProvider` entirely in 471fe0d.

Match is now `supportedModels[]` + env-level `attributes`. If the workload's model is in the catalog and the region/tier/compliance attributes match, it can route. Workload-side `requires.engineFeatures` becomes implicitly a cluster-only constraint (matched against the cluster's `KServeBackend`, never against providers). `matchTrace` will surface "skipped: provider doesn't expose engine features" for workloads that require them.

Documented this field-level matching semantics in the `ModelDeployment` XRD. Will defer renaming `requires.engineFeatures` → something cluster-explicit until Nic weighs in.

design/scheduler-deliverables/examples/gpt-oss-20b.yaml
```diff
@@ -0,0 +1,56 @@
...
+  engine:
+    name: vLLM
+    image: vllm/vllm-openai:v0.8.0
+    args:
+      - "--max-model-len=131072"
+      - "--gpu-memory-utilization=0.90"
+      - "--max-num-seqs=64"
+      - "--enable-prefix-caching"
+
+  environments: 2                         # fan out across two matched regions
```
bassam: I don't think we should support this. Users can create multiple deployments and route across them.

design/scheduler-deliverables/examples/gpt-oss-20b.yaml
```diff
@@ -0,0 +1,56 @@
...
+  environmentSelector:
+    matchLabels:
+      modelplane.ai/tier: production
+
+  scaling:
+    signal: Concurrency
+    concurrency:
+      minReplicas: 0                      # scale-to-zero
```
bassam: do we create a modeplacement per replica?

design/scheduler-deliverables/examples/inferenceprovider-together.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - structured-output
+      - tool-calling
+      - multi-lora
+      # Deliberately NOT declared (workloads requiring these are excluded
+      # from this provider, with field-level reasons in matchTrace):
+      #   - prefill-decode-disagg  (internal disagg topology not exposed)
+      #   - eagle-speculation / ngram-speculation (internal optimizations)
+      #   - fp8-kv-cache  (internal — user can't constrain KV precision)
+
+  supportedModels:
```
bassam: where is this used?

design/scheduler-deliverables/examples/kimi-k2.yaml
```diff
@@ -0,0 +1,77 @@
...
+    name: moonshotai/Kimi-K2-Instruct
+  source: HuggingFace
+  huggingFace:
+    repo: moonshotai/Kimi-K2-Instruct
+    revision: main
+    secretRef:
+      namespace: platform-system
+      name: hf-token
+
+  topology:
```
bassam: why are these not `nodeClaim` and `deviceClaim`. Similar to my sketch in #56.

design/scheduler-deliverables/examples/kimi-k2.yaml
```diff
@@ -0,0 +1,77 @@
...
+    devicesPerNode: 8
+    parallelism:
+      tensor: 8
+      pipeline: 2
+      expert: enabled
+    roles:                                # 5P3D disaggregation; roles inherit root topology
+      prefill: { replicas: 5 }
+      decode:  { replicas: 3 }
+
+  requires:
```
bassam: these should be DRA claim style, no? see #56

design/scheduler-deliverables/examples/assistant-endpoint.yaml
```diff
@@ -0,0 +1,32 @@
...
+#   - Routes select ModelPlacements by label; Modelplane's matcher controls
+#     which placements exist via the underlying ModelDeployments
+apiVersion: modelplane.ai/v1alpha1
+kind: ModelEndpoint
+metadata:
+  name: assistant
+  namespace: app-team
+spec:
+  routes:
+    - type: Placement
```
bassam: I think the route should go to all the placements of a modeldeployment -- i.e all its replicas on one or more inferenceclusters.

design/scheduler-deliverables/examples/gpt-oss-20b.yaml
```diff
@@ -0,0 +1,65 @@
...
+    requests:
+      - name: gpus
+        count: 1
+        perNode: 1
+        selector:
+          matchAttributes:
+            gpu.nvidia.com/architecture: [Ada, Hopper, Blackwell]
+            gpu.nvidia.com/memory: ">=48Gi"
+
+  requiredEngineFeatures:
```
bassam: I don't think this needed

design/scheduler-deliverables/examples/gpt-oss-20b.yaml
```diff
@@ -0,0 +1,65 @@
...
+  # No nodeClaim — single GPU, no inter-node requirement.
+
+  deviceClaim:
+    requests:
+      - name: gpus
+        count: 1
+        perNode: 1
+        selector:
+          matchAttributes:
+            gpu.nvidia.com/architecture: [Ada, Hopper, Blackwell]
```
bassam: This looks neat and powerful. Maybe we do leave DRA as break glass . Simple hardware string could be a convention as well

design/scheduler-deliverables/examples/gpt-oss-20b.yaml
```diff
@@ -0,0 +1,65 @@
...
+#   - For multi-region spread, create a second ModelDeployment targeting a
+#     different region and route across both via a ModelEndpoint (or compose
+#     them via a Crossplane XR).
+apiVersion: modelplane.ai/v1alpha1
+kind: ModelDeployment
+metadata:
+  name: gpt-oss-20b
+  namespace: app-team
+spec:
+  replicas: 0                             # scale-to-zero baseline; KEDA scales up on traffic
```
bassam: Is this different from the scaling below? Do we need both?

design/proposed-modelplane-api/examples/kimi-k2.yaml
```diff
@@ -0,0 +1,88 @@
...
+        selector:
+          matchAttributes:
+            gpu.nvidia.com/architecture: [Hopper, Blackwell]
+            gpu.nvidia.com/memory: ">=141Gi"
+            modelplane.ai/interNodeFabric: "ib-400g-or-better"  # node-level attr; uniform across devices
+        constraints:
+          - matchAttribute: gpu.nvidia.com/nvlinkDomain   # TP=8 in one NVLink domain
+            requests: [gpus]
+
+  requiredEngineFeatures:
```
bassam: Is still don't understand the tole of these. What do they match against?

design/proposed-modelplane-api/examples/kimi-k2.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - prefill-decode-disagg
+    - kv-cache-routing
+    - chunked-prefill
+    - fp8-kv-cache
+
+  parallelism:
```
bassam: These are now effectively modelplane features - something we need to ensure we support on kserve or any future backends

design/proposed-modelplane-api/examples/modelservice-together.yaml
```diff
@@ -0,0 +1,41 @@
...
+        key: api-key
+
+  attributes:
+    "modelplane.ai/region":        "us-east"
+    "modelplane.ai/provider":      "togetherai"
+    "modelplane.ai/tier":          "production"
+    "modelplane.ai/billingModel":  "per-token"
+    "modelplane.ai/networkAccess": "internet-routable"
+
+  supportedModels:
```
bassam: Is this just metadata? Is it used anywhere in Modelplane?

design/proposed-modelplane-api/examples/qwen3-coder.yaml
```diff
@@ -0,0 +1,97 @@
...
+  scaling:
+    signal: Concurrency
+    concurrency:
+      minReplicas: 1
+      maxReplicas: 6
+      target: 16
+      window: 60s
+      scaleDownDelay: 300s
+
+  adapters:                             # multi-LoRA + LoRA-aware routing
```
bassam: Where are these used?

design/scheduler-deliverables/examples/kimi-k2.yaml
dennis-upbound: Renamed to `clusterSelector` / `deviceSelector` — the *claim* framing implies allocation, which fits DRA at the runtime layer but not what we are doing pre-provisioning. The doc has a "Two-stage scheduling: federation vs in-cluster" section that lays this out: federation borrows DRA vocabulary (typed attributes, domain-prefixed keys, CEL, the `device.attributes[domain].name` access pattern) but evaluates against declared pool attributes; the backend adapter emits real `ResourceClaim`s only at the in-cluster grounding stage (mandatory for BYOC with DRA, optional for Modelplane-provisioned).

design/proposed-modelplane-api/examples/kimi-k2.yaml
dennis-upbound: Field dropped. The matcher derives the required-feature set from declared config:

- `roles.{prefill, decode}` present → `prefill-decode-disagg`
- `engine.optimizations.chunkedPrefill: true` → `chunked-prefill`
- `engine.optimizations.kvCacheRouting: true` → `kv-cache-routing`
- `engine.quantization.target` contains `kvCache` → `fp8-kv-cache`
- `parallelism.expert: enabled` → `expert-parallelism`
- non-empty `adapters[]` → `multi-lora`

It then unions the set and matches against each candidate cluster's `KServeBackend.spec.engine.features`. Single source of truth: user declares what they want, not which backend features that requires. The kimi-k2 example header now spells out this derivation; new typed `engine.optimizations` field carries the knobs that previously had no structural home.

design/proposed-modelplane-api/examples/kimi-k2.yaml
dennis-upbound: Right — parallelism (TP / PP / EP) is canonical at the Modelplane layer; the backend adapter translates to KServe LWS / Dynamo graph / vLLM args. The matcher excludes any backend that can't honor a requested combination (e.g. backends without expert-parallelism for MoE). Spelled this out in the `ModelDeployment` XRD comment and the kimi-k2 example header so the contract is explicit: every backend we ship promises to honor the parallelism dimensions we expose.

design/proposed-modelplane-api/examples/modelservice-together.yaml
dennis-upbound: Real consumer: `ModelEndpoint` route filtering. When a request targets model X, ME can only route to a `ModelService` whose `supportedModels` contains X — without this, ME and MS drift out of sync silently. Auto-refresh plan: a controller polls the SaaS catalog API (Together exposes `GET /v1/models`, OpenAI-compatible providers have the same shape). Documented in the `ModelService` XRD and the example.

Caveat that should be more prominent: `ModelService` itself is a sketch — Nic owns the dedicated-SaaS placement design (provisioning a Together / Baseten dedicated endpoint), and this CR is a routing-only placeholder pending that work. Promoted that framing in the doc TL;DR.

design/proposed-modelplane-api/examples/qwen3-coder.yaml
dennis-upbound: Two consumers, both real:

1. **Backend adapter** — passes adapters into the engine at startup (vLLM `--lora-modules name=path` for each entry); the engine loads them as servable variants alongside the base model.
2. **Gateway** — LoRA-aware request routing. Client specifies adapter via `model: qwen3-coder:<adapter-name>` (or a header); router picks a replica that has it loaded. Without this, LoRA load distribution is uneven and adapters get loaded-and-evicted by the engine's LRU.

Presence of a non-empty `adapters[]` also implies the `multi-lora` engine feature for the matcher's required-feature derivation. Comment block spelling both consumers out is now in the XRD and the qwen3-coder example.

design/proposed-modelplane-api/examples/endpoints/multi-region.yaml
```diff
@@ -0,0 +1,36 @@
...
+      deployment:
+        ref: { name: kimi-k2 }            # us-east-1 MD; covers all its placements
+    - type: Deployment
+      weight: 45
+      deployment:
+        ref: { name: kimi-k2-eu }         # eu-west-1 MD; covers all its placements
+    - type: InferenceProvider
+      weight: 5
+      inferenceProvider:
+        ref: { name: together-prod }      # SaaS spillover
```
negz: Why explicit refs? Wouldn't matching on labels be more flexible?

dennis-upbound: Good call — fixed in 1ca8a00.

The example now uses `selector.matchLabels` instead of `ref` across all three routes (the two Deployments and the InferenceProvider). MDs labelled `modelplane.ai/{model: kimi-k2, region: us-east-1|eu-west-1, environment: production}`; the InferenceProvider gets `role: spillover, environment: production`. Bumping or replacing an MD stays transparent to the ME as long as the new MD carries the same labels — that's the environment-promotion pattern you flagged earlier.

Two changes underneath:

1. **ModelEndpoint XRD** — added `selector.matchLabels` to InferenceProvider routes. Was `ref`-only; Deployment routes already had both. Symmetry across all three target types now.
2. **Top-of-file comment** on the XRD now calls out label-based selection as the natural pattern for environment promotion (stable URL across MD revisions).

Left `assistant.yaml` on `ref` — that one's a heterogeneous canary-style endpoint where you specifically want each named MD, and refs read better there. Both patterns are first-class on the schema.

Issue #48, opened 2026-04-30
Autoscale a model deployment


dennis-upbound:
### What problem are you facing?

A ModelDeployment runs at a fixed replica count. `spec.replicas` is whatever you set, and nothing moves it as load changes. The building blocks for autoscaling are already here: the deployment exposes the Kubernetes scale subresource, so `spec.replicas` can be driven from outside, and node pools autoscale and scale to zero through `minNodeCount` and `maxNodeCount`. What is missing is a way to tie a load signal to the replica count. Replica pods also run on remote workload clusters rather than next to the XR, so a metric-based HorizontalPodAutoscaler cannot observe them.

### How could Modelplane help solve your problem?

A model should be able to scale its replicas with demand. Rather than bake in one autoscaling policy, Modelplane should treat replica scaling as an extension point: keep the scale subresource as the supported way to drive `spec.replicas`, so an operator can point the autoscaler they already run at a ModelDeployment, for example KEDA against a concurrency or queue-depth metric, and let the node autoscaling underneath provision and reclaim GPUs as replicas come and go. The outcome is a model whose replica count follows load, driven by the operator's own signal, without Modelplane owning the policy.

negz:
@dennis-upbound can we consider this done? We do support the scale sub resource now iirc. Or do we want to keep it open until we have the metrics defined?

PR #64, opened 2026-05-07
Update design to reflect latest thinking

negz:
The team spent a lot of time on Modelplane and its API last week. The most significant change we've aligned on is to dial down the separation of concerns on deploying a model. Specifically we've dropped the model catalog and we're going to expose more knobs to the ML teams authoring a ModelDeployment.

We've made ModelDeployment much more expressive in order to make sure it could deploy a frontier (open weight) model. We've adopted a subset of DRA to express scheduling constraints. We've also decided to scale only at the replica level - i.e. a ModelDeployment can scale the number of ModelReplicas but there's no scaling within one ModelReplica.


design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+## Resource model
+
+| Resource | Scope | Created by | Purpose |
+|---|---|---|---|
+| `InferenceGateway` | Cluster | Platform team | Control plane routing infrastructure |
+| `InferenceClass` | Cluster | Platform team (or Modelplane defaults) | Named hardware topology bundle |
+| `InferenceCluster` | Cluster | Platform team | A cluster in the inference fleet |
+| `ModelDeployment` | Namespace | ML team | Self-contained model deployment spec |
+| `ModelPlacement` | Namespace | Modelplane (composed) | Per-cluster realization of a deployment replica |
+| `ModelService` | Namespace | ML team | Routing surface across deployments and external endpoints |
```
dennis-upbound: thanks for doing this. the service and endpoint naming make way more sense now that it aligns with k8s (and I called the managed SAAS thing an `InferenceProvider` aka the URL+auth+some meta for things like together etc)

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+    gpu.product: H200
+    gpu.architecture: Hopper
+    gpu.vramGiB: 141
+    gpu.count: 8
+    gpu.features: [fp8, bf16, transformer-engine, mig]
+    interconnect.intraNode: nvswitch
+    interconnect.intraNodeBandwidthGBs: 900
+    network.interNode: infiniband
+    network.interNodeBandwidthGbps: 400
+    driver.version: {type: version, value: "535.129.03"}
```
bassam: Is this needed here?

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+  cluster:
+    source: Existing
+    existing:
+      secretRef:
+        name: coreweave-kubeconfig
+        key: kubeconfig
+
+  nodePools:
+  - name: frontier
+    class: h200-nvl-8x-ib
```
bassam: If this cluster where to provision the node pools the inference class feels farther away now to match provisioning configuration with hardware capabilities. We should consider making the class optional. So either a class or capabilities.

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+      perNode: 8
+      cel: |
+        capabilities["gpu.vramGiB"] >= 141 &&
+        "fp8" in capabilities["gpu.features"] &&
+        capabilities["network.interNode"] == "infiniband" &&
+        capabilities["network.interNodeBandwidthGbps"] >= 400
+
+    parallelism:
+      tensor: 8
+      pipeline: 2
```
bassam: Does this imply that kserve will run two pods on different nodes?

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+  different hardware (H200 with PP=2, H100 with PP=4). The common case is a
+  single entry.
+- **Placement count, cluster selection, and scaling are deployment-level.**
+  `spec.placements` controls how many `ModelPlacement`s are created (one per
+  cluster). `spec.clusterSelector` filters the candidate clusters. Optional
+  `spec.scaling` configures per-placement pod autoscaling (KEDA targeting the
+  per-cluster `LLMInferenceService`). Omitting `spec.scaling` means fixed
+  replicas — one pod per placement. Profiles never carry these concerns — only
+  one profile is active per placement, and the deployment's intent doesn't
+  change between profiles.
```
bassam: I don't think we need to do per placement pod autoscaling. This is more complicated. Instead I think we should autoscale placements. Placement 1:1 replica in this model. This would ensure that each replica goes back through scheduling and could land on other clusters. Scheduler can get more optimal here. Regional restrictions still possible here with cluster labels.

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+      modelplane.ai/tier: production
+
+  serving:
+  - name: default
+    poolSelector:
+      count: 16
+      perNode: 8
+      cel: |
+        capabilities["gpu.vramGiB"] >= 141 &&
+        "fp8" in capabilities["gpu.features"] &&
```
bassam: The serving profiles have to be compatible with the model image. For example fp8 profile requires fp8 image but others use a standard image. 

We could move model image and maybe cluster selector for not the serving profile but now this kind feels like an array.

Alternatively we accept that serving profiles are just for hardware variations across cloud providers, but then again this is what we use cel and capability matching for. 

I'd suggest we start without serving profiles and figure out if we need them in the future

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+      name: vLLM
+      image: vllm/vllm-openai:v0.9.0
+      args:
+      - "--max-model-len=65536"
+      - "--gpu-memory-utilization=0.9"
+      - "--enable-auto-tool-choice"
+      - "--tool-call-parser=hermes"
+
+  - name: h100-fallback
+    poolSelector:
```
bassam: Im questioning if poolSelector is the right term here. We're selecting a node or instance right? Or is it multiple nodes being selected? 

If the intention here to actually select a pool that has at least two nodes ( count / pernode ) that matches the cel expression, shouldn't we specify:

nodes: 2
gpuCount: 8

Using count:16 implies that we're talking one thing that has 16 gpus. 

I wonder if the selectors here is for a replica which could be made up of one or more nodes and hardware requirements.

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+  composes one `ModelEndpoint` per `ModelPlacement`. The ML team can also
+  create `ModelEndpoint`s manually to point at external services (Together,
+  BaseTen, Bedrock). Both look the same to `ModelService` — `spec.url` and
+  `spec.api` describe the endpoint, `auth` is optional for endpoints that
+  need credentials.
+- **`ModelService` always uses `spec.endpoints`.** No separate path for the
+  simple case versus weighted routing. Single-entry list for one deployment,
+  multi-entry with weights for canary, A/B, or SaaS overflow. Each entry
+  selects `ModelEndpoint`s by label — Kubernetes-native, no special
+  endpointRef syntax.
```
bassam: Should we also consider a model class? Feels like the intention behind inference class applies to model class too.

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+spec:
+  description: "1x NVIDIA L4, 24 GiB GDDR6, PCIe"
+  capabilities:
+    gpu.vendor: nvidia
+    gpu.product: L4
+    gpu.architecture: Ada
+    gpu.vramGiB: 24
+    gpu.count: 1
+    gpu.features: [fp8, bf16, int8]
+    interconnect.intraNode: pcie
```
bassam: These have to match the actual hardware provisioning on each of the clouds. Picking the instance types and configuring it all. I wonder if the more useful concept here is a inferenceClusterClass that combines hardware provisioning with matching capabilities. 

Modelplane can ship with one of these for each infra provider and set nodepools for each of the common hardware configurations. The node counts might vary though.

design/api-update.md
```diff
@@ -0,0 +1,457 @@
...
+  - name: h200-preferred
+    poolSelector:
+      count: 16
+      perNode: 8
+      cel: |
+        capabilities["gpu.product"] == "H200" &&
+        "fp8" in capabilities["gpu.features"] &&
+        capabilities["network.interNode"] == "infiniband"
+
+    parallelism:
```
bassam: These are for kserve only right? I'm still thinking about supporting dynamo and other backends.

design/api-update.md
negz: Agree - I like the simplification of only scaling on one axis.

Worth noting this kinda matches how an LLMIS scales. When you have a model that needs an LWS, their scaler scales number of LWS.

design/api-update.md
negz: Yeah I agree. Good point that quantization is coupled to the model weights.

design/api-update.md
negz: No, I think it was just contrived to show what a version would look like in this schema dialect.

bassam:
I think we should drop the `source` discriminator and `huggingFace:` block on ModelDeployment. I don't think they represent how LLMs are actually packaged today and they get in the way of future caching features..

LLM serving has settled into three distinct factorings of `(runtime, weights, optional compiled engine)`:

**Pattern 1 — engine fetches weights at startup.** Generic engine image (vLLM, SGLang, TGI upstream); engine pulls weights from HuggingFace, S3, or NGC via its native mechanism. The deployment specifies `engine.image` and engine args; the engine handles the rest. This is the dominant pattern for vLLM/SGLang/TGI workloads.

**Pattern 2 — engine image includes weights.** NIM is the canonical case. Runtime, optimization metadata, and weights are baked into one OCI image. Pull, run, serve. No separate fetch step.

**Pattern 3 — runtime and artifacts stored separately.** Generic runtime image plus separately-stored artifacts: weights the platform fetches and stages, compiled TensorRT-LLM `.engine` files in object storage, weights in an internal registry, or weights pre-staged on a PVC. Runtime mounts the artifacts at a known path and reads from there.

For **pattern 1**, `source` is redundant with what the engine arg already says. If the user writes `engine.args: ["--model=meta-llama/Llama-3.1-70B-Instruct"]`, having a parallel `source: { huggingFace: { repo: meta-llama/Llama-3.1-70B-Instruct } }` field means the user has to keep two fields in sync, and Modelplane has to decide whether to download via storage initializer (and then have the engine read from a mount) or let the engine fetch directly. Either way, one of the two configurations is doing nothing.

For **pattern 2**, `source` is meaningless. The image is the source. There's nothing external to point at.

For **pattern 3**, `source` could point at the external artifacts and have Modelplane stage them. This works, but it's the case where I believe we want the primitive to be at fleet-level — a future separate `ModelCache` resource that names artifacts see #66. Putting source on ModelDeployment scopes the staging to a single deployment, which is exactly the wrong scope for the cases where staging matters.

I propose we drop `source` and the `huggingFace:` block from ModelDeployment for v0.1. Patterns 1 and 2 work entirely from `engine.image` and `engine.args`. Pattern 3 waits for `ModelCache` in v0.2, which is also the right layer for fleet-aware caching optimizations.

For private repo access (HF tokens, NGC API keys, S3 credentials), engines already accept env vars. Add an `env` field on `engine` mirroring standard PodSpec env, with `valueFrom: secretKeyRef` for pull tokens:

```yaml
spec:
  engine:
    name: vLLM
    version: "0.8.5"
    image: vllm/vllm-openai:v0.8.5
    imagePullSecrets:
    - name: nvcr-creds
    env:
    - name: HF_TOKEN
      valueFrom:
        secretKeyRef:
          name: hf-token
          key: token
    args:
    - "--model=moonshotai/Kimi-K2-Instruct"
    - "--tensor-parallel-size=8"
    - "--pipeline-parallel-size=2"
```

The Kimi K2 example becomes simpler — no `source:` block, no `huggingFace.secretRef`, just the engine config. Same for Mixtral, Qwen3-Coder, and the disagg example.

design/api-update.md
```diff
@@ -0,0 +1,735 @@
...
+
+  engine:
+    name: vLLM
+    image: vllm/vllm-openai:v0.8.5
+    # Engine args pass through opaquely to the engine container.
+    args:
+    - "--tensor-parallel-size=2"
+    - "--max-model-len=32768"
+    - "--gpu-memory-utilization=0.9"
+```
```
bassam: I think we need engine `version` and likely `imagepullSecrets` and `env` support here too.

design/api-update.md
negz: (Resolving as we decided not to in Slack discussion.)

design/api-update.md
negz: We addressed this by moving provisioning config into the class.

design/api-update.md
negz: Updated to `nodeSelector` (and added a topology block to describe node layout).

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+   InferenceClass attributes and capacity.
+3. **Derive physical shape** from `workers.topology`: nodes per worker, GPUs per
+   node.
+4. **Check capacity.** Does the pool have enough available nodes? Available =
+   `maxNodeCount` minus nodes consumed by existing ModelReplicas on that
+   cluster.
+
+Modelplane will support affinity and anti-affinity in a future version.
+
+DRA is the device binding mechanism on every InferenceCluster. The composition
```
dennis-upbound, line 655: does this mean BYOC requires DRA/a particular k8s version?

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+
+The fleet scheduler picks `(InferenceCluster, pool)` for each ModelReplica:
+
+1. **Filter clusters** by `clusterSelector.matchLabels` against InferenceCluster
+   labels.
+2. **Filter pools** by evaluating `nodeSelector.cel` against each pool's
+   InferenceClass attributes and capacity.
+3. **Derive physical shape** from `workers.topology`: nodes per worker, GPUs per
+   node.
+4. **Check capacity.** Does the pool have enough available nodes? Available =
```
dennis-upbound, line 649: for existing clusters, are we assuming ModelReplicas are the only workload?

design/design.md
```diff
@@ -1,818 +1,732 @@
...
-Platform teams with existing Kubernetes clusters can bring their own
-infrastructure instead of having Modelplane provision it:
+For provisioned clusters (source: GKE), the composition function reads
+`InferenceCluster.cluster.gke` for the project and region, and each pool's
+`InferenceClass.provisioning.gke` for the machine type and GPU config. It
+combines them to provision the GKE node pool. System pools (non-GPU, for
+running the inference stack) are provisioned automatically.
+
+For BYO clusters (source: Existing), a kubeconfig Secret provides access.
+Modelplane installs the inference stack on the cluster but doesn't provision
```
dennis-upbound: can we expand a little on what gets installed and some of the failure modes/compatibility versions here?

design/design.md
```diff
@@ -1,818 +1,732 @@
...
-   routes requests across placements.
+1. **End-to-end demo works.** A platform team can install Modelplane, create an
+   InferenceCluster, and an ML team can create a ModelDeployment and get a
+   working OpenAI-compatible endpoint serving a credibly complex model at scale
+   with concurrency-based autoscaling.
 
-3. **The abstraction layer is credible:** The XR APIs are clean enough that an
-   enterprise platform team can look at them and see how they'd extend to their
-   requirements. The resource model makes intuitive sense.
+2. **Frontier models work.** The same API can express deployments for Kimi K2
```
dennis-upbound, line 96: for frontier models to work - multi-node weights that are quite large - ~1TB of weights for Kimi that will get downloaded independently for each pod (often too large to even download). We probably need ModelCache (#66) in v0.1 scope if we want to land frontier model demos well

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+
+The top-level `nodeSelector`, `workers`, and `engine` are always the decode (or
+unified) settings. Adding a `prefill` block makes the deployment disaggregated.
+The `prefill` block is self-contained: it repeats all settings rather than
+inheriting from the root, because explicit repetition is easier to reason about
+than implicit merge.
+
+The P:D ratio is expressed via `workers.count` on each role. It's a topology
+parameter (fixed per deployment), not a scaling knob. Decode and prefill must
+land on the same InferenceCluster (KV cache transfer needs co-location) but can
```
dennis-upbound, line 438: Co-location handles the KV-cache-transfer requirement but not the admission requirement. A disagg ModelReplica is non-functional unless both decode and prefill worker sets are up — decode alone has nothing to consume, prefill alone has nothing to feed. They need atomic admission, not just co-located scheduling.

How does this happen?

- With llm-d / native LWS: each role is its own LWS group. LWS gangs admit atomically within a group, but nothing coordinates between the two groups by default. Decode could come up while prefill is still pending node capacity.
- With Dynamo: does DynamoGraphDeployment admit both worker sets atomically? Worth confirming with NVIDIA.
- With KAI scheduler (#68): we'd compose a PodGroup with `minMember = sum(prefill_pods + decode_pods)`. Gang admission across roles.
- With Kueue: workload-granularity. One workload per role could admit independently → degraded "decode up, prefill pending" window.

The doc should at least call out the mechanism for v0.1, even if the answer is "rely on the backend orchestrator to handle it and verify per-backend."

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+
+#### Disaggregated prefill/decode
+
+The top-level `nodeSelector`, `workers`, and `engine` are always the decode (or
+unified) settings. Adding a `prefill` block makes the deployment disaggregated.
+The `prefill` block is self-contained: it repeats all settings rather than
+inheriting from the root, because explicit repetition is easier to reason about
+than implicit merge.
+
+The P:D ratio is expressed via `workers.count` on each role. It's a topology
```
dennis-upbound, line 436: Worth clarifying the behavior of `workers.count` here vs. `spec.replicas` for the disagg case. As I read it:

- `workers.count` inside a role block (decode or prefill) is the P:D-ratio knob: 3 decode workers + 5 prefill workers per replica, all sharing KV transport, all admitted together as one logical serving instance.
- `spec.replicas` adds whole copies of that instance — replica 2 is its own complete (5P, 3D) set, scheduled independently by the matcher, with its own KV transport between its own workers.

So a deployment with `replicas: 2` + `decode.workers.count: 3` + `prefill.workers.count: 5` ends up with 16 workers total (2 × (3+5)) in two independent disagg groups. Worth saying explicitly somewhere in this block — the doc currently has both knobs but doesn't pin down what each scales independently of the other.

design/design.md
dennis-upbound, line 655: For a multi-pod LWS gang or a disagg replica, what's the ResourceClaim shape? DRA claims are tied to pod containers via the claim name in `spec.resourceClaims`. For an LWS gang of 2 pods each needing 8 GPUs, options are:

- 16 individual claims (one per GPU)
- 2 claims of count=8 (one per pod)
- 1 claim of count=16 with cross-pod allocation (DRA supports this for some drivers)

The choice affects how the DRA driver groups devices for NVLink-domain co-location (TP=8 wants all 8 GPUs on the same NVLink fabric per pod). For disagg, decode and prefill pools are different InferenceClasses → separate claims per role.

Worth a paragraph on the claim shape for each topology strategy (Tensor / TensorPipeline gang / DataExpert / disagg). This is where the matcher's CEL becomes a concrete DRA selector and the composition decision becomes load-bearing for scheduling correctness.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+      modelplane.ai/tier: production
+  replicas: 2
+  nodeSelector:
+    cel: |
+      capacity["gpu.nvidia.com/memory"].compareTo(quantity("80Gi")) >= 0
+  workers:
+    topology:
+      strategy: Tensor
+      tensor: 2
+  engine:
```
dennis-upbound: Three pod-spec gaps the engine block currently can't express, all of which block real v0.1 demos:

- `shmSize`: vLLM with TP>1 needs GB-scale `/dev/shm` for multi-process IPC; container default is 64MB → silent OOM.
- `imagePullSecrets`: NIM ships from NGC behind creds; customer mirrors are a workaround but not a baseline.
- `env` / `envFrom`: NCCL tuning, VLLM tuning, runtime env vars that aren't pass-through args.

Weight/artifact auth (HF tokens, S3 creds for engines, custom chat templates) is largely solvable via ModelCache #66 — commented there with the scope boundary. These three are what ModelCache can't address.

Could land as typed fields (`engine.shmSize` / `engine.imagePullSecrets` / `engine.env`) or a single `engine.podSpec` escape hatch covering all three.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+    - "--model=meta-llama/Llama-3.1-405B-Instruct"
+    - "--max-model-len=131072"
+    - '--kv-transfer-config={"kv_role":"kv_consumer"}'
+
+  # Prefill: 5 workers, each single-GPU. Self-contained.
+  prefill:
+    nodeSelector:
+      cel: |
+        capacity["gpu.nvidia.com/memory"].compareTo(quantity("80Gi")) >= 0 &&
+        attributes["modelplane.ai/networkInterNode"].string == "infiniband"
```
dennis-upbound, line 477: Each role's `nodeSelector.cel` independently checks `networkInterNode == "infiniband"`. Matcher verifies decode pool has IB and prefill pool has IB, but doesn't verify they're on the same IB fabric.

A cluster with two separate IB networks (one wired to each pool) passes both CELs but can't actually KV-transfer between them. Silent runtime failure — pods come up Ready, KV transport never establishes.

Either a `modelplane.ai/networkFabric` attribute that matches between physically-connected pools (matcher cross-checks decode/prefill fabric IDs), or an explicit "same fabric" constraint at the API level. Real concern for multi-pool clusters where pools were provisioned with separate networking switches.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+      tensor: 8
+      pipeline: 2
+  engine:
+    name: vLLM
+    image: vllm/vllm-openai:v0.9.1
+    args:
+    - "--model=meta-llama/Llama-3.1-405B-Instruct"
+    - "--max-model-len=131072"
+    - '--kv-transfer-config={"kv_role":"kv_consumer"}'
+
```
dennis-upbound, line 471: Passing `--kv-transfer-config` to an engine that doesn't support disagg (older vLLM, plain TGI, some NIM versions) silently no-ops. Engine starts, pod Ready, inference returns wrong results (no KV transferred → decode generates from a stale-or-empty cache).

Not asking for admission-time gating — engine-feature awareness in the matcher is complexity for marginal upside, and `engine.args` is opaque pass-through by design. Asking: the composer should wire the engine so mis-config surfaces as `MR.status.Ready=False` with reason, not silently-broken `Ready=True`.

Options: engine fail-fast flag if available, a post-startup readiness probe that verifies the active feature set (`/v1/info` includes `kv_transfer: active`), or just rely on the engine to crash and propagate the crash reason. Same concern applies for multi-LoRA, speculative decoding, etc — goal is fail-loud at startup, not fail-silent at serve.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+| `DataExpert` | `tensor`, `data`, `dataLocal` | `data / dataLocal` | `dataLocal * tensor` | `data * tensor` |
+
+The scheduler derives the physical shape from the topology. No separate node
+count or GPU count fields; the topology fully determines the resource
+requirements. The scheduler checks: does the matched pool's InferenceClass have
+`gpu.count` >= GPUs-per-node, and does the pool have enough available nodes?
+
+#### Disaggregated prefill/decode
+
+The top-level `nodeSelector`, `workers`, and `engine` are always the decode (or
```
dennis-upbound: The disagg shape is decode-required-with-optional-prefill. The inverse — prefill-only — is a real future case:

- Pre-compute system-prompt KV once per cluster, share across decode replicas
- Batch RAG context warming for known retrieved fragments
- Cache warming before traffic ramps

Not expressible today (top-level is decode, prefill is sub-block) and shouldn't be in v0.1 — it requires a fleet-level KV cache layer (LMCache federation, Mooncake, etc.) as the destination for prefill output. Without that destination there's nowhere for a prefill-only worker to write.

No easy fix without the cache primitive. Worth a sentence in the doc that prefill-only lands when fleet KV cache federation lands (v0.3+), so reviewers see it on the roadmap.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+describes the shape of one worker; `workers.count` (default 1) says how many
+workers of that shape exist per ModelReplica.
+
+`topology.strategy` is a required discriminator:
+
+| Strategy | Required fields | Nodes per worker | GPUs per node | Total GPUs per worker |
+|---|---|---|---|---|
+| `Tensor` | `tensor` | 1 | `tensor` | `tensor` |
+| `TensorPipeline` | `tensor`, `pipeline` | `pipeline` | `tensor` | `tensor * pipeline` |
+| `DataExpert` | `tensor`, `data`, `dataLocal` | `data / dataLocal` | `dataLocal * tensor` | `data * tensor` |
```
dennis-upbound: The strategy enum is pairwise: TP-only, TP+PP, or DP+EP. Frontier MoE configs being deployed in production combine three axes simultaneously:

- DeepSeek V3 / V3.1: TP=2 + EP=64 + PP=2
- Llama 4 Maverick: TP+PP+EP combinations
- Qwen2.5-MoE-72B: TP=8 + EP=8

Not in v0.1 scope (named targets are dense or pairwise MoE), but customers serving DeepSeek-class models will hit this in v0.2+. Worth a sentence on the evolution path:

- A fourth strategy (`TensorPipelineExpert`) — keeps the discriminated-union shape but doesn't scale: each new combination is a new row.
- A more compositional shape where `workers.topology` carries `{tensor, pipeline, expert, dataLocal}` independently and the physical shape is derived from whichever are non-default.

Either's fine; worth committing to one so the table doesn't lock us out.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+
+The scheduler derives the physical shape from the topology. No separate node
+count or GPU count fields; the topology fully determines the resource
+requirements. The scheduler checks: does the matched pool's InferenceClass have
+`gpu.count` >= GPUs-per-node, and does the pool have enough available nodes?
```
dennis-upbound: The matcher checks `pool's InferenceClass.gpu.count >= GPUs-per-node` and `pool has enough available nodes` — treats all nodes within a pool as uniformly interconnected.

For GB200 NVL72 (72 GPUs across 18 nodes in one NVLink domain), a TP+PP gang wants to land inside one NVL72 rack — not spread across racks where you fall back from NVLink to IB. The current shape can't express this; the matcher could happily pick 4 nodes from a pool that's actually split across two NVL72 racks, and TP performance silently drops.

GB200 is generally available now (CoreWeave et al). Options:

- One NVL72 = one pool (with `gpu.count: 72`). Falls out if operators declare it that way; no API change.
- A `modelplane.ai/scaleUnit: nvl72` capability attribute plus a matcher constraint that gang workers come from the same scale unit.
- Separate pool-level subdivision concept.

Doesn't need v0.1, but worth a sentence on how the matcher will handle scale units when GB200 customers show up.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+`workers` groups the worker count and compute topology. `workers.topology`
+describes the shape of one worker; `workers.count` (default 1) says how many
+workers of that shape exist per ModelReplica.
+
+`topology.strategy` is a required discriminator:
+
+| Strategy | Required fields | Nodes per worker | GPUs per node | Total GPUs per worker |
+|---|---|---|---|---|
+| `Tensor` | `tensor` | 1 | `tensor` | `tensor` |
+| `TensorPipeline` | `tensor`, `pipeline` | `pipeline` | `tensor` | `tensor * pipeline` |
```
dennis-upbound: `TensorPipeline` per this row says nodes-per-worker = `pipeline`, so `pipeline=1` means single-node. Pure cross-node TP with no pipeline stage — e.g. TP=16 across 2 nodes of 8 GPUs each — isn't expressible.

Niche but real: some Llama-70B-on-2-nodes setups want TP=16 for lower latency rather than TP=8+PP=2 (which adds pipeline-stage overhead).

Cheap fix: allow `TensorPipeline pipeline=1, tensor=16` to mean "TP=16 spread across `ceil(tensor / pool.gpu.count)` nodes" — derive node count from tensor + pool's gpu.count. Or document explicitly that the inverse case isn't supported in v0.1 so customers know to use PP instead.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
+      string: "NVIDIA H200 141GB HBM3e"
+    gpu.nvidia.com/cudaComputeCapability:
+      version: "9.0.0"
+    gpu.nvidia.com/gpuCount:
+      int: 8
+    gpu.nvidia.com/features:
+      strings: [fp8, bf16, transformer-engine, mig]
+    modelplane.ai/interconnectIntraNode:
+      string: nvswitch
+    modelplane.ai/networkInterNode:
```
dennis-upbound, line 245: The `modelplane.ai/*` networking attributes are Modelplane-owned (vs `gpu.nvidia.com/*` which the DRA driver publishes), so we control the vocabulary and customers need to know what values to use.

Realistic values across clouds:

- `modelplane.ai/networkInterNode`: `infiniband` | `gpudirect-tcpx` | `gpudirect-tcpxo` | `nccl-fast-socket` | `efa` | `efav3` | `roce`
- `modelplane.ai/interconnectIntraNode`: `nvswitch` | `nvlink` | `pcie` | `infinity-fabric`

Without a documented list, operators across clouds will invent different strings (`infiniband-ndr` vs `ib` vs `InfiniBand`) and matching CEL silently fails to match. Either an OpenAPI enum on these specific keys or just a documented convention in this section.

design/design.md
```diff
@@ -1,818 +1,732 @@
...
   labels:
-    modelplane.ai/deployment: llama-70b-global
+    modelplane.ai/deployment: kimi-k2
 spec:
-  # What to deploy
-  modelRef:
-    kind: ClusterModel
-    name: llama-3.1-70b-instruct
+  url: http://10.0.1.50/ml-team/kimi-k2/
+  api: OpenAI
```
dennis-upbound: `api: OpenAI` implies an enum. v0.1 OpenAI-only is fine, but worth flagging the extension path so future additions don't break compat — and the `ModelService` composition function will eventually branch on it (Anthropic format isn't `messages` shape, embedding endpoints don't have `/chat/completions`, etc.).

Future values worth reserving syntax for:

- `Anthropic` (Claude API format)
- `Embedding` (`/v1/embeddings`-only, no completions)
- `RawCompletion` (legacy `/v1/completions`)
- `Gemini` (Google format)

Worth declaring it explicitly as an OpenAPI enum with `OpenAI` as the only v0.1 value.

design/design.md
dennis-upbound: `ComputeDomain` is the upstream primitive for this. The NVIDIA DRA driver for GPUs publishes a `ComputeDomain` CRD naming a single MNNVL/NVL72 fabric; workloads reference it and the driver only places pods on nodes belonging to that ComputeDomain. We don't need to invent fabric topology in InferenceClass — surface a `modelplane.ai/computeDomain` attribute (or similar) so the matcher knows a pool participates, then have the renderer thread the ComputeDomain ref through the ResourceClaim for multi-node MNNVL workloads. GPU Operator v25.10+ installs the driver.

design/design.md
dennis-upbound: KEP-5729 (DRA: ResourceClaim Support for Workloads) is the upstream direction here — associate ResourceClaims with Workload/PodGroup objects so one claim is shared across the gang's pods, getting around the 256-entry `reservedFor` limit. Still in design upstream (drafted Dec 2025, combined with KEP-5194 in Jan); no K8s alpha yet. Per-pod-role differences in claim shape (leader=8 GPU vs worker=4 GPU in one gang) aren't directly addressed by 5729 — that stays separate-claim-per-shape. For v0.1 we emit one ResourceClaim per role (decode, prefill); revisit when 5729 lands so naming maps cleanly.

design/design.md
```diff
@@ -1,818 +1,731 @@
...
+      count: 5
+      topology:
+        strategy: Tensor
+        tensor: 1
+    engine:
+      name: vLLM
+      image: vllm/vllm-openai:v0.9.1
+      args:
+      - "--model=meta-llama/Llama-3.1-405B-Instruct"
+      - '--kv-transfer-config={"kv_role":"kv_producer"}'
```
dennis-upbound: vLLM's NixlConnector — the transport behind both llm-d disagg and Dynamo — requires `VLLM_NIXL_SIDE_CHANNEL_HOST` set to the pod's own IP via downward API, on every prefill pod. Without it, prefill doesn't know its routable address and decode can't reach the KV side channel. The engine block today carries `image` + `args` only — no `env`, no downward-API hook. The renderer can't materialize a working DeepSeek V3 wide-EP or Kimi K2 4P/12D pod from this example without injecting env. Either expose `env` on the engine block or have the renderer auto-inject the canonical NIXL env when disagg is configured.

design/design.md
```diff
@@ -1,818 +1,731 @@
...
+describes the shape of one worker; `workers.count` (default 1) says how many
+workers of that shape exist per ModelReplica.
+
+`topology.strategy` is a required discriminator:
+
+| Strategy | Required fields | Nodes per worker | GPUs per node | Total GPUs per worker |
+|---|---|---|---|---|
+| `Tensor` | `tensor` | 1 | `tensor` | `tensor` |
+| `TensorPipeline` | `tensor`, `pipeline` | `pipeline` | `tensor` | `tensor * pipeline` |
+| `DataExpert` | `tensor`, `data`, `dataLocal` | `data / dataLocal` | `dataLocal * tensor` | `data * tensor` |
```
dennis-upbound: KServe LLMInferenceService (since v0.16) exposes a flat `spec.parallelism` block — `tensor`, `pipeline`, `data`, `dataLocal` as int sizes, `expert` as a boolean toggle — no strategy discriminator. Our enum collapses every combination into a named bucket (`Tensor`, `TensorPipeline`, `DataExpert`), which forces a new bucket every time a real recipe mixes axes (DeepSeek V3 = TP+EP+PP; Llama 4 Maverick = TP+EP). Worth considering whether to align with KServe's flat shape — same vocabulary, no discriminator. Strategies become derived (renderer infers from which fields are set); validation moves to the composition function. Saves us from growing the enum every model release.

design/design.md
dennis-upbound: This silent-degrade pattern extends past the engine. llm-d's EPP (the Gateway API Inference Extension endpoint-picker) scores pods by queue depth, KV-cache utilization, and running requests — metrics that look fine on a misconfigured-but-Ready pod. A broken pod (wrong kv role, missing NIXL env) gets traffic just like a healthy one; only request latency/errors give it away. NVIDIA Dynamo + Grove takes the opposite tack — Grove provides atomic gang admission, so a partial gang refuses to schedule. We should pick a side. At minimum, surface `Degraded` on ModelReplica.status with the failing role + reason so users aren't debugging 504s via vLLM container logs.

design/design.md
dennis-upbound: `expert` is becoming a first-class kwarg upstream. KServe's `ParallelismSpec` exposes `tensor`, `pipeline`, `data`, `dataLocal` as int sizes and `expert` as a boolean toggle. SGLang ships `--expert-parallel-size`; vLLM 0.9 ships `--enable-expert-parallel` paired with `--data-parallel-size` (EP requires DP). So `expert` isn't just one more bucket in the discriminator — it's a named axis engines, KServe, and Dynamo already agree on. Strengthens the case for moving to flat axes (separate comment on this table) rather than extending the enum to `TensorDataExpert` / `TensorPipelineDataExpert` / etc.

design/design.md
```diff
@@ -1,818 +1,731 @@
...
+      strings: [fp8, bf16, transformer-engine, mig]
+    modelplane.ai/interconnectIntraNode:
+      string: nvswitch
+    modelplane.ai/networkInterNode:
+      string: gpudirect-tcpx
+    modelplane.ai/networkBandwidthGbps:
+      int: 200
+  capacity:
+    gpu.nvidia.com/memory:
+      value: "141Gi"
```
bassam, line 250: why is capacity separate from attributes? why not have them all at the same level? should `modeplane.ai/networkBandwidthGbps` also go under `capacity`?

design/design.md
bassam: I proposed we add imagePullSecrets and env in a prior version of this PR.

dennis-upbound: Drilling in on `env` / `envFrom` from the KV offload angle — every backend the renderer might need to wire has the same shape of requirements that don't fit `args` alone:

- LMCache: env (`LMCACHE_USE_EXPERIMENTAL`, `LMCACHE_CONFIG_FILE`) + a ConfigMap mount holding the YAML
- NixlConnector: env via downward API (`VLLM_NIXL_SIDE_CHANNEL_HOST`) — see separate comment above
- Mooncake: env (`MOONCAKE_CONFIG_PATH`, `MOONCAKE_MASTER`) + ConfigMap mount + optional Secret for object-backend creds
- Redis-as-KV: env (`LMCACHE_REMOTE_URL`) + Secret for auth

So the engine block needs `env` with `valueFrom`/`secretKeyRef`/`configMapKeyRef`, plus volume mounts for ConfigMaps and Secrets. Rather than letting ML teams hand-wire all of this per-deployment (and re-discover every backend's quirks), proposing a typed `kvCache:` block on the engine that references a fleet-level `KVOffloadTier` — filed as #72. The platform team declares the tier once per InferenceCluster (DaemonSets, control plane, hardware shape); the renderer expands the reference into the right env / mount / arg combination on every pod. Keeps the engine block from absorbing every backend's specifics.

design/design.md
negz: I'm trying to match DRA as closely as possible, and it's what they do.

https://github.com/kubernetes/enhancements/tree/cfc5d3729438f203b7428229730819bdb3d9910d/keps/sig-node/4381-dra-structured-parameters#devices-as-a-named-list

Yes bandwidth seems like it should be in capacity.

design/design.md
```diff
@@ -1,818 +1,731 @@
...
+
+Replicas are the only scaling axis. Each ModelReplica is a complete,
+fixed-topology serving instance. Scaling `spec.replicas` adds or removes whole
+instances. There's no in-cluster pod autoscaling, no KEDA or similar on workload
+clusters.
+
+The ModelDeployment XRD declares a Kubernetes scale subresource
+(`specReplicasPath: .spec.replicas`, `statusReplicasPath: .status.replicas`).
+Autoscaling is opt-in via a separate KEDA `ScaledObject`, the same pattern as
+Kubernetes Deployment + HPA. KEDA watches aggregate gateway metrics (concurrency
```
dennis-upbound: What's the telemetry stack assumption for "aggregate gateway metrics" here? Prometheus on the control plane gateway? OTel? Custom? Worth pinning down because the same surface is going to carry more than KEDA's concurrency signal — near-term consumers I can see:

- **#70 capacity signal** — per-cluster GPU availability rolled up to fleet view for the federation matcher
- **#71 prefix-hash routing** — cross-cluster prefix overlap so the gateway picks the warm cluster
- **#48 overflow** — aggregate queue depth + cost signal to decide when to spill to burst/external
- **Future KV / prefix-pool optimization (#72-shaped)** — prefix distribution + per-cluster hit rate
- **Drift / model-recall observability** — response-quality samples per `(model, version, cluster)`

Same shape across all of them: typed event from workload plane → aggregate at control plane → consumed by either a scheduler primitive or a user-facing surface. Define the contract once and every fleet-level primitive in v0.2+ plugs into it.

Also doubles as user-facing observability — "see your fleet's prefix hit rate / capacity / $/token / drift" is unique surface no per-cluster operator can offer (Baseten/Together/Fireworks don't expose fleet aggregates to customers; KServe/llm-d are cluster-local by design). Real product surface in its own right, separate from the scheduler optimizations.

Worth a sentence or two in the doc on stack assumption even if the broader signal-bus implementation lands in v0.2+. Right now the doc actively rejects per-cluster Prometheus scraping (line ~694) for autoscaling reasons — which is fine, but pushes observability up to the fleet layer without defining what lives there.

dennis-upbound: Filed as #74 with a concrete v0.1 signal-type list — request lifecycle, capacity, cold start, cost, reliability. Most of these are what Kiely's *Inference Engineering* calls out as production observability fundamentals (§7.4.3 obs list, §7.2.2 cold start phases, §7.4.2 cost decomposition, §7.3.3 reliability/failure rates), just aggregated at fleet rather than cluster level.

Happy to own the exact signal list + emitter agent + control-plane aggregation + tests for v0.1 if the design feels directionally right to you. Stack choice (OTel vs Prometheus + remote_write vs custom) plugs into whatever you land on here for the aggregate gateway metrics — I'd build the consumer side once that contract is pinned.

design/design.md
```diff
@@ -1,818 +1,747 @@
...
+      capacity["gpu.nvidia.com/memory"].compareTo(quantity("141Gi")) >= 0 &&
+      attributes["gpu.nvidia.com/features"].includes("fp8") &&
+      attributes["modelplane.ai/networkInterNode"].string == "infiniband"
+  workers:
+    topology:
+      strategy: TensorPipeline
+      tensor: 8
+      pipeline: 2
+  engine:
+    name: vLLM
```
negz: I think we can probably drop this, given engine is just an arbitrary container now.

design/design.md
bassam: I assume gpuCount as well then?

design/design.md
negz: What are `imagePullSecrets` and `envFrom` referencing? Secrets in the Modelplane control plane? Or are they just being passed down through to the inference clusters to be resolved there?

Given it's the inference clusters that do the pulling I don't think we have much choice for `imagePullSecrets`.

I note that LWS, Deployment, and Dynamo DGD all just use `podSpecTemplate` for the engine. I'm wondering if we should do the same. Or at least start with a compatibly shaped subset so we can grow into it without a breaking change. For example:

```yaml
engine:
  imagePullSecrets:
  - name: nvcr-creds
  containers:
  - image: nvcr.io/nim/meta/llama-3.1-70b-instruct:1.8.1
    env:
    - name: HF_TOKEN
      valueFrom:
        secretKeyRef:
          name: hf-token
          key: token
```

In this example `nvcr-creds` and `hf-token` both need to exist on the cluster this'll get scheduled to, not on the Modelplane cluster.

If we want these to be references to the Modelplane cluster, we'd presumably need to handle propagating them down to inference clusters. At least for pull secrets.

negz: > `shmSize`

I'm a bit wary of this - it could be the first step down a path toward rebuilding `podSpecTemplate`.

It seems like DGD and LLMIS just hardcode this to a sufficiently big default. I'd start there. If we did want to make it configurable we might consider exposing it using `engine.volumes`.

negz: See https://github.com/modelplaneai/modelplane/pull/64/changes/4f06264b640deb0d38492b07f34d9121754e29e6.

I've moved `spec.engine` -> `spec.workers.template` and made it a subset of `podSpecTemplate`.

design/design.md
negz: Yes, moved.

design/design.md
dennis-upbound: I'd vote for deferring this to v0.2 — `spec.prefill` is explicitly deferred in the latest commit, so admission can be designed alongside disagg.

design/design.md
dennis-upbound: Punt with disagg to v0.2 imo.

design/design.md
dennis-upbound: Inclined to defer — multi-fabric clusters are an early-v0.2 case. ComputeDomain (separate thread) is the upstream path.

design/design.md
dennis-upbound: Happy to defer — explicit future case. Logging here as a vocab note for when disagg lands.

design/design.md
dennis-upbound: Lean defer — niche; lands later as a TP-only `pipeline=null` variant.

design/design.md
dennis-upbound: Fine to defer to v0.2+. The flat-parallelism thread covers the v0.1 design hook.

design/design.md
dennis-upbound: OK to defer with disagg to v0.2. The engine → workers.template move gives us PodSpec downward-API when we get there.

design/design.md
dennis-upbound: Voting to defer — NVL72 placement is v0.2+. Adding `modelplane.ai/computeDomain` to the attribute vocab when we revisit.

design/design.md
dennis-upbound: Defer with multi-pod gangs to v0.2+. KEP-5729 is the upstream direction to track.

design/design.md
dennis-upbound: Marking this as defer — mostly v0.2+ engine-config-mismatch territory. v0.1 surface doesn't expose it.

design/design.md
dennis-upbound: Closing — extension path can be designed when a real second API lands. v0.1 is OpenAI-only.

design/design.md
dennis-upbound: Closing — tracking install + failure modes in PR #78. Design doc reading is fine as-is.

design/design.md
dennis-upbound: Closing — operational policy rather than design shape. Pin down in BYOC docs when we ship.

design/design.md
dennis-upbound: Closing — BYOC will carry k8s version requirements anyway, so 1.32+ for DRA is acceptable. Worth stating that requirement explicitly in the BYOC section of the design.

design/design.md
negz: Ack. Let's keep working on that as a separate design.

design/design.md
negz: > we control the vocabulary and customers need to know what values to use.

I wonder if we should avoid prefixing these with `modelplane.ai` and let the end user decide. I presume then it just becomes a contract between the team who creates InferenceClusters and the teams who create ModelDeployments.

negz: Pushed https://github.com/modelplaneai/modelplane/pull/64/changes/6472f528946e2b3b815092e47395fe0a516939c5 to clarify this.

Quick version of what I'm thinking:

* The `InferenceClass` author must author these attrs / capacities - they're not auto discovered.
* Mostly they should match real DRA attributes. We'll pass them down to form resource claims.
* However anything prefixed with `modelplane.ai/` will be filtered - it's for node pool level scheduling only.

With that in mind I don't think we need to be opinionated about the `modelplane.ai/foo` keys. They're an internal contract between IC author and MD author (despite the arguably confusing prefix). I could imagine us wanting to standardize on a convention if we had a marketplace of InferenceClasses, but I don't think we need to do that here in this design.

design/design.md
negz: I think I ended up proposing this because I was trying to avoid Expert being a boolean, which Kubernetes API conventions recommend avoiding.

It ended up with what looks like a discriminator field for something that isn't a discriminated union though. So I agree we should just go with the flat block. I've left expert out because AFAICT it's more of an engine flag than something that affects worker topology.

See https://github.com/modelplaneai/modelplane/pull/64/changes/9a92d93c2ad159b2e9469fc8ede838a3b3ab58b8.

dennis-upbound (approved):
Yay!!!

PR #75, opened 2026-05-13
Implement the updated API shape

negz:
This PR implements most of the API shape we aligned on in #64.

The exceptions are:

* No CEL (no `spec.nodeSelector`)
* No DRA support - we don't set pod requirements at all, except number of GPUs needed
* ModelEndpoint doesn't support specifying addresses as FQDNs - only IPs
* Prefill/decode (no `spec.prefill`)
* We're still built on KServe only


lib/naming.py
```diff
@@ -1,28 +1,32 @@
...
+    Deterministic names are needed so the deployment function knows the
+    LLMInferenceService name on the remote cluster for URL rewriting.
     """
-    s = s.lower()
-    s = re.sub(r"[^a-z0-9-]", "-", s)  # Replace invalid chars with hyphens
-    s = re.sub(r"-+", "-", s)  # Collapse consecutive hyphens
-    s = s.strip("-")
-    s = f"model-{s}"
-    return s[:DNS_LABEL_MAX]
+    return f"{deployment_name}-{cluster_name}"[:DNS_LABEL_MAX]
```
negz: These could collide if deployment name was long enough. Same issue with `endpoint_name`.

functions/compose-inference-cluster/main.py
```diff
@@ -32,32 +38,48 @@
...
 BACKEND_RESOURCE_KEY = "kserve-backend"
 
+# Hardcoded system pool config. Provisioned for every GKE cluster to
+# host control-plane components (Envoy Gateway, KEDA, etc.). Not exposed
+# in the user-facing API.
+SYSTEM_POOL_NAME = "system"
+SYSTEM_POOL_MACHINE_TYPE = "e2-standard-4"
+SYSTEM_POOL_NODE_COUNT = 1
+SYSTEM_POOL_MIN_NODE_COUNT = 1
+SYSTEM_POOL_MAX_NODE_COUNT = 2
```
negz: Should these be hardcoded in the GKECluster function (or XR spec?) instead?

functions/compose-inference-cluster/main.py
```diff
@@ -219,18 +274,39 @@ def derive_conditions(self, cluster_ready):
...
+        """
+        gke_node_pools: list[gkev1alpha1.NodePool] = [self.system_pool_for_gke()]
+
+        for pool in self.xr.spec.nodePools or []:
+            cls = self.classes.get(pool.className)
+            if not cls or not cls.spec.provisioning or not cls.spec.provisioning.gke:
+                # Class has no GKE provisioning block - can't provision
+                # this pool. Skip rather than fail; capacity reporting
+                # still works for pre-existing pools.
+                continue
```
negz: Shouldn't we fail? What does "capacity reporting still works for pre-existing pools" mean? How would pools be pre-existing? Is this covering the case where we used an InferenceClass to provision but the class was since deleted? If so I assume we'd have bigger problems; like being unable to schedule to it.

negz: Or maybe not, since we could use the cluster's `status.capacity` when scheduling.

functions/compose-model-deployment/main.py
```diff
@@ -1,9 +1,13 @@
...
-and composes Envoy Gateway Backend + HTTPRoute resources on the control plane
-for unified endpoint routing.
+This function discovers InferenceClusters, matches the deployment's
+topology against available capacity, creates a ModelReplica per
+selected cluster, and creates one ModelEndpoint per replica for
+ModelService to route to.
+
+Routing on the control plane is the responsibility of ModelService:
+this function does not compose the HTTPRoute. Users author a
+ModelService to expose a deployment.
```
negz: This feels like an artifact of an LLM explaining how this used to work before this change. I don't think we need to explain what the function _doesn't_ do unless it might be surprising to a reader.

functions/compose-model-endpoint/main.py
```diff
@@ -0,0 +1,103 @@
...
+ModelEndpoint is a reachable inference endpoint. This function parses
+spec.url and composes an Envoy Gateway Backend on the control plane
+pointing at the URL's host:port. ModelService reads the resulting
+backend name from status.routing.backendName to build its HTTPRoute.
+
+External / SaaS endpoints (fqdn-style Backends) are deferred. For now,
+spec.url is expected to be an http://<ip>:<port>/... shape.
+"""
+
+from urllib.parse import urlparse
```
negz: I'd like to avoid importing specific functions. (Per Google Python style.)

apis/modeldeployments/definition.yaml
```diff
@@ -15,157 +15,183 @@ spec:
...
-                          load decreases. Prevents flapping when traffic
-                          is bursty. Defaults to 300 (5 minutes).
+                          Pod spec for inference workers.
+                        properties:
+                          containers:
+                            type: array
+                            minItems: 1
+                            description: >-
+                              Containers for the inference pod. The container
+                              named "engine" is the inference engine; additional
```
dennis-upbound, line 112: is this `engine` magic-name a normal convention?

docs/concepts.md
```diff
@@ -12,127 +13,147 @@ This page explains the key resources and how they relate.
...
 
 ```bash
 kubectl get ig default
 ```
 
-## InferenceEnvironment
+## InferenceClass
 
-An InferenceEnvironment represents a Kubernetes cluster configured for model
+An InferenceClass is a tested recipe for a GPU node pool. It bundles:
```
dennis-upbound, line 64: "tested recipe" seems a little vague here to the user. are we saying that InferenceClasses are valid/blessed by platform team/Upbound?

docs/concepts.md
```diff
@@ -12,127 +13,147 @@ This page explains the key resources and how they relate.
...
+
+- **Resources**: what hardware the class exposes (GPU count, memory). Used by
```
dennis-upbound, line 66: the difference between provisioning and resources make sense but I wonder if the naming could be clearer? I'm still not sure what cases which is which - are we saying for things that have pre-defined pool types vs those you can "assemble" them?

functions/compose-model-replica/main.py
```diff
@@ -0,0 +1,259 @@
...
+        model_name = ""
+        container_args = []
+        for arg in list(engine.args or []):
+            if arg.startswith("--model="):
+                model_name = arg.split("=", 1)[1]
+            else:
+                container_args.append(arg)
+
+        container = self._build_container(engine, gpu_per_pod, container_args)
+        pod_spec = self._build_pod_spec(template, container)
```
dennis-upbound, line 118: if the template has a spec with sidecars/containers other than engine, do we silently drop them here? can we add a test for engine + sidecar?

apis/modeldeployments/definition.yaml
```diff
@@ -15,157 +15,183 @@ spec:
...
+                              containers pass through as sidecars.
+                            x-kubernetes-validations:
+                            - rule: "self.exists_one(c, c.name == 'engine')"
+                              message: "containers must include exactly one container named 'engine'"
+                            items:
+                              type: object
+                              required: [name, image]
+                              properties:
```
dennis-upbound, line 120: to support multi-node with Kserve, I needed Ray which needed `command` to bootstrap (and in some case `volumeMounts`, `volumes`) - dropping a note here for myself to revisit adding to this as we land multi-node and/or modelcache

functions/compose-model-replica/main.py
```diff
@@ -0,0 +1,259 @@
...
+        llmis_template = self._build_llmis_template(template, pod_spec)
+
+        llmis_spec: dict = {
+            "model": {"uri": f"hf://{model_name}" if model_name else "hf://unknown"},
+            "replicas": 1,
```
dennis-upbound: shouldnt this be spec.workers.count instead of 1? can we add a test for multi-worker?

docs/concepts.md
```diff
@@ -12,127 +13,147 @@ This page explains the key resources and how they relate.
...
+### Scaling
 
-A ModelDeployment is the ML team's interface. It says "deploy this model to N
-environments" and produces a working endpoint.
+Replicas are the only scaling axis. Each `ModelReplica` is a complete,
+fixed-topology serving instance. Scaling `spec.replicas` adds or removes whole
+instances. There's no in-cluster pod autoscaling.
 
-When a ModelDeployment is created, the scheduler:
+The `ModelDeployment` XRD declares a Kubernetes scale subresource. Autoscaling
```
dennis-upbound: does it?

apis/modelendpoints/definition.yaml
```diff
@@ -0,0 +1,63 @@
...
+        required: [spec]
+        properties:
+          spec:
+            type: object
+            required: [url]
+            properties:
+              url:
+                type: string
+                description: >-
+                  Informational URL for this endpoint. For Modelplane-
```
dennis-upbound: i'm a little confused by the word Informational here. what happens with an invalid url? I assume broken routing?

dennis-upbound (approved):
Awesome! Looks great! A few minor comments/questions

apis/modeldeployments/definition.yaml
negz: No, not that I'm aware of. Open to other ideas to figure out which container is the engine one. To be fair we might not even need to know - we do right now but it's for a temporary hack where parse `--model` for KServe.

docs/concepts.md
negz: Ack, but I see these docs as internal only. I assume we'll do a significant rework when we're getting ready to launch.

docs/concepts.md
negz: I'm having trouble parsing your question 😆.

Provisioning is what Crossplane uses to call the cloud API to actually create a node pool.

Resources is temporary. It'll become attributes and capacity once we implement that (in a future PR). It's only used as as scheduler hint for matching which cluster a ModelReplica should land on. (Same will be true for attrs and cap).

Is that what you were unsure about?

functions/compose-model-replica/main.py
negz: Yes we should, but let's tackle that when we switch the backend from LLIMS -> Deployment/LWS/DGD.

docs/concepts.md
negz: It doesn't. 😬 We actually need Crossplane v2.3 (released any day now) to support this. I'll remove it for now.

apis/modelendpoints/definition.yaml
negz: Good catch - it's not informational. Not sure why the LLM that wrote this decided it was. I had to fix that in the docs also recently.

docs/concepts.md
dennis-upbound: ahh ok gotcha makes sense. hadn't grokked that.

apis/modeldeployments/definition.yaml
dennis-upbound: no other bright ideas

functions/compose-model-replica/main.py
negz: Good catch - fixed. I think my multi-node changes were actually completely wrong / hallucinated. It should now match KServe's docs though I haven't actually deployed it to test.

PR #76, opened 2026-05-13
WIP: ModelCache design doc + examples

dennis-upbound:
Design doc + 11 example yamls for the `ModelCache` primitive. Pure design — no code, no CRDs. Supersedes the sketch in #66.

## Why v0.1 PVC-backed (TL;DR)

Multi-node LWS serving needs the same weight bytes mounted by every pod in the gang. The minimum primitive for that is an RWX PVC hydrated once by a side Job (not an init container — KServe's storage-initializer OOMs at 4/8/16 GiB on Kimi K2 / Llama 405B) and mounted read-only by every pod that references it. v0.1 ships exactly that. Single-node scale-up benefits as a side effect (no per-replica HF pull on scale events). The harder content-addressed wins — cross-deployment dedup, lazy load, cross-cluster sharing — stay in v0.2 behind the same user-facing API.

## Read

- [design/modelcache/design.md](design/modelcache/design.md) — ~440 lines, 5 Mermaid diagrams
- [design/modelcache/examples/](design/modelcache/examples) — 8 v0.1 OSS references at the top level + 3 v0.2 previews under `content-addressed/` (commercial substrate vs OSS `Custom` hook explained in `content-addressed/README.md`)

## Alignment asks

1. **Timeline shift**: v0.2 → v0.1. Driven by multi-node serving (#61) and DRA landing in v0.1 (#56). Flagged at the top of the doc; rationale in *Why v0.1 PVC-backed* above.
2. **Commercial substrate** (per Bassam): `ContentAddressed` = Upbound hosted weight delivery; `Custom` (webhook) = OSS extension point. Customers run fully OSS via `Custom`; commercial product is fleet-scale dedup + Modal-class cold-start hosted by Upbound. Reflected in the Storage-backends table + v0.2 section.
3. **Five v0.1 open questions** in the doc's *Open questions* section. Each has a documented lean; please push back if any feel wrong.

## Companion

PR #78 (`dennis/modelcache-impl`) lands the v0.1 OSS scaffolding (XRD, composition function, `ModelDeployment.spec.caches` wiring, four standalone examples, an end-to-end Qwen-on-GKE cold-start demo with idempotent setup / demo / cleanup scripts) against this design.

## Related

#66 (implementation tracker, body refactored) · #61 (closed; mechanism absorbed) · #56 (DRA, v0.1) · #72 KVOffloadTier · #71 ModelService routing affinity · PR #64 · PR #75 · PR #78

Issue #74, opened 2026-05-12
Fleet signal bus: capture layer for cross-cluster signals (v0.1)


dennis-upbound:
Multiple primitives need the same thing: typed signals captured at the workload plane, aggregated at the control plane, consumed by either a scheduler primitive or a user-facing surface. Each will rebuild the same plumbing if we don't define it once.

This is the v0.1 substrate. [#77](https://github.com/modelplaneai/modelplane/issues/77) is the parallel v0.2 primitive for app-observability traces — both share OTLP as the transport (one workload-plane agent emits, one control-plane collector fans out to two different consumers).

## Why fleet-level

Per-cluster engines (vLLM, SGLang, Dynamo) and per-cluster gateways already emit most of this data. **Modelplane's job is the rollup and cross-cluster comparison, not the emission.** A single cluster's P99 TTFT is a deployment concern; "P99 TTFT in eu-west-1 is 3× us-east-1, the cluster is degrading" is a fleet concern. Same for capacity, cost, prefix overlap, failure rates.

## Consumers and emitters on the bus

| Primitive | Emits | Consumes |
|---|---|---|
| [#70 capacity signal](https://github.com/modelplaneai/modelplane/issues/70) | (none — pure consumer) | Per-cluster GPU availability for the federation matcher |
| [#71 routing affinity](https://github.com/modelplaneai/modelplane/issues/71) | (none — pure consumer) | Cross-cluster prefix-hash overlap so the gateway picks the warm cluster |
| [#48 overflow](https://github.com/modelplaneai/modelplane/issues/48) | (none — pure consumer) | Aggregate queue depth + cost to decide when to spill |
| [#66 ModelCache](https://github.com/modelplaneai/modelplane/issues/66) | Hydration latency, bytes staged, per-cluster ready state | (none in v0.1) |
| [#72 KVOffloadTier](https://github.com/modelplaneai/modelplane/issues/72) | Per-tier hit rate, eviction rate, capacity util | (none in v0.1) |
| **Future** | Spec-dec acceptance, quality drift, cost samples, model-recall | Cost-aware placement, drift detector, intent-based serving SLAs (`ttft.p99` on `ModelService`) |

## Transport: OTLP

[...]

PR #78, opened 2026-05-14
v0.1 ModelCache + multi-node LWS unblock

dennis-upbound:
Warm-cluster demo recording

*Warm-cluster run: cache pre-hydrated, LWS gang Ready in 65s, engine answers 86s after gang Ready. 4× speed.*

---

Adds the v0.1 `ModelCache` primitive (XRD + composition function + `ModelDeployment.spec.caches[]` wiring) and turns multi-node `TensorPipeline` serving from broken-by-default into a working path. See *Multi-node LWS — what was broken and what we did* below for the specific issues we untangled.

**Base:** `demonstration` (#75) so the diff stays scoped. GitHub will rebase to `main` automatically when #75 merges.

## What ships

**`ModelCache`** — composes a per-cluster RWX PVC + a one-shot hydration Job on every matched `InferenceCluster`. Pods that reference the cache via `ModelDeployment.spec.caches[]` get the PVC mounted at the engine's expected path; no per-replica HuggingFace pull at boot.

| Aspect | v0.1 values |
|---|---|
| Artifact kinds | `Weights`, `Tokenizer`, `Bytes` |
| Sources | `huggingFace`, `s3`, `http`, `inline` implemented; `oci`, `configMap` discriminators reserved |
| Storage backends | `PVC` (Modelplane-managed), `ExistingPVC` (customer-managed) |
| Replication | `AllMatchingClusters` |
| Deployment wiring | `ModelDeployment.spec.caches: [{ name }]` → `LLMInferenceService.spec.model.uri = pvc://...` |

**Cloud-agnostic CSI capability plumbing.** `InferenceCluster.spec.storage.csiDrivers: [SharedFilesystem | ObjectStorageMount | BlockDevice]` declares the cluster's semantic capability. The GKE composition branch reads the underlying VPC name from `GKECluster.status.network.name` and auto-composes a `modelplane-rwx` StorageClass on the workload cluster with `parameters.network` pinned. Same pattern will hook into EKS / AKS with their respective knobs.

**GCP API auto-enable.** `compose-gke-cluster` emits a `ProjectService` MR for `file.googleapis.com` whenever the user opts into `SharedFilesystem`.

**1-pager design:** [`design/modelcache/README.md`](design/modelcache/README.md). Replaces the earlier verbose design doc (PR #76, closed in favor of this focused page).

**End-to-end demo:** [`examples/qwen-cached-demo/`](examples/qwen-cached-demo/) deploys Qwen 2.5 0.5B on a 2-pod LWS gang with a real chat-completion response. See [`TOPOLOGY.md`](examples/qwen-cached-demo/TOPOLOGY.md) for the composed XR / MR layout.

**Tests:** model-cache × 3 + model-replica-with-cache + the full existing suite — 20/20 pass.

## Multi-node LWS — what was broken and what we did

Bringing the demo up needed fixes layered across the function, the LIS shape, the cache hydration command, the engine wrapper, and the cluster's StorageClass. Each row below is something an OSS user would hit on a fresh cluster; calling them out so reviewers know what each fix buys us.

| Problem | Fix |
|---|---|
| **No Ray bootstrap on the gang.** vLLM with `--pipeline-parallel-size > 1` needs a Ray cluster spanning the gang's pods. KServe's `ServingRuntime` injects this for you; the custom-template path we use doesn't. Worker pod never joined the leader's Ray cluster, vLLM's placement group hung forever. | `compose-model-replica` injects a 5-line shell wrapper as the container `command` when `topology.strategy == TensorPipeline`. Leader runs `ray start --head` then execs the engine; worker runs `ray start --address=$LWS_LEADER_ADDRESS:6379 --block`. Bootstrap lives as a Python constant — when we swap KServe for our own `LeaderWorkerSet` composition later, the same constant moves verbatim. |
| **KServe v0.17+ flat worker schema.** The existing function generated `worker: {size, template}` (v0.16 shape); v0.17+ wants `worker` as a flat `PodSpec` with `containers[]`. `parallelism.tensor` is now per-pod GPU count and `parallelism.pipeline` is required when `worker` is present. | Switch the generated LIS shape to flat `worker:` + set `parallelism.{tensor,pipeline}` from the topology. |
| **KServe v0.17+ stopped injecting `--model` into engine argv.** `model.uri = pvc://...` mounts the PVC at `/mnt/models` but the engine container's args don't get `--model=<path>` anymore. vLLM defaulted to `facebook/opt-125m` and silently downloaded it from HF instead of using the mounted cache. | Append `--model=/mnt/models` to the engine args when `spec.caches[]` is set. |
| **`vllm serve` vs `python -m vllm.entrypoints.openai.api_server`.** `vllm serve` takes the model as a positional argument; the api_server entrypoint (the image's default `ENTRYPOINT`) takes `--model=`. The first wrapper invoked `vllm serve "$@"`, which silently exited on argparse mismatch. | Wrapper invokes the same `python -m vllm.entrypoints.openai.api_server` the default `ENTRYPOINT` uses. Same arg shape for single-node and multi-node. |
| **Filestore CSI defaults to the GCP `default` VPC.** GKE's chart-shipped `standard-rwx` StorageClass has no `network` parameter, so Filestore instances landed on `default` and were unreachable from cluster nodes — PVCs sat Pending forever with NFS mount timeouts. | `compose-inference-cluster` (GKE branch) auto-composes a workload-cluster StorageClass (`modelplane-rwx`) with `parameters.network = <our VPC>`. Network surfaced via `GKECluster.status.network.name`. |
| **`lost+found` triggered a false-positive skip.** The hydrate Job's "already populated?" check did `ls -A /mnt/model`. Filestore's ext4-style filesystem puts `lost+found` at mount root, so a fresh PVC always looked non-empty; the Job exited with `artifact already hydrated, skipping` and the cache stayed Ready with no weights on disk. | Filter `lost+found` out of the emptiness check. |
| **`huggingface-cli` was deprecated** in `huggingface-hub 1.x` and the `[cli]` extra removed. The hydrate Job's `pip install huggingface_hub[cli]` printed a warning and `huggingface-cli download` exited with a deprecation notice, no bytes written. | Switch to `hf download` (new CLI, same flags). |
| **Filestore API isn't enabled by default** on fresh GCP projects. PVCs sit `Pending` indefinitely with `SERVICE_DISABLED` in their workload-cluster events; the GKE cluster itself looks healthy. | `compose-gke-cluster` emits a `ProjectService` MR for `file.googleapis.com` whenever the Filestore CSI addon is on. |
| **T4 / Turing doesn't support bfloat16** — Qwen 2.5 ships in bf16. vLLM exits with `ValueError: Bfloat16 is only supported on GPUs with compute capability of at least 8.0`. Not a Modelplane-layer issue; documented as an example concern. | Demo deployment passes `--dtype=half`. |

Net result: a fresh `kubectl apply -f examples/qwen-cached-demo/` provisions a GKE cluster with Filestore CSI enabled + a VPC-pinned StorageClass, hydrates Qwen 2.5 0.5B onto a 100 GiB RWX PVC, boots a 2-pod LWS gang on two T4 nodes, both pods mount the same PVC, the leader's Ray head and the worker's Ray client form one cluster, vLLM's pipeline-parallel placement group satisfies, and `curl /v1/chat/completions` returns a real Qwen response over HTTP 200.

## Out of scope for v0.1

- `Adapter` / `Engine` artifact kinds — composition shape needs more design before locking
- Storage backends beyond `PVC` / `ExistingPVC` — the `backend` enum is the extension point
- `AllMatchingNodes` replication granularity — per-node staging is a separate problem
- Signal-bus emission (hydration latency, bytes staged) — depends on #74

## Workarounds / hacks (cleanup before / after merge)

Real composition-code workarounds shipping in this PR. The demo scripts under `examples/qwen-cached-demo/` have their own pile of provider-gcp / provider-helm / provider-kubernetes workarounds — those will get gitignored and graduate into a smoke-test fixture, so they're not called out here.

### Composition-code hacks

| Where | What | Upstream root cause | Cleanup |
|---|---|---|---|
| `functions/compose-gke-cluster/main.py` (TODO L129) | `ZONE_RESOURCE_POOL_EXHAUSTED` isn't surfaced through the nodepool MR's `conditions`; GCE keeps the pool in `PROVISIONING` forever. | GCE/GKE not propagating instance-level capacity failures up to the pool's status. | Add an IGM-status check or gcloud query in the function (bigger lift, defer). |
| `functions/compose-kserve-backend/main.py` (LLMInferenceServiceConfig presets) | Composes five empty `spec: {}` `LLMInferenceServiceConfig` instances on the workload cluster because the `kserve-llmisvc-resources` chart defines the CRD but doesn't ship the controller-required defaults. | KServe chart packaging gap — chart should land these. | File upstream KServe issue. Once they ship the defaults, drop the preset compose code. |
| `functions/compose-kserve-backend/main.py` (router-route exclusion) | We deliberately skip `kserve-config-llm-router-route` because the chart DOES include it, and naive composition leads to Helm "exists and cannot be imported" failures. The list-overlap is fragile across KServe versions. | Same chart packaging gap. | Same as above. |

### Mid-session debugging interventions captured so they don't repeat

| Bug | What we fixed | Test that now catches it |
|---|---|---|
| Workload-cluster `Object` MR for `storage-class-rwx` composed without `metadata.namespace` set; Crossplane reconcile errored with "an empty namespace may not be set when a resource name is provided". | Added `metadata.namespace=modelplane-system` to the composed Object. | `test-inference-cluster-csi-shared-filesystem` asserts on the namespace. |
| `storage-class-rwx` Object used `policy: DeriveFromObject` readiness, but StorageClass has no Ready condition — Object MR stayed `Ready=False` forever. | Switched to `policy: SuccessfulCreate`. | Same test. |
| Function composed the SC Object but didn't propagate observed Ready into `rsp.desired.resources["storage-class-rwx"].ready` — XR aggregator left the InferenceCluster `Ready=False`. | Read observed condition and set `READY_TRUE` on the response. | Same test (regression-friendly; the compositiontest framework doesn't expose `.ready` for direct assertion). |
| `compose-gke-cluster` didn't track `projectservice-filestore` in `mark_readiness`; XR stuck on "Unready resources: projectservice-filestore" even after the API was enabled. | Added it to the tracked-resource list. | `test-gkecluster-filestore-addon`. |
| Preset Object MRs pointed at `<xr-name>-cluster` (the Helm ProviderConfig) instead of `<inferencecluster>-cluster-kubeconfig` (the ClusterProviderConfig). | Strip the `-kserve` suffix from the KServeBackend XR name to derive the parent + the right CPC name. | `test-backend-presets` asserts on `providerConfigRef.name`. |

### Must

- [ ] **Rebase to `main`**. PR is currently based on `demonstration` (#75) so the diff stays scoped. GitHub will fast-forward once #75 lands.
- [ ] **Migrate `spec.engine` → `spec.workers.template`** per Nic's merged commit on [PR #64](https://github.com/modelplaneai/modelplane/pull/64) (the `engine` field is now a curated subset of `PodTemplateSpec`). Affects `apis/modeldeployments/definition.yaml`, `apis/modelreplicas/definition.yaml`, `functions/compose-model-replica/main.py`.
- [ ] **Gitignore the `examples/qwen-cached-demo/` demo scripts** (or relocate under an ignored path). They'll graduate into a smoke-test fixture; not part of the merged surface area.
- [ ] **Review**.

### Should

- [ ] File upstream KServe issue for the chart packaging gap (missing default `LLMInferenceServiceConfig` instances + `kserve-llmisvc-resources` Helm release blocked by chart-shipped vs externally-created preset overlap).

### Nice to have

- [ ] kind-cluster smoke test in CI that reconciles Crossplane against the package and asserts XR Ready aggregation. Would have caught half the bugs we hit during this PR before they hit a live demo.

## Demo flow

```sh
export GCP_PROJECT=my-gcp-project
./setup.sh         # ~5-10 min on first run: GKE provision + Filestore CSI install + stack
./demo.sh          # cache hydrate + LWS gang up + curl
./cleanup-demo.sh  # iterate: deletes workload, cluster + infra stay
./cleanup.sh       # full teardown
```

Recordings of the warm-cluster demo and the full cold-start flow attached in a follow-up comment.

## Related

#74 Fleet signal bus (status emission, future) · #75 API shape parent · PR #76 (closed; design folded into `design/modelcache/README.md`)


dennis-upbound:
## Known issue: control-plane envoy proxy gets stuck on xds during long demos

Hit this twice now during cold-start demo recordings — flagging so it's not lost.

**Symptom**: the control-plane envoy proxy pod (`envoy-modelplane-system-modelplane-*` in the `envoy-gateway-system` namespace) goes to **1/2 Ready** and the Gateway reports **`PROGRAMMED=False`** even though the InferenceCluster + ModelService composition is fine. Curl from inside the cluster gets `Connection refused` on `172.18.255.200:80`, not a 5xx.

**Root cause** (suspected): the proxy's envoy container can't reach its xds source. Container logs show repeated:

```
[warning][config] DeltaAggregatedResources gRPC config stream to xds_cluster closed: 14,
  upstream connect error or disconnect/reset before headers.
  reset reason: connection timeout
```

The long-running gRPC config stream from the envoy proxy to the envoy-gateway control plane goes stale during multi-hour sessions, and the proxy never recovers on its own.

**Workaround** (what unblocked both demos):

```
kubectl delete pod -n envoy-gateway-system envoy-modelplane-system-modelplane-<hash> --wait=false
```

The Deployment recreates the pod, the new proxy gets a fresh xds stream, the Gateway flips to `PROGRAMMED=True` within ~15s, and routing works again.

**Why this matters here**: not something this PR introduces, but it bites the demo flow when the recording session is long enough — by the time we get to the curl step the proxy has already drifted. Worth either:

1. Filing upstream on `envoy-gateway` (long-running xds gRPC stream not self-healing) — the symptom + fix matches several open issues there.
2. Adding a defensive controller on Modelplane's side that watches the Gateway condition and bounces the proxy pod when `PROGRAMMED` flips False for >N minutes.
3. (Demo-only) `kubectl rollout restart` of the envoy proxy as the first step of `demo.sh` so the recording can't catch it mid-drift. Trivial but not addressing the underlying issue.

Not blocking for this PR.

PR #80, opened 2026-05-20
Add ModelCache v1alpha1 API

dennis-upbound:
Follow-up to #82. Builds on the cluster-level `spec.storage.rwxCache` shape.

Adds the ML-team-facing opt-in for cross-deployment sharing, independent cache lifecycle, and proactive pre-staging on top of the invisible per-replica caching Modelplane already provides for multi-node deployments.

API surface + docs + examples only. Composition function follows in a separate MR.

## What this adds

- `apis/modelcaches/definition.yaml` — `ModelCache` XRD. Artifact source + mount path + optional size override + cluster selector. Storage class is always inherited from the target cluster's `rwxCache`; ML teams don't pick it.
- `apis/modeldeployments/definition.yaml` — `spec.modelCacheRef` (singular, matching the existing `inferenceClusterRef` pattern).
- `apis/modelreplicas/definition.yaml` — same `spec.modelCacheRef`, inherited from the parent ModelDeployment.
- `docs/concepts.md` — `## ModelCache` section positioned as the optional shared/lifecycled overlay on top of the cluster-level invisible caching; mermaid diagram updated.
- `examples/cache/model-cache-qwen.yaml` — Qwen 0.5B cache example.
- `examples/deployment/model-deployment-cached.yaml` — deployment referencing the cache via `spec.modelCacheRef`.

## How this relates to #82

| Scenario | #82 (this PR's base) | This PR adds |
|---|---|---|
| Multi-node, no cache reference | Auto-provision per-replica PVC from cluster `rwxCache` | (no change) |
| Multi-node + `modelCacheRef` set | n/a | Mount shared cache PVC across replicas; lifecycle decoupled from deployment |
| Single-node + `modelCacheRef` set | n/a | Mount shared cache PVC; cold-start optimization |
| Single-node, no cache reference | Ephemeral fetch in engine container | (no change) |

ML teams who don't care about sharing or pre-staging never see this surface — the cluster default still handles multi-node invisibly.

## What this does NOT change

- `examples/deployment/model-deployment.yaml` (the canonical single-node example) — untouched.
- `docs/getting-started.md` — untouched. ModelCache stays an opt-in feature; the basic flow doesn't use it.
- No composition function logic.

### DIFF
--- a/apis/modelcaches/definition.yaml
+++ b/apis/modelcaches/definition.yaml
@@ -1,0 +1,135 @@
+apiVersion: apiextensions.crossplane.io/v2
+kind: CompositeResourceDefinition
+metadata:
+  name: modelcaches.modelplane.ai
+spec:
+  group: modelplane.ai
+  names:
+    categories: [crossplane, modelplane]
+    kind: ModelCache
+    plural: modelcaches
+    shortNames: [mc]
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    referenceable: true
+    additionalPrinterColumns:
+    - name: REPO
+      type: string
+      jsonPath: .spec.artifact.source.huggingFace.repo
+    - name: READY
+      type: string
+      jsonPath: .status.summary.ready
+    schema:
+      openAPIV3Schema:
+        type: object
+        required: [spec]
+        properties:
+          spec:
+            type: object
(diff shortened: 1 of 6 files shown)


dennis-upbound:
Closing in favor of consolidated cache design — superseded by the new branch combining the final API + design doc.

Issue #13, opened 2026-03-30
Use the Crossplane CLI instead of the Upbound CLI


negz:
### What problem are you facing?

Modelplane currently uses the `up` (Upbound) CLI for project building, testing, and packaging. The project tooling is [being contributed to Crossplane](https://github.com/crossplane/crossplane/issues/6840) — `up project` will become `crossplane project`. At the same time, the CLI (crank) is being broken out of the crossplane/crossplane repo into its own repo.

For launch, Modelplane should use the Crossplane CLI rather than the Upbound CLI. Modelplane is an open-source project built on Crossplane, and its build toolchain should reflect that — contributors shouldn't need the Upbound CLI to build and test.

### How could Modelplane help solve your problem?

Once the project tooling lands in the Crossplane CLI, switch `upbound.yaml` to whatever the Crossplane project manifest is, update the Nix flake to provide the Crossplane CLI instead of (or alongside) `up`, and update any documentation or CI that references `up project build`/`up project test`.

negz:
Done in https://github.com/modelplaneai/modelplane/pull/83.

Issue #10, opened 2026-03-30
Build function images with Nix instead of the up CLI


negz:
### What problem are you facing?

The `up` CLI's Python function builder tars each function directory (following symlinks) and injects it into a base image. This works for simple functions but creates friction for a project of Modelplane's size:

- **Shared library requires symlinks.** The top-level `lib/` directory is symlinked into each function directory so functions can import shared code. This is a workaround, not a feature.
- **No third-party packages.** Functions are limited to the standard library and the function SDK. We store CRDs as JSON instead of YAML because we can't depend on `pyyaml`.
- **Type checking in CI doesn't work.** Function directories aren't Python packages on disk — they have no `__init__.py` or `pyproject.toml`, so type checkers like pyright and ty can't resolve relative imports. IDE-based type checking generally works, but we can't run type checks in CI.

### How could Modelplane help solve your problem?

Once [crossplane/cli#21](https://github.com/crossplane/cli/issues/21) lands, switch to building function runtime images with Nix (or Docker) and passing them to the `up` CLI as pre-built images. The CLI would skip the build step and just add the Crossplane package metadata layer.

This lets us structure the project as a standard Python monorepo — a real package with `pyproject.toml`, dependencies declared in one place, shared library installed via normal packaging, type checking and test runners that work out of the box.

Blocked on [crossplane/cli#21](https://github.com/crossplane/cli/issues/21).

negz:
Done in https://github.com/modelplaneai/modelplane/pull/83.

PR #83, opened 2026-05-21
Migrate to the Crossplane CLI

negz:
Addresses #10 and #13.

There are two related things going on in this PR.

### Moving to the Crossplane CLI

Modelplane has depended on the Upbound `up` CLI for building, testing, and pushing function packages since the project started. The tooling that `up` provides is being upstreamed into the [Crossplane CLI][cli-issue], where the open-source community will align on tooling for control plane projects.

I want Modelplane to explore what using control plane projects looks like for a real, complex project and to feed that experience back into the Crossplane CLI.

This PR swaps `up` for `crossplane`. `upbound.yaml` becomes `crossplane-project.yaml`  We keep `docker-credential-up` because we still need it for `xpkg.upbound.io` registry auth.

[cli-issue]: https://github.com/crossplane/crossplane/issues/6840
[cli-repo]: https://github.com/crossplane/cli

### Limiting the Crossplane CLI to schemas and packaging

The `crossplane` CLI (and `up`) aim to handle the entire lifecycle of a control plane project. That includes building and testing function packages. This makes a lot of sense for folks starting out with control planes. We want them to be able to configure their control planes using Python, Go, KCL etc without needing to become experts in the the relevant language toolchain.

On the other hand, one of the biggest benefits of using a general purpose programming language like Python to configure your control plane is the extensive and mature ecosystem for things like testing and linting. With `up`, Modelplane was in a split brain situation. `up` built the Python code into function runtimes, but we still needed another build tool (in our case Nix) to run code quality checks like linters consistently across CI and dev environments. It was easy for our two build tools (Nix and the `up` CLI) to drift.

To that end I've opened https://github.com/crossplane/cli/pull/24, and pinned Modelplane on that PR. The PR does two things:

1. It lets you tell `crossplane` not to build function runtimes (i.e. function OCI images). You can instead use your own build tool to roll your own.
2. It lets you tell `crossplane` to only generate schemas for the languages you use.

The result is that you can choose to use the `crossplane` CLI only for (Crossplane) package dependency management, schema generation, and Crossplane packaging - i.e. decorating a regular OCI image with Crossplane package metadata.

I've committed the generated Python models to the repo. This allows us to use them for reproducible builds, type checking, etc in CI without needing network to pull them on every build. (The tradeoff is they're a massive amount of code.)

I've also switched the Python build system from Hatch to `uv`, since `uv2nix` seems to be the modern and minimal way to do Python builds with Nix.


dennis-upbound:
lgtm afaict!

PR #84, opened 2026-05-21
API: Add ModelCache for multi-node support with backend override

dennis-upbound:
Multi-node serving today has no safe way to share model weights across the pods of a LeaderWorkerSet gang. Concurrent fetches across pods don't coordinate, and the failure modes are mostly invisible: torn files from simultaneous writes, version drift when the source updates mid-pull, partial downloads serving incorrect weights with no surface error, HuggingFace rate-limit storms.

This PR proposes `ModelCache`, a resource that stages a model artifact on a workload-cluster RWX PVC once and mounts it into every consuming pod. Multi-node deployments require a `modelCacheRef`; single-node deployments may reference one as a cold-start optimization. The full proposal — including alternatives considered (`CacheClass`, storage backend enum, cluster-level capacity envelope, configurable mount path) — is in `design/modelcache.md`, intended as the first one-pager addition alongside the base design.

This PR ships the API surface only: XRDs for `ModelCache`, `modelCacheRef` on `ModelDeployment` and `ModelReplica`, `cluster.<source>.cache.storageClassName` on `InferenceCluster`, an end-to-end example, and `docs/concepts.md` updates. Composition functions follow.

### DIFF
--- a/apis/inferenceclusters/definition.yaml
+++ b/apis/inferenceclusters/definition.yaml
@@ -81,6 +81,20 @@ spec:
                             default: private_key
                             minLength: 1
                             maxLength: 253
+                      cache:
+                        type: object
+                        description: >-
+                          ModelCache configuration for this cluster.
+                        properties:
+                          storageClassName:
+                            type: string
+                            default: modelplane-rwx
+                            description: >-
+                              Name of the RWX StorageClass for ModelCache
+                              PVCs. The admin creates the StorageClass on
+                              the workload cluster (must support
+                              ReadWriteMany dynamic provisioning).
+                            minLength: 1
                   gke:
                     type: object
                     description: >-
@@ -98,6 +112,22 @@ spec:
                       kubernetesVersion:
                         type: string
                         default: "1.35"
+                      cache:
+                        type: object
+                        description: >-
+                          ModelCache configuration for this cluster.
+                        properties:
+                          storageClassName:
+                            type: string
+                            default: modelplane-rwx
+                            description: >-
+                              Name of the RWX StorageClass for ModelCache
+                              PVCs. At the default value, Modelplane
+                              provisions Filestore Enterprise via the
+                              Filestore CSI addon and composes the
+                              StorageClass; set this to a different name to
+                              use one the admin has already created.
+                            minLength: 1
               nodePools:
                 type: array
                 description: >-
--- a/design/design.md
+++ b/design/design.md
@@ -715,3 +715,4 @@ Instead, I'd prefer Modelplane to be opinionated about which orchestrator to
 use, dispatching based on topology and engine requirements. Users describe what
 they want; Modelplane picks the lightest composition path that satisfies it.
 
+
--- a/apis/modelcaches/definition.yaml
+++ b/apis/modelcaches/definition.yaml
@@ -1,0 +1,110 @@
+apiVersion: apiextensions.crossplane.io/v2
+kind: CompositeResourceDefinition
+metadata:
+  name: modelcaches.modelplane.ai
+spec:
+  group: modelplane.ai
+  names:
+    categories: [crossplane, modelplane]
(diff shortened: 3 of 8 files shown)


apis/inferenceclusters/definition.yaml
```diff
@@ -81,6 +81,15 @@ spec:
...
                             maxLength: 253
+                      cacheStorageClassName:
+                        type: string
+                        description: >-
+                          Name of the StorageClass to use for ModelCache
+                          PVCs (must support ReadWriteMany dynamic
+                          provisioning). When unset, Modelplane expects a
+                          StorageClass named `modelplane-rwx` to exist on
+                          the workload cluster.
+                        minLength: 1
```
negz: What're the chances we'd ever want other cache related settings on IC? I ask because I wonder if this should be:

```yaml
cache:  # Or modelCache, weightCache etc
  storageClassName: foo
```

In part to leave room for other cache knobs, and in part because my first thought when I saw `cacheStorageClass` was to go look for a new `CacheStorageClass` type. It took me a minute to realize it was "the `StorageClass` for caches" rather than a new type.

design/design.md
```diff
@@ -488,6 +488,73 @@ spec:
           - '--kv-transfer-config={"kv_role":"kv_producer"}'
 ```
 
+### ModelCache
```
negz: I could be convinced this deserves its own one-pager - the first thing "added" to the base design. Happy either way though.

design/design.md
```diff
@@ -488,6 +488,73 @@ spec:
...
+
+The cache's PVC backend is picked per cluster. For Modelplane-provisioned
+clusters (`cluster.source: GKE` today), Modelplane provisions a Filestore
+Enterprise instance with multishare on first ModelCache creation, composes a
+StorageClass named `modelplane-rwx` against it, and marks that StorageClass
+as the cluster's default — RWX is the right default for an inference cluster.
+Subsequent caches share the same Filestore instance. For existing clusters
+(`cluster.source: Existing`), the admin creates an RWX StorageClass on the
+workload cluster; Modelplane uses the convention name `modelplane-rwx` unless
+overridden, and doesn't touch the cluster's default class annotation.
```
negz: If empty `cacheStorageClassName` means "use the default" (as opposed to "don't use cache") perhaps it should default to a `StorageClass` named `default` or similar?

design/design.md
```diff
@@ -488,6 +488,73 @@ spec:
...
+      authSecret:
+        name: hf-token
+      sizeGiB: 1500
+```
+
+The cache mounts at `mountPath` (default `/mnt/models`) on every consuming
+pod; the engine container's model arg should reference this path. This keeps
+Modelplane out of engine-specific arg rewriting — different engines take
+different flags (e.g. `--model=` for vLLM, `--model-path=` for SGLang) and
+users pass whatever their engine expects.
```
negz: So users need to be careful to specify the model path correctly?

That seems reasonable compared to e.g. having modelplane try to mutate flags. I wonder though if we should be more opinionated about the mount path to make it more predictable for users.

negz (approved):
A few questions and comments but overall LGTM.

design/design.md
dennis-upbound: good idea to stamp out the pattern with this one. let me do! and I can format the PR message etc to follow any conventions we want.

apis/inferenceclusters/definition.yaml
dennis-upbound: this makes a lot of sense and yes worth breaking out because I think there will be more settings

design/design.md
dennis-upbound: Makes sense - I'll drop mount path too until there's a compelling case for it

design/design.md
dennis-upbound: ah yes that makes it much clearer. let me fix

PR #89, opened 2026-05-27
Switch control plane gateway from Envoy Gateway to Traefik

### LINKED ISSUE #85: `compose-model-service` puts URLRewrite filter under backendRef, breaking the data path
### What happened?

Every request through a `ModelService` returns HTTP 500 from the control plane gateway. The model serving pod on the workload cluster is healthy and never receives the request.

`compose-model-service` builds an HTTPRoute that attaches a `URLRewrite` filter to each `backendRef`:

https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-service/function/fn.py#L181-L192

The Gateway API spec only allows `RequestHeaderModifier` and `ResponseHeaderModifier` at the backendRef level. `URLRewrite` is a rule-level-only filter. Envoy Gateway rejects the route accordingly:

```
status:
  parents:
  - conditions:
    - type: ResolvedRefs
      status: "False"
      reason: UnsupportedRefValue
      message: Specific filter is not supported within BackendRef, only
        RequestHeaderModifier and ResponseHeaderModifier are supported
```

With the route unresolved, Envoy emits a `direct_response` for every matching request:

```
{
  "method": "POST",
  "x-envoy-origin-path": "/ml-team/qwen/v1/chat/completions",
  "response_code": "500",
  "response_code_details": "direct_response",
  "route_name": "httproute/ml-team/qwen-3e76dee66d25/rule/0/match/0/*"
}
```

The function comment says per-backend filters are intentional, to support endpoints with different `rewritePath` values:

> Each backendRef carries its own URLRewrite filter so endpoints with different rewritePaths (e.g. composed replicas vs. external SaaS providers) are rewritten correctly per-backend.

But Gateway API doesn't allow this shape.

### How can we reproduce it?

Walk through `docs/getting-started.md` to deploy `qwen-demo` on GKE. Once `ModelDeployment` reports `READY=True`, hit the endpoint:

```bash
address=$(kubectl get ms qwen -n ml-team -o jsonpath='{.status.address}')
kubectl run curl-test --image=curlimages/curl --restart=Never \
  --command -- curl -sS -i "$address/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen2.5-0.5B-Instruct","messages":[{"role":"user","content":"hi"}]}'
kubectl wait --for=jsonpath='{.status.phase}'=Succeeded pod/curl-test
kubectl logs curl-test
```

Response: `HTTP/1.1 500 Internal Server Error`.

Then inspect the composed HTTPRoute:

```bash
kubectl get httproute -n ml-team -o yaml
```

The `ResolvedRefs` condition reports `UnsupportedRefValue` with the message above.

Modelplane version: `v0.1.0-dev.1779486230.g401e962-dirty` (built from #83)
Crossplane version: v2.3.1
Inference backend: KServe v0.16.0
Cloud provider: GKE (g2-standard-8, NVIDIA L4)
Kubernetes version: control plane v1.34.0 (kind), workload v1.35.3-gke.1389000
provider-helm: v1.2.4
provider-kubernetes: v1.2.5

### LINKED ISSUE #88: Support FQDNs in `ModelEndpoint`

Currently `ModelEndpoint` composes Envoy backends that only support IP addresses. We need to support FQDNs as well - especially for a `ModelEndpoint` that points to e.g. Together AI or some other external URL.

What it says on the tin. Support FQDNs in `ModelEndpoint`

negz:
Fixes #85
Fixes https://github.com/modelplaneai/modelplane/issues/88

Envoy Gateway doesn't support per-backendRef URLRewrite filters. This is a Gateway API Extended feature that allows each backend in a weighted traffic split to have its own path rewrite. Modelplane needs this to route across endpoints with different path conventions -- for example a self-hosted model serving at /v1/ alongside Groq at /openai/v1/.

The limitation is in Envoy itself: WeightedCluster.ClusterWeight doesn't support prefix_rewrite. See envoyproxy/gateway#7099. Traefik is the only confirmed Gateway API implementation that supports this feature.

This PR replaces Envoy Gateway with Traefik Proxy on the control plane. Workload clusters remain on Envoy Gateway. The HTTPRoute now uses a single rule with per-backendRef URLRewrite filters instead of grouping endpoints by rewritePath into separate rules.

The Envoy Gateway Backend CRD is replaced with standard Kubernetes Service plus EndpointSlice resources. Both IP-based endpoints (workload cluster gateway addresses) and FQDN-based endpoints (external SaaS providers) use the same pattern. ExternalName Services aren't an option because Traefik's Gateway API provider explicitly rejects them.


apis/inferencegateways/definition.yaml
```diff
@@ -33,17 +33,20 @@ spec:
               backend:
                 type: string
                 description: >-
-                  Gateway implementation. MVP supports EnvoyGateway only.
-                enum: [EnvoyGateway]
-              envoyGateway:
+                  Gateway implementation. Traefik is used because it
+                  supports per-backendRef URLRewrite filters, which
+                  Modelplane needs to route across endpoints with
+                  different path conventions.
```
negz: I don't think we need an explanation of why Traefik here. It reads strangely to have an enum but then frame it as the one required option.

functions/compose-inference-gateway/function/fn.py
```diff
@@ -197,59 +205,72 @@ def compose_metallb(self):
...
-            self.rsp.desired.resources["envoy-gateway"],
+            self.rsp.desired.resources["traefik"],
             _helm_release(
-                chart="gateway-helm",
-                repo="oci://docker.io/envoyproxy",
-                version=eg.version if eg else "v1.3.0",
-                namespace="envoy-gateway-system",
+                chart="traefik",
+                repo="https://traefik.github.io/charts",
+                version=t.version if t else "40.2.0",
```
negz: Should these be constants?

Why do we default to 40.2.0? Should version just be required?

functions/compose-model-endpoint/function/fn.py
```diff
@@ -1,14 +1,16 @@
...
-backend name from status.routing.backendName to build its HTTPRoute.
+spec.url and composes a headless Service and EndpointSlice on the control
+plane pointing at the URL's host:port. ModelService reads the resulting
+service name from status.routing.backendName to build its HTTPRoute.
 
-External / SaaS endpoints (fqdn-style Backends) are deferred. For now,
-spec.url is expected to be an http://<ip>:<port>/... shape.
+For IP-based URLs (e.g. workload cluster gateways), the EndpointSlice uses
+addressType IPv4. For FQDN-based URLs (e.g. external SaaS providers), an
+ExternalName Service is composed instead.
```
negz: Is this still accurate?

functions/compose-model-endpoint/function/fn.py
```diff
@@ -80,42 +95,139 @@ def parse_url(self):
...
+                },
+            )
+
+    def compose_fqdn_backend(self, host: str, port: int):
+        """Compose a selectorless Service and EndpointSlice for an
+        FQDN-based endpoint (e.g. an external SaaS provider like
+        Together or Groq).
+
+        ExternalName Services are explicitly rejected by Traefik's
+        Gateway API provider (see kubernetes.go in the Traefik source).
```
negz: If we're going to reference source code it should be a GH link pinned to a commit and line.

functions/compose-model-endpoint/function/fn.py
```diff
@@ -80,42 +95,139 @@ def parse_url(self):
...
+                    },
+                    "addressType": "IPv4",
+                    "ports": [{"name": "", "port": port, "protocol": "TCP"}],
+                    # Traefik's Gateway API provider explicitly checks
+                    # Conditions.Ready and skips endpoints where it's
+                    # nil. The Kubernetes spec says nil should be
+                    # interpreted as "true", but Traefik doesn't honor
+                    # that — so we set it explicitly. See
+                    # pkg/provider/kubernetes/gateway/kubernetes.go in
+                    # the Traefik source.
```
negz: If we're going to reference source code it should be a GH link pinned to a commit and line.

negz (commented):
Self-review of LLM-assisted code.

functions/compose-model-endpoint/function/fn.py
```diff
@@ -80,42 +95,139 @@ def parse_url(self):
...
             {
-                "apiVersion": "gateway.envoyproxy.io/v1alpha1",
-                "kind": "Backend",
-                "metadata": {"namespace": self.xr.metadata.namespace},
+                "apiVersion": "v1",
+                "kind": "Service",
+                "metadata": {"namespace": ns},
                 "spec": {
-                    "endpoints": [{"ip": {"address": host, "port": port}}],
+                    "ports": [{"port": port, "protocol": "TCP"}],
```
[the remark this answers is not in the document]
negz: Turns out Traefik has code that handles this automatically and the defaults seem to work well.

https://github.com/traefik/traefik/blob/v3.7.1/pkg/provider/kubernetes/gateway/httproute.go#L933

functions/compose-model-endpoint/function/fn.py
```diff
@@ -80,42 +95,139 @@ def parse_url(self):
...
+        # Headless Service (no selector, no ClusterIP).
+        resource.update(
+            self.rsp.desired.resources[SERVICE_RESOURCE_KEY],
+            {
+                "apiVersion": "v1",
+                "kind": "Service",
+                "metadata": {"namespace": ns},
+                "spec": {
+                    "clusterIP": "None",
+                    "ports": [{"port": port, "protocol": "TCP"}],
```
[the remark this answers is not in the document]
negz: It turns out Traefik does in fact infer HTTPS from the port.

dennis-upbound (approved):
lgtm

PR #98, opened 2026-05-29
Add EKS support for inference clusters

### LINKED ISSUE #92: Support EKS for cluster provisioning

`InferenceCluster` currently only supports [GKE for cluster provisioning](https://github.com/modelplaneai/modelplane/blob/main/apis/inferenceclusters/definition.yaml#L41). Teams running on AWS can't have Modelplane provision a cluster for them — they'd have to fall back to BYO cluster via `source: Existing` (#3).

Add `EKS` as a value for `spec.cluster.source` on `InferenceCluster`. Same pattern as GKE: a new internal `EKSCluster` XR composed of an EKS cluster plus GPU node groups via the AWS provider, and the cluster composition function dispatches on `source` to delegate to it. The rest of the stack (KServe, etc.) is provider-agnostic — it just needs a kubeconfig.

negz:
Fixes #92

Modelplane could provision inference clusters on GKE but not on AWS. This adds an EKSCluster XR (Namespaced, group infrastructure.modelplane.ai) and its composition function, plus an EKS source on InferenceCluster and an EKS provisioning provider on InferenceClass that mirror the existing GKE shapes. compose-inference-cluster dispatches on spec.cluster.source: for EKS it composes an EKSCluster XR from the region and the GPU pools' InferenceClass provisioning blocks, waits for it to be ready, then wires its kubeconfig into the KServe backend. The GKE and Existing paths are unchanged.

The EKSCluster function composes a VPC with one public subnet per AZ, internet gateway and routing, cluster and node IAM roles, the EKS cluster, system and GPU node groups, and the vpc-cni / kube-proxy / coredns addons. It also installs the NVIDIA device plugin on GPU nodes, which EKS does not provide automatically the way GKE does.

GPU node groups are tainted nvidia.com/gpu=true:NoSchedule. GPU pods still schedule because EKS enables the ExtendedResourceToleration admission controller, which injects the matching toleration. Cross-XR teardown ordering, where the KServe backend's Helm releases must uninstall while the cluster is still reachable, relies on foreground cascading deletion of the InferenceCluster. The cleanup docs cover it, and it's the same shape as the GKE path.

Validated manually end to end against real AWS: provisioning, kubeconfig auth, GPU scheduling, stack install, model inference through the gateway, and clean foreground teardown.


apis/eksclusters/definition.yaml
```diff
@@ -0,0 +1,229 @@
...
+      type: string
+      jsonPath: .spec.region
+    schema:
+      openAPIV3Schema:
+        description: >-
+          An EKSCluster provisions an EKS cluster with dedicated node groups
+          for GPU inference and system workloads. It outputs a Secret
+          containing the cluster kubeconfig that consumers use to target
+          the cluster. The kubeconfig embeds a static bearer token that the
+          AWS provider refreshes; no separate IAM credential is emitted.
```
negz: Drop "no separate IAM credential is emitted". It's not interesting except in comparison to GKE.

apis/eksclusters/definition.yaml
```diff
@@ -0,0 +1,229 @@
...
+                          type: integer
+                          default: 1
+                          description: Number of GPUs per node.
+                          minimum: 1
+                          maximum: 16
+                        memory:
+                          type: string
+                          description: >-
+                            Per-GPU VRAM (e.g. "80Gi"). Reported on the
+                            consuming InferenceCluster's status and used
```
negz: Is this all just informational, or is it actually used by the EKS cluster?

If it's just informational for scheduling to InferenceClusters, does it need to even exist at the EKSCluster XR level?

functions/compose-eks-cluster/function/fn.py
```diff
@@ -0,0 +1,755 @@
...
+from models.io.upbound.m.aws.eks.clusterauth import v1beta1 as clusterauthv1beta1
+from models.io.upbound.m.aws.eks.nodegroup import v1beta1 as ngv1beta1
+from models.io.upbound.m.aws.iam.role import v1beta1 as rolev1beta1
+from models.io.upbound.m.aws.iam.rolepolicyattachment import v1beta1 as rpav1beta1
+
+# System node group injected into every EKS cluster to host control-plane
+# components (Envoy Gateway, KEDA, KServe controller, etc.). Not part of
+# the user-facing API — compose-inference-cluster only passes GPU groups.
+_SYSTEM_POOL_NAME = "system"
+_SYSTEM_POOL_INSTANCE_TYPE = "m6i.xlarge"
```
negz, line 44: Why this instance type for the system pool?

functions/compose-eks-cluster/function/fn.py
```diff
@@ -0,0 +1,755 @@
...
+_ADDONS = ("vpc-cni", "kube-proxy", "coredns")
+
+# NVIDIA device plugin. The GPU node groups run the AL2023 NVIDIA AMI,
+# which ships the driver and container toolkit but not the Kubernetes
+# device plugin. Without the plugin, nodes never advertise the
+# nvidia.com/gpu resource and GPU pods stay Pending. Managed clusters
+# like GKE install this automatically; EKS does not, so we install it.
+# It runs as a DaemonSet on GPU nodes (selected by the GPU pool label)
+# and tolerates the GPU taint those nodes carry.
+_DEVICE_PLUGIN_IMAGE = "nvcr.io/nvidia/k8s-device-plugin:v0.17.1"
```
negz: Why this version? v0.19.2 is latest.

functions/compose-inference-cluster/function/fn.py
```diff
@@ -185,6 +189,47 @@ def compose_gke(self, gke):
...
 
+    def compose_eks(self, eks):
+        """Compose an InferenceCluster backed by a Modelplane-provisioned
+        EKS cluster. Composes the EKSCluster XR, waits for it to be ready,
+        then wires its kubeconfig into the backend.
+
+        EKS does not need an identity secret: the kubeconfig from
+        ClusterAuth contains a static bearer token that the AWS provider
+        refreshes periodically, and the cluster grants the AWS provider's
+        principal cluster-admin via bootstrapClusterCreatorAdminPermissions.
```
negz: My hunch is only GKE _will_ need an identity secret. So it's more interesting for GKE to say it needs it than for other clusters (like EKS) to say they don't. However the rest of the description of how auth works is useful.

functions/compose-inference-cluster/function/fn.py
```diff
@@ -185,6 +189,47 @@ def compose_gke(self, gke):
...
+
+        eks_ready = resource.get_condition(self.req.observed.resources.get("eks-cluster"), "Ready").status == "True"
+        kubeconfig = self.observed_eks_secret(_SECRET_TYPE_KUBECONFIG)
+        cpc_exists = "cluster-provider-config-kubernetes" in self.req.observed.resources
+        backend_exists = BACKEND_RESOURCE_KEY in self.req.observed.resources
+
+        if (eks_ready and kubeconfig) or cpc_exists:
+            self.compose_cluster_provider_config(
+                kubeconfig.name if kubeconfig else "",
+                kubeconfig.key if kubeconfig else "",
```
negz: What'll `compose_cluster_provider_config` do if we hit these `else ""` cases?

negz (commented):
Self-review of AI generated code.

dennis-upbound (approved):

Issue #90, opened 2026-05-27
Support weighted traffic splitting across endpoint selectors


negz:
### What problem are you facing?

ModelService selects endpoints by label and load-balances across them equally. Each matched endpoint gets `weight: 1` in the composed HTTPRoute's backendRefs. This works for the common case where all endpoints are interchangeable, but it doesn't support scenarios like canary rollouts or gradual migration between providers.

For example, if a team is migrating from a self-hosted deployment to an external SaaS provider (or rolling out a new model version), they'd want to send 80% of traffic to one set of endpoints and 20% to another. Today the only lever is the number of endpoints — which conflates scaling with traffic policy.

### How could Modelplane help solve your problem?

Add an optional `weight` field to each `spec.endpoints[]` entry in ModelService. The weight would apply to the selector as a group — all endpoints matched by a selector collectively receive traffic proportional to that selector's weight.

```yaml
apiVersion: modelplane.ai/v1alpha1
kind: ModelService
metadata:
  name: qwen
spec:
  endpoints:
  - weight: 80
    selector:
      matchLabels:
        modelplane.ai/deployment: qwen-prod
  - weight: 20
    selector:
      matchLabels:
        modelplane.ai/deployment: qwen-canary
```

The interesting part is mapping group-level weights to per-backendRef weights in the HTTPRoute. Gateway API only supports per-backendRef weights, so compose-model-service would need to distribute each group's weight across its matched endpoints. For example, if the first selector matches 3 endpoints with group weight 80, those endpoints could get weights [27, 27, 26] — preserving the 80:20 ratio between groups.

An edge case worth considering is when a group's weight is smaller than its endpoint count (e.g., weight 1 across 5 endpoints). Some endpoints would get weight 0, which in Gateway API means no traffic. Scaling all weights up by a common factor would avoid this.

When `weight` is omitted, it should default to 1.

Issue #8, opened 2026-03-30
Explore inference-aware routing on the control plane gateway


negz:
### What problem are you facing?

The control plane's InferenceGateway uses basic path-based HTTPRoute routing across model placements. The [Gateway API Inference Extension](https://gateway-api.sigs.k8s.io/geps/gep-3567/) adds `InferenceModel` and `InferencePool` CRDs to Envoy Gateway with features like criticality-based priority and weighted traffic splitting between backends.

### How could Modelplane help solve your problem?

The ModelDeployment function could compose `InferenceModel` and `InferencePool` resources on the control plane instead of (or alongside) plain HTTPRoutes. This would give us:

- **Criticality-based priority** — Critical requests preempt BestEffort under load, useful for platform teams that want to prioritize production traffic over experimentation.
- **Weighted traffic splitting** — Route a percentage of traffic to specific environments, enabling canary deployments across placements.

One design constraint is that model-name routing (single endpoint, model name in the request body selects the backend) doesn't fit cleanly in a multi-tenant platform — two teams deploying the same model would have the same model name but need separate endpoints. The current path-based approach (`/<namespace>/<deployment>/v1/chat/completions`) handles multi-tenancy correctly.

PR #99, opened 2026-06-02
Drop KServe — dispatch to native + llm-d (Dynamo-ready)

### LINKED ISSUE #65: Drop the KServe dependency - compose to llm-d and Dynamo for multi-pod serving
Modelplane is a fleet-level control plane. Many of the features built by cluster-level orchestrators — routing, scheduling, caching, autoscaling — Modelplane does at the fleet level, i.e across clusters. What workload clusters are responsible for is serving: single-pod deployments, and multi-pod deployments with advanced topologies (LeaderWorker, disaggregated prefill/decode).

This issue proposes that Modelplane drop its KServe dependency and compose directly to **llm-d** (default for multi-pod cases) and **NVIDIA Dynamo** (for NVIDIA-optimized multi-pod with distinctive features). The rest — routing, scheduling, caching, autoscaling — happens at the Modelplane fleet layer.

This is primarily about getting the layering right. Feedback welcome before we land the composition logic in v0.1.

## What Modelplane does at the fleet level

Modelplane's primary concerns are above the cluster:

**Scheduling.** Picks `(cluster, pool)` per replica based on declared capabilities and capacity across the fleet. Replicas of the same deployment can land on different clusters depending on placement policy.

**Routing.** A fleet-aware gateway (Envoy-based) routes requests across replicas. Replicas can span clusters; the gateway watches ModelEndpoint health and routes accordingly. ModelService rules compile to gateway configuration.

**Caching.** model weight caching should be a fleet-aware primitive for staging weights to clusters and pools.

**Autoscaling.** KEDA on the scale subresource of ModelDeployment, fleet-wide. Each ModelReplica is a fixed-topology serving instance; scaling adds or removes whole instances.

These are above the cluster. Within a cluster, Modelplane only needs the workload to serve.

## What workload clusters need to do

Workload clusters need to run inference. Two cases:

**Single-pod.** One pod, possibly with multiple GPUs, running the engine container. This is just Kubernetes — Deployment + Service. No specialized orchestrator needed.

**Multi-pod with advanced topologies.** Multiple coordinated pods forming one logical inference unit: tensor or pipeline parallelism across nodes, disaggregated prefill/decode with KV cache transfer, expert parallelism. This needs a multi-pod orchestrator.

## Where KServe fits — and why we don't need it

KServe is a comprehensive ML serving platform. For LLMs specifically, it provides LLMInferenceService (introduced in v0.16), which is built on llm-d's architecture — the substantive multi-pod orchestration work happens in llm-d.

Looking at what KServe LLMInferenceService brings beyond llm-d:
- Lifecycle management (Modelplane handles at fleet layer)
- Storage initializer integration (future model caching will also be handled at the fleet layer in v0.2 #66)
- Autoscaling hooks, Knative integration (Modelplane uses KEDA at fleet layer)
- The LLMInferenceService CRD as a stable user-facing API (Modelplane composes; users don't write LLMInferenceService directly)

For Modelplane, all of these are either handled at the fleet layer or wrapped behind our own abstractions. The wrapper KServe provides on top of llm-d adds redundant layering — two reconciliation loops doing similar work, two lifecycle managers, two autoscaling integrations.

## Where Dynamo fits

NVIDIA Dynamo is a multi-pod LLM orchestration project at roughly the same architectural layer as llm-d. Dynamo's distinctive features go beyond what llm-d offers:
- KV Block Manager (multi-tier KV cache offload to CPU/storage)
- ModelExpress (checkpoint restore, weight streaming over NVLink for fast replica startup)
- NVL72 topology awareness via Grove
- more

Dynamo's distributed runtime is NVIDIA-coupled in practice — the shipped runtime images are CUDA-targeted, and the distinctive features are tuned for NVIDIA hardware. Dynamo is the right backend for NVIDIA customers wanting these specific capabilities.

## Proposed composition rules

Three paths, dispatched by topology and engine:

| Topology | Composition |
|---|---|
| Single-pod (any GPU count, single node) | Native Kubernetes Deployment + Service |
| Multi-pod (LeaderWorker, Disaggregated) | llm-d by default. Dynamo when distinctive Dynamo features are needed and the engine is NVIDIA-compatible. |

Modelplane infers the path from `engine.name`, `engine.version`, requested topology, and matched hardware capabilities. Users describe what they want to run; Modelplane picks the lightest composition path that satisfies it.

dennis-upbound:
Fixes https://github.com/modelplaneai/modelplane/issues/65.

Removes the KServe dependency and composes serving directly.

**Status:** rebased onto latest `main`; native single-pod path validated end-to-end on a real GKE cluster (see below). Reworking the llm-d multi-pod path to be **gateway-agnostic / Traefik-compatible** — see *Routing approach* below.

## What changed
- **`compose-model-replica` is now a dispatcher** (`backends/{native,llmd,dynamo}.py`):
  - single self-contained pod → **native** Kubernetes `Deployment` + `Service` + `HTTPRoute`
  - multi-pod (`pipeline > 1`) → **llm-d** `LeaderWorkerSet` + `Service` + `HTTPRoute` (see *Routing approach*)
  - **Dynamo** stub (dormant in v0.1)
  - No user-facing backend field; KServe `LLMInferenceService` emission deleted.
- **`KServeBackend` XR → `ServingStack`**: installs the serving substrate (LeaderWorkerSet, Gateway API, cert-manager, Prometheus, Envoy Gateway); drops KServe + KEDA. Creates the `modelplane-system` namespace on the workload cluster.
- **In-cluster HTTPRoute strips the `/<ns>/<deployment>/` prefix** (URLRewrite) so the engine sees `/v1/...`.
- `compose-model-deployment` endpoint path reconciled with the new HTTPRoute convention.
- Weight loading: no `ModelCache` → engine fetches directly; documented in getting-started.

## Routing approach — GAIE vs Traefik (how we're doing it)
v0.1 does no inference-aware (KV-/load-aware) endpoint picking, so the GAIE **`InferencePool` v1 + EPP** machinery the llm-d path originally emitted isn't needed yet.

**Decision: drop `InferencePool`/EPP and route `HTTPRoute → Service`** (plain Gateway API — `Service` + `EndpointSlice`, no Envoy `Backend` CRD), exactly like the native path. The HTTPRoute attaches to the **workload** cluster's inference gateway (Envoy `inference-gateway`, installed by `ServingStack`); the Service selects the **LWS leader pods** (only leaders serve the OpenAI API for vLLM). llm-d now routes identically to native on any Gateway API impl, so the GAIE v1 CRD swap and the inference-extension CRD install are gone from this PR.

Two distinct layers, kept straight (an earlier version of this section conflated them):
- **Workload gateway** (Envoy `inference-gateway`, one per cluster): routes to a replica's serving pods. The dropped `InferencePool`/EPP did KV-/load-aware **pod** picking *here*. Reintroducing it is a workload-gateway concern — it needs a GAIE-conformant workload gateway (Envoy Gateway's `InferencePool` v1 support is unconfirmed; alternatively switch the workload gateway to Istio/agentgateway). **Traefik is not involved at this layer.**
- **Control-plane gateway** (Traefik `modelplane`): routes *across replicas/clusters* via weighted `backendRef`s. **[#8](https://github.com/modelplaneai/modelplane/issues/8)** (inference-aware routing on the control plane) and **[#90](https://github.com/modelplaneai/modelplane/issues/90)** (weighted splits) live here — Traefik does weights natively but is not GAIE-conformant, so any pod-level picking at this layer would need an in-path picker rather than a GAIE gateway extension.

## Multi-node bootstrap (Ray)
The LWS gang needs a Ray cluster spanning its pods (vLLM `--pipeline-parallel-size > 1`). Approach (matches the upstream LWS/vLLM/KServe convention; reuses the proven bootstrap from the closed PR #78):
- **Separate `leaderTemplate`/`workerTemplate`** with role-specific `command`s (no `if WORKER_INDEX==0` branch): leader runs `ray start --head` then execs the engine; worker runs `ray start --address=$LWS_LEADER_ADDRESS:6379 --block`.
- **`LWS_*` env (`LWS_WORKER_INDEX`/`LWS_LEADER_ADDRESS`/`LWS_GROUP_SIZE`) is the documented public interface.**
- **User `command` override = escape hatch**: an optional `command` on the curated Container (added in this PR, schemas regenerated) bypasses injection and runs verbatim on both templates — covers non-vLLM engines (e.g. SGLang's torch.dist) against the `LWS_*` contract. No init containers / no sidecar in v0.1.

## Live validation (real GKE)
Provisioned a fresh GKE cluster end-to-end and served Qwen2.5-0.5B single-pod via the **native** path — pod `1/1` on an L4, real OpenAI completions. Surfaced + fixed: URLRewrite prefix-strip, and the `modelplane-system` namespace creation. Re [#102](https://github.com/modelplaneai/modelplane/issues/102): this PR doesn't change `RoutingReady`-vs-`ModelReady` gating, so #102 stands separately.

## Revised follow-ups
1. **llm-d path → Service routing** (this PR): drop `InferencePool`/EPP + the inference-extension CRD install; emit `HTTPRoute → Service` selecting LWS leaders.
2. **Multi-node Ray bootstrap** (this PR): separate leader/worker templates + `LWS_*` contract + override escape hatch.
3. Live-validate multi-node (`pipeline: 2`) on a 2-GPU-node cluster.
4. Inference-aware routing → **[#8](https://github.com/modelplaneai/modelplane/issues/8)** (separate); weighted splitting → **[#90](https://github.com/modelplaneai/modelplane/issues/90)** (Traefik-native, no GAIE needed).


negz:
> The original llm-d path emitted a GAIE InferencePool v1 + EPP and pointed an HTTPRoute backendRef at the pool. That mechanism requires a GAIE-conformant, ext-proc-capable data plane (Envoy/Istio/agentgateway). But the control-plane gateway is now Traefik (Nic, https://github.com/modelplaneai/modelplane/commit/d0df0e7cdaa48333193d170d8afb8c038709a3ae — Traefik is the only Gateway API impl that does per-backendRef URLRewrite for weighted cross-endpoint splits), and Traefik is not GAIE-conformant — it can't consume an InferencePool backendRef or call an ext-proc EPP.

I'm just starting my review - but does this matter? llm-d is running on the InferenceCluster right? We only need to run Traefik on the InferenceGateway - i.e. on the modelplane control plane. It sits in front of whatever the InferenceCluster's stack uses for routing. I would've thought the InferenceCluster (and thus llm-d etc) would be blissfully unaware of Traefik (or whatever) sitting in front of it.

apis/modeldeployments/definition.yaml
```diff
@@ -140,6 +140,19 @@ spec:
                                   type: string
                                   minLength: 1
                                   description: Container image.
+                                command:
```
negz, line 145: Did you end up needing this? I'm curious what for.

apis/servingstacks/definition.yaml
```diff
@@ -1,37 +1,37 @@
...
     shortNames:
-    - ksb
+    - ss
   scope: Namespaced
   versions:
   - name: v1alpha1
     referenceable: true
     additionalPrinterColumns:
-    - name: KSERVE
+    - name: GAIE
```
negz: Do we use GAIA? 🤔 The PR description said it was dropped in favor of Traefik.

examples/platform/inference-cluster-gke.yaml
```diff
@@ -2,7 +2,7 @@
 #
 # Modelplane provisions the full GKE cluster (VPC, subnet, system pool,
 # GPU pools, service account, IAM bindings) and installs the inference
-# stack (cert-manager, Envoy Gateway, Prometheus, KEDA, LeaderWorkerSet).
+# stack (cert-manager, Envoy Gateway, Prometheus, LeaderWorkerSet, Gateway API Inference Extension).
```
negz: Nit: The rest of this is wrapped at 80 chars.

(Here and in a few other comments.)

functions/compose-model-deployment/function/fn.py
```diff
@@ -197,18 +194,18 @@ def compose_endpoints(self, matched):
...
 
         Replicas pinned to clusters that are currently unavailable (no
         gateway address) get no endpoint. Routing must not direct
         traffic at a dead backend. When the cluster recovers and its
         gateway address is observed again the endpoint will be composed
         on the next reconcile.
         """
-        llmis = resource.child_name(self.xr.metadata.name)
-        rewrite_path = f"/{_NAMESPACE_REMOTE}/{llmis}/"
+        deployment_name = self.xr.metadata.name
```
negz: Nit: Why break `self.xr.metadata.name` out into a var (used once AFAICT) but not `self.xr.metadata.namespace`? My preference would be avoid the single use var in both cases.

functions/compose-model-replica/function/backends/native.py
```diff
@@ -0,0 +1,129 @@
...
+class NativeBackend:
+    def build(
+        self,
+        replica: v1alpha1.ModelReplica,
+        cluster: icv1alpha1.InferenceCluster,
+        deployment_name: str,
+    ) -> dict[str, k8sobjv1alpha1.Object]:
+        engine = base.engine_container(replica)
+        pc = cluster.status.providerConfigRef.name
+        name = resource.child_name(deployment_name)
```
negz: Could this cause a collision if multiple replicas land on the same IC? They can't with the scheduler today, but we do intend to allow it in v0.1.

functions/compose-model-replica/function/backends/native.py
```diff
@@ -0,0 +1,129 @@
...
+            "metadata": {"name": name, "namespace": _REMOTE_NAMESPACE},
+            "spec": {
+                "parentRefs": [{"name": "inference-gateway", "namespace": "modelplane-system"}],
+                "rules": [
+                    {
+                        "matches": [
+                            {
+                                "path": {
+                                    "type": "PathPrefix",
+                                    "value": f"/{replica.metadata.namespace}/{deployment_name}/",
```
negz: Is there a potential collision here too (N replicas of the same MD on the same IC)?

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+            "metadata": {"name": name, "namespace": _REMOTE_NAMESPACE},
+            "spec": {
+                "parentRefs": [{"name": "inference-gateway", "namespace": "modelplane-system"}],
+                "rules": [
+                    {
+                        "matches": [
+                            {
+                                "path": {
+                                    "type": "PathPrefix",
+                                    "value": f"/{replica.metadata.namespace}/{deployment_name}/",
```
negz: Same concern WRT collisions here.

functions/compose-model-replica/function/backends/base.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Dynamo is dormant in v0.1: no Dynamo-only capability is wired, so a
+    multi-pod replica always selects llm-d.
+    """
+    if not needs_cross_pod_coordination(replica):
+        return NATIVE
+    return LLMD
+
+
+class Backend(Protocol):
```
negz, line 67: TIL about `typing.Protocol.`

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+Selected only for multi-node replicas (pipeline > 1), so this always renders a
+LeaderWorkerSet whose gang size is the per-worker node count.
+
+Routing is plain Gateway API — `HTTPRoute -> Service`, exactly like native.py —
+NOT a GAIE `InferencePool`. The control-plane gateway is Traefik (which is not
+GAIE-conformant: it can't consume an `InferencePool` backendRef or call an
+ext-proc endpoint-picker), so the llm-d path routes to a Service that selects
+the LWS *leader* pods (only the leader serves the OpenAI API; workers just join
+the gang). Inference-aware endpoint picking (KV-/load-aware) is tracked
+separately in issue #8 as a Traefik-compatible in-path picker.
```
negz: Same comment as on the PR description: is this accurate given Traefik runs at the layer above llm-d's routing? 🤔

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+
+from crossplane.function import resource
+from models.ai.modelplane.inferencecluster import v1alpha1 as icv1alpha1
+from models.ai.modelplane.modelreplica import v1alpha1
+from models.io.crossplane.m.kubernetes.object import v1alpha1 as k8sobjv1alpha1
+
+from function.backends import base
+
+# Namespace for serving workloads on remote clusters.
+_REMOTE_NAMESPACE = "default"
```
negz, line 46: I wonder if we should put stuff in endpoint-name derived namespaces on remote clusters. (Not a blocker for this PR, just thinking out loud.)

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+class LLMDBackend:
+    def build(
+        self,
+        replica: v1alpha1.ModelReplica,
+        cluster: icv1alpha1.InferenceCluster,
+        deployment_name: str,
+    ) -> dict[str, k8sobjv1alpha1.Object]:
+        engine = base.engine_container(replica)
+        pc = cluster.status.providerConfigRef.name
+        name = resource.child_name(deployment_name)
```
negz: Another potential collision?

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+                c["ports"] = [{"containerPort": _ENGINE_PORT}]
+                c["readinessProbe"] = {
+                    "httpGet": {"path": "/health", "port": _ENGINE_PORT},
+                    "initialDelaySeconds": 30,
+                    "periodSeconds": 10,
+                }
+            return c
+
+        def pod_spec(c: dict) -> dict:
+            spec = {"containers": [c], "volumes": [{"name": "dshm", "emptyDir": {"medium": "Memory"}}]}
```
negz, line 158: What if the ModelDeployment specified containers other than `engine` (e.g. sidecars)? Should we support that here, or does it not make sense in the context of a LeaderWorkerSet?

functions/compose-model-replica/function/backends/llmd.py
```diff
@@ -0,0 +1,228 @@
...
+                        "matches": [
+                            {
+                                "path": {
+                                    "type": "PathPrefix",
+                                    "value": f"/{replica.metadata.namespace}/{deployment_name}/",
+                                }
+                            }
+                        ],
+                        # Strip the /<ns>/<deployment>/ routing prefix so the engine
+                        # (which serves /v1/...) sees the path it expects.
```
negz: Does this mean at the IG layer we rewrite to `/ns/deployment` but then we just strip it off here? 🤔

functions/compose-model-replica/function/backends/native.py
```diff
@@ -0,0 +1,129 @@
...
+                "periodSeconds": 10,
+            },
+        }
+        if engine.command:
+            container["command"] = list(engine.command)
+        if engine.env:
+            container["env"] = [e.model_dump(exclude_none=True) for e in engine.env]
+
+        pod_spec = {
+            "containers": [container],
```
negz, line 71: Same question WRT ModelDeployment having multiple containers.

functions/compose-model-replica/function/backends/native.py
```diff
@@ -0,0 +1,129 @@
...
+                        "matches": [
+                            {
+                                "path": {
+                                    "type": "PathPrefix",
+                                    "value": f"/{replica.metadata.namespace}/{deployment_name}/",
+                                }
+                            }
+                        ],
+                        # Strip the /<ns>/<deployment>/ routing prefix so the engine
+                        # (which serves /v1/...) sees the path it expects.
```
negz: Same question WRT the `/ns/deployment` rewrite potentially being a no-op.

functions/compose-model-replica/function/fn.py
```diff
@@ -136,131 +144,23 @@ def resolve_inputs(self):
...
-                    ),
-                ),
-            ),
-        )
+        """Dispatch to the backend that matches the replica's topology."""
+        self.engine = base.engine_container(self.xr)
+        backend = _BACKENDS[base.select_backend(self.xr)]()
+        deployment_name = self._deployment_name()
+        for key, composed in backend.build(self.xr, self.ic, deployment_name).items():
+            resource.update(self.rsp.desired.resources[key], composed)
```
negz, line 148: Nice, I like this clean pattern.

functions/compose-model-replica/tests/test_backends.py
```diff
@@ -0,0 +1,271 @@
...
+                    spec=v1alpha1.Spec(
+                        containers=[v1alpha1.Container(name="engine", image="img")],
+                    ),
+                ),
+            ),
+        ),
+    )
+
+
+class TestDispatch(unittest.TestCase):
```
negz: WDYT about making this test match the existing conventions `Case` table, etc. I could see the argument it doesn't make sense for backends, but this file feels like it introduces a second test style.

negz (commented):
Thanks @dennis-upbound! I like the backend dispatch pattern this introduces. I have a few questions though - and one concern WRT potential collisions if we allow multiple replicas per IC.

dennis-upbound:
Thanks for the thorough review @negz! Addressed in 9362db74:

- **Collisions (your main concern)** — backends named workload resources after the *deployment* (shared by all replicas). They're now named after the **replica** (unique per placement: `child_name(deployment, cluster)`), and the routing path is per-replica (`/<ns>/<replica>/`). `compose-model-deployment` emits a matching per-replica `rewritePath`/`url`, so N replicas of one deployment can co-exist on one IC without colliding. Also dropped the now-redundant `deployment_name` arg from the `Backend` protocol + all backends.
- **`servingstacks` GAIE references** — removed the `GAIE` printer column, the `spec.versions.gatewayApiInferenceExtension` field, and the "Inference Extension" wording, and regenerated schemas. Nothing read the field after the InferencePool path was dropped.
- **llm-d docstring / PR "Traefik" accuracy** — you're right, I'd conflated layers. Reworded: the model `HTTPRoute` attaches to the **workload** Envoy gateway; GAIE/inference-aware routing (and Traefik's non-conformance) is the **control-plane / #8** concern at the layer above. PR description updated too.
- **`/ns/deployment` rewrite then strip — no-op?** Not a no-op: the control plane rewrites the public `/<ns>/<service>/` to the replica's `/<ns>/<replica>/`, which must *survive* to the workload gateway so it can address the right replica among multiple deployments on that IC; the backend then strips it so the engine sees `/v1/`. Added a clarifying comment in both backends.
- **The `command` field** — it's the non-vLLM multi-node escape hatch: when set on the `engine` container it bypasses the built-in vLLM/Ray bootstrap and runs verbatim on both LWS templates, so engines like SGLang (`--nnodes/--node-rank/--dist-init-addr`, symmetric across pods) work against the `LWS_*` contract.
- **`test_backends.py` style** — refactored to the `Case`-table convention (selection + Dynamo stay focused, since they assert dispatch/raises, not manifests).
- **Nits** — gke example wrapped at 80; inlined the single-use var.

Deferred (TODO'd in code): **sidecars** (non-engine containers in the MD template) — out of v0.1 scope and non-trivial for the LWS gang; happy to file an issue. The endpoint-name-derived-namespaces idea I left as a thought for later.

dennis-upbound:
Correction to my note above on the GAIE/Traefik layering — I conflated two layers, want to set it straight:

The dropped `InferencePool`/EPP were composed on the **workload** cluster and consumed by the **workload Envoy gateway** (`inference-gateway`), doing KV-/load-aware **pod** picking there. So whether to keep them is a **workload-gateway** question (Envoy's `InferencePool` v1 support is unconfirmed; or switch the workload gateway to Istio/agentgateway) — **Traefik (the control-plane `modelplane` gateway) never sees them.** Saying "Traefik isn't GAIE-conformant, so drop the InferencePool" (and my softer "it's the #8/control-plane concern") was wrong.

Cleaner split:
- **Workload** (Envoy `inference-gateway`): pod-level KV-/load-aware picking — the dropped `InferencePool`'s job. Dropped for v0.1 simply because we don't need picking yet; reintroducing it is the workload-gateway-conformance question, no Traefik involvement.
- **Control plane** (Traefik `modelplane`): routing *across replicas* — this is #8, and weighted splits are #90 (Traefik does weights natively).

The code is unchanged (Service routing on the workload gateway is correct either way); I've fixed the wording in the `llmd.py` docstring and the PR description.

negz:
> out of v0.1 scope and non-trivial for the LWS gang; happy to file an issue

Please do. If we can't actually support multiple containers we shouldn't allow the MD to specify multiple containers. I think we do at the moment.

dennis-upbound:
Two things:

**On the Traefik/routing question** (your "does this matter? llm-d would be blissfully unaware of Traefik") — exactly right, and that's the correction I posted above. The InferenceCluster's stack (llm-d/Envoy `inference-gateway`) is unaware of Traefik; Traefik (`modelplane`) only fronts it on the control plane. My original wording wrongly tied the workload `InferencePool` drop to Traefik — the drop is purely a workload-gateway decision (v0.1 needs no KV-aware picking; Envoy `InferencePool` v1 conformance unconfirmed). Fixed in the docstring + PR description.

**On sidecars** — agreed, and good catch that the description even claimed they "pass through" (they were silently dropped). Filed #108. For this PR I've **constrained `containers` to `maxItems: 1`** (single `engine` container) in both the ModelDeployment and ModelReplica XRDs, so we no longer advertise multi-container support we don't implement. Relaxing it is the work in #108 (needs design for which containers land on the LWS leader vs workers).

negz (approved):

Issue #95, opened 2026-05-28
InferenceGateway never becomes ready on a fresh control plane: Gateway API CRDs not installed


negz:
### What happened?

The `InferenceGateway` composition installs Traefik (chart `40.2.0`, app `v3.7.1`) with the Kubernetes Gateway provider enabled, but nothing installs the Gateway API CRDs on the control plane. On a fresh control plane cluster (one that has never had Gateway API CRDs applied), the gateway never becomes ready.

Two distinct failures surface in sequence:

1. The Traefik Helm release fails to install because its rendered manifest contains a `GatewayClass`, but the CRD is absent:

   ```
   create failed: failed to install release: unable to build kubernetes objects from release manifest: resource mapping not found for name: "traefik" namespace: "" from "": no matches for kind "GatewayClass" in version "gateway.networking.k8s.io/v1"
   ensure CRDs are installed first
   ```

   Traefik's chart stopped shipping the Gateway API CRDs as of the v3.7 line — they must be installed separately (see the chart's own deprecation notice).

2. Even after installing the *standard* channel Gateway API CRDs (v1.2.1), Traefik v3.7.1 watches `v1.TLSRoute` and `v1.BackendTLSPolicy`, which the standard channel doesn't serve as `v1`. The Traefik pod logs a tight loop of:

   ```
   Failed to watch *v1.TLSRoute: the server could not find the requested resource (get tlsroutes.gateway.networking.k8s.io)
   Failed to watch *v1.BackendTLSPolicy: the server could not find the requested resource (get backendtlspolicies.gateway.networking.k8s.io)
   ```

   and the `GatewayClass` / `Gateway` stay `Accepted: Unknown` / `Programmed: Unknown`, so the `InferenceGateway` never goes ready.

The fix that works is to install the **experimental** channel Gateway API CRDs at **v1.5.1** (which serve `TLSRoute` and `BackendTLSPolicy` as `v1`), then restart Traefik. After that the GatewayClass is Accepted, the Gateway is Programmed, and the `InferenceGateway` becomes ready with its MetalLB address.

This is independent of the workload cluster's cloud — it reproduces on a plain kind control plane. It likely doesn't reproduce when reusing a control plane that already had Gateway API CRDs applied by a previous run, which may be why it isn't caught.

The composition should install a known-good Gateway API CRD bundle (experimental channel, version matching the Traefik appVersion) before the Traefik release, rather than relying on the CRDs being pre-present.

### How can we reproduce it?

[...]

PR #100, opened 2026-06-02
Fix InferenceGateway readiness and ModelEndpoint routing on fresh control planes

### LINKED ISSUE #96: ModelEndpoint routing fails: prerequisite ClusterRole missing services/endpoints/endpointslices RBAC
### What happened?

The `ModelEndpoint` composition composes a selectorless `Service` plus an `EndpointSlice` on the control plane to route the unified gateway to a remote model endpoint (see [`compose-model-endpoint/function/fn.py`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-endpoint/function/fn.py) — it builds a `Service` and a `discovery.k8s.io/v1` `EndpointSlice`).

But the `crossplane-compose-modelplane` ClusterRole in the getting started prerequisites doesn't grant Crossplane access to `services`, `endpoints`, or `endpointslices`. So the `ModelEndpoint` never becomes ready and the gateway can't route to the model:

```
cannot apply composed resource "endpointslice": endpointslices.discovery.k8s.io "qwen-demo-..." is forbidden: User "system:serviceaccount:crossplane-system:crossplane" cannot patch resource "endpointslices" in API group "discovery.k8s.io" in the namespace "ml-team"

system:serviceaccount:crossplane-system:crossplane is not allowed to [get, list, watch, create, update, patch, delete] resource endpointslices.discovery.k8s.io/v1 in namespace ml-team
```

The `ModelEndpoint` stays `SYNCED=False` / `Ready=False` with `Unready resources: service`, and the `ModelService` stays `READY=False`. Requests to the gateway return `no available server`.

Adding the missing rules unblocks it:

```yaml
- apiGroups: [""]
  resources: ["services", "endpoints"]
  verbs: ["*"]
- apiGroups: ["discovery.k8s.io"]
  resources: ["endpointslices"]
  verbs: ["*"]
```

After granting these, the `Service` and `EndpointSlice` are created, the `ModelEndpoint` goes `RoutingReady=True`, and inference flows through the control plane gateway.

This is cloud-agnostic — it affects any `ModelDeployment` that produces a `ModelEndpoint`, regardless of where the workload cluster runs. The getting started guide's prerequisite ClusterRole should include these rules (alongside the existing gateway/metallb/usage grants).

### How can we reproduce it?

1. Stand up a control plane and at least one ready `InferenceCluster` per the getting started guide
2. `kubectl apply -f examples/deployment/model-deployment.yaml` and `model-service.yaml`
3. `kubectl describe modelendpoint -n ml-team <name>` shows the `endpointslices` forbidden error above
4. `kubectl get ms -n ml-team` stays `READY=False`; the gateway returns `no available server`

Modelplane version: `ecks` branch built from source (`compose-model-endpoint` and the prerequisite RBAC unchanged from `main`)

* Crossplane: v2.3.1
* Inference backend: KServe
* Control plane: kind, Kubernetes v1.34.0
* Workload cluster: EKS (but not cloud-specific)

negz:
Fixes #95
Fixes #96

On a fresh control plane, two independent gaps stop the unified inference gateway from serving traffic.

**Gateway API CRDs are never installed (#95).** The InferenceGateway composition enables Traefik's Kubernetes Gateway provider, but nothing installs the Gateway API CRDs. Traefik's chart doesn't ship them and renders a `GatewayClass` even when `gateway.enabled` is false, so the release fails with `no matches for kind "GatewayClass"`. The composition now composes the standard-channel v1.5.1 CRDs directly onto the control plane before Traefik, vendored from upstream and loaded with PyYAML. Traefik is gated on the CRDs reporting `Established`, its chart `GatewayClass` is disabled in favour of the composed one, and Usages order the GatewayClass and Gateway ahead of the Traefik release so the controller can clear their finalizers on teardown. The standard channel at v1.5.1 serves `TLSRoute` and `BackendTLSPolicy` as `v1`, so the experimental channel suggested in #95 isn't needed — that was a consequence of first trying the older v1.2.1 CRDs.

**ModelEndpoint routing RBAC is missing (#96).** The ModelEndpoint composition composes a selectorless `Service` and a `discovery.k8s.io` `EndpointSlice` to route the gateway to a remote endpoint, but the prerequisite ClusterRole didn't grant `services` or `endpointslices`, so the gateway returned `no available server`. This adds the missing rules to both the getting started guide and the qwen-demo prerequisites.

A preparatory commit also switches `compose-kserve-backend` from hand-converted JSON CRDs to the vendored upstream YAML via PyYAML, the pattern the #95 fix reuses.


docs/getting-started.md
```diff
@@ -60,8 +60,9 @@ kind: Namespace
 metadata:
   name: modelplane-system
 ---
-# Grant Crossplane permissions to compose Gateway API, MetalLB, and Usage
-# resources. This ClusterRole is aggregated into Crossplane's role automatically.
+# Grant Crossplane permissions to compose Gateway API, MetalLB,
+# Service/EndpointSlice routing, and Usage resources. This ClusterRole is
```
negz: "and Usage resources" is stale - Crossplane has this by default.

examples/qwen-demo/00-prerequisites.yaml
```diff
@@ -1,9 +1,10 @@
 # Modelplane prerequisites. Apply once after installing Crossplane.
 #
 # These resources grant Crossplane and provider-helm the permissions
-# needed to compose Gateway API, Envoy Gateway, MetalLB, and Usage
-# resources. They cannot be self-composed because Crossplane needs the
-# permissions before it can compose anything.
+# needed to compose Gateway API, Envoy Gateway, MetalLB,
+# Service/EndpointSlice routing, and Usage resources. They cannot be
```
negz: Same here.

functions/compose-inference-gateway/function/fn.py
```diff
@@ -14,15 +14,45 @@
 envoyproxy/gateway#7099.
 """
 
+from pathlib import Path
```
negz: I'd like to avoid importing functions and types directly - so import `pathlib` and use `pathlib.Path`.

functions/compose-inference-gateway/function/fn.py
```diff
@@ -14,15 +14,45 @@
...
 
+_HERE = Path(__file__).parent
+
+# Gateway API CRDs (standard channel, v1.5.1) vendored from upstream:
+# https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.1/standard-install.yaml
+#
+# Traefik's Helm chart does not ship the Gateway API CRDs, and on a fresh
+# control plane nothing else installs them, so the Traefik release fails to
+# render its GatewayClass. We compose the CRDs directly onto the control
+# plane (they are plain Kubernetes resources, not managed resources) before
```
negz: > (they are plain Kubernetes resources, not managed resources)

This is obvious - not worth mentioning.

functions/compose-inference-gateway/function/fn.py
```diff
@@ -14,15 +14,45 @@
...
+# standard channel serves TLSRoute and BackendTLSPolicy as v1, which Traefik
+# watches.
+#
+# We install only the CustomResourceDefinitions. The bundle also ships a
+# ValidatingAdmissionPolicy ("safe-upgrades") that blocks installing
+# experimental-channel or pre-v1.5.0 CRDs cluster-wide. We don't compose it:
+# it would impose a cluster-scoped policy the user didn't ask for (e.g.
+# blocking a later switch to experimental CRDs for another controller), and
+# composing a policy that governs CRD writes alongside the very CRD writes it
+# governs is needlessly fragile.
```
negz: This logic feels flawed. The user didn't "ask for" these CRDs either - we make the decision about what to install to implement an InferenceGateway.

I'm not necessarily proposing we install the VAP, but I am challenging the stated reason for why we don't.

functions/compose-inference-gateway/function/fn.py
```diff
@@ -160,6 +198,36 @@ def compose_provider_config(self):
...
+        must exist before the Traefik release renders its resources and before
+        Traefik watches the Gateway API types.
+
+        They need no deletion protection. The only Gateway API objects on the
+        control plane are the GatewayClass and Gateway this function composes;
+        Traefik renders none itself (gateway and gatewayClass are disabled in
+        its chart). A CRD won't delete until its instances are gone, and those
+        instances are ordered ahead of the Traefik controller by their own
+        Usages (see compose_gateway_usages), so the CRDs are free to delete
+        last."""
```
negz: Is this comment actually valuable to the reader?

functions/compose-inference-gateway/function/fn.py
```diff
@@ -160,6 +198,36 @@ def compose_provider_config(self):
         )
         self.rsp.desired.resources["provider-config-helm"].ready = fnv1.READY_TRUE
 
+    def compose_gateway_api_crds(self):
+        """Compose the Gateway API CRDs onto the control plane.
+
+        These are composed unconditionally and directly (not via a provider),
```
negz: Why is this interesting?

functions/compose-inference-gateway/function/fn.py
```diff
@@ -160,6 +198,36 @@ def compose_provider_config(self):
...
+        last."""
+        for doc in _GATEWAY_API_CRDS:
+            key = _crd_key(doc)
+            resource.update(self.rsp.desired.resources[key], doc)
+            if self._crd_established(key):
+                self.rsp.desired.resources[key].ready = fnv1.READY_TRUE
+
+    def _crd_established(self, key):
+        """Whether a composed CRD reports Established (its schema is served)."""
+        return resource.get_condition(self.req.observed.resources.get(key), "Established").status == "True"
```
negz: Does this helper earn its place?

negz (commented):
Self-review of LLM generated code.

dennis-upbound (approved):
shipit!

PR #101, opened 2026-06-03
Match hardware as DRA-style devices and form DRA ResourceClaims

### LINKED ISSUE #56: Align hardware capabilities design with Kubernetes Dynamic Resource Allocation

### LINKED ISSUE #103: `nodeSelector` should describe hardware as a list of DRA devices

A `ModelDeployment` `spec.nodeSelector.cel` is a single CEL expression matched against a pool's merged attributes and capacity, written in a flat dialect: `attributes["gpu.nvidia.com/cudaComputeCapability"]...` and `capacity["gpu.nvidia.com/memory"]...`. The design says we'll translate it into a DRA `ResourceClaim` as "a straightforward split of `domain/name` into `device.attributes["domain"].name`."

Two problems.

**It looks like DRA but doesn't work like DRA.** The flat `attributes[...]`/`capacity[...]` shape, the `version(...)` wrapping, and the `version()` (vs DRA's `semver()`) naming all diverge from real DRA CEL. A user familiar with DRA will write `device.attributes["gpu.nvidia.com"].architecture` and `.isGreaterThan(semver("9.0.0"))` and be surprised when it doesn't work. The design already leans hard on DRA (KEP-4381, typed attributes, qualified keys), so the half-resemblance is the worst of both worlds.

**It can't describe more than one device.** A node can have a GPU, an InfiniBand NIC, and potentially other accelerators. In real DRA these are distinct devices, often from distinct drivers, and a `ResourceClaim` makes one request per device. You can't express "a GPU like X and a NIC like Y" in a single selector. Our flat model flattens everything onto one synthetic device, so a platform team can't fully describe an `InferenceClass`'s hardware, an ML team can't filter to a pool that has a specific GPU *and* NIC, and the ResourceClaim translation isn't actually mechanical for multi-device requirements.

The `modelplane.ai/*` fleet-attribute convention (e.g. `modelplane.ai/networkInterNode: infiniband`) was a workaround for this: inter-node networking had nowhere to live except a synthetic pool attribute, distinguished from real device attributes only by a key prefix.

GPUs are the only hardware real DRA drivers expose as devices today (NVIDIA's driver exposes `GPU` and `ComputeDomain`; InfiniBand/RDMA is still handled via SR-IOV/device-plugin, not DRA). So a NIC "device" is partly aspirational. But unlike DRA, we don't depend on a real driver publishing facts: the platform team authors the `InferenceClass` by hand. We can model InfiniBand as a device even where no DRA driver does. The constraint that replaces "did a driver publish this" is "is this device claimed via DRA, or synthetic and enforced only by our scheduler."

Make `InferenceClass` and `nodeSelector` describe hardware as a list of DRA-style devices, with CEL that *is* DRA CEL.

`InferenceClass` declares an array of devices, each with attributes and capacity (DRA's typed schema) and a driver. A `claim` discriminator says how the device is claimed: `DRA` (default; emitted as a request in a real `ResourceClaim`) or `Synthetic` (described for scheduling only, never claimed). It's an enum so future claim mechanisms (device plugin, extended resource) can be added without a breaking change.

```yaml
kind: InferenceClass
metadata:
  name: gke-h200-8x-a3-ib
spec:
  provisioning: { ... }   # unchanged
  devices:
  - name: gpu
    claim: DRA                      # default; emitted as a DeviceRequest in the ResourceClaim
    driver: gpu.nvidia.com
    count: 8
    attributes:
      architecture: { string: Hopper }
      cudaComputeCapability: { version: "9.0.0" }
    capacity:
      memory: { value: "141Gi" }
  - name: ib
    claim: Synthetic                # described for scheduling only; not in the ResourceClaim
    driver: nic.nvidia.com          # no real DRA driver yet; we author it anyway
    count: 8
    attributes:
      linkType: { string: infiniband }
    capacity:
      bandwidth: { value: "400Gi" }
```

This replaces the the `modelplane.ai/*` convention. The real/synthetic boundary becomes an explicit per-device `claim` discriminator instead of a key-prefix heuristic.

`nodeSelector` becomes a list of device requests, each with its own DRA CEL selectors and count, mirroring a `ResourceClaim`'s requests:

```yaml
# Minimal: one GPU, >= 80Gi.
nodeSelector:
  devices:
  - name: gpu
    selectors:
    - cel: device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("80Gi")) >= 0
```

```yaml
# Maximal: 8 Hopper GPUs >= 141Gi, and 8 InfiniBand NICs >= 400Gb/s.
nodeSelector:
  devices:
  - name: gpu
    count: 8
    selectors:
    - cel: |
        device.driver == "gpu.nvidia.com" &&
        device.attributes["gpu.nvidia.com"].architecture == "Hopper" &&
        device.attributes["gpu.nvidia.com"].cudaComputeCapability.isGreaterThan(semver("9.0.0")) &&
        device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("141Gi")) >= 0
  - name: nic
    count: 8
    selectors:
    - cel: |
        device.driver == "nic.nvidia.com" &&
        device.attributes["nic.nvidia.com"].linkType == "infiniband" &&
        device.capacity["nic.nvidia.com"].bandwidth.compareTo(quantity("400Gi")) >= 0
```

Each request has a `name`, as a DRA `DeviceRequest` does. It's required, so future additions that reference a request by name (e.g. `constraints`) stay additive.

`selectors` is a list, as in DRA, where each entry is a one-of (today just `cel`). A list of CEL selectors is ANDed, equivalent to joining with `&&`, so it adds no matching expressiveness on its own. We mirror the shape so a future non-CEL selector kind can be added as a new one-of member without a breaking change.

Inside `devices[].cel` it's real DRA CEL: `device.driver`, `device.attributes["domain"].name`, `device.capacity["domain"].name`, `quantity()`, `semver()`, version attributes pre-parsed as semver (no wrapping). A DRA user's expressions transfer verbatim, and translation is mechanical: one `claim: DRA` device becomes one `DeviceRequest` (count + selector); `claim: Synthetic` devices are dropped and enforced by our scheduler.

negz:
Fixes #56.
Fixes #103.

A `ModelDeployment` could pick a cluster by label but couldn't steer a replica to a node pool with the right hardware. Its `nodeSelector.cel` matched against a pool's flattened attributes in a dialect that resembled DRA CEL but didn't behave like it, and couldn't describe a node with more than one device. `InferenceClass` had no DRA-style way to declare a node's hardware, and worker pods bound GPUs through the legacy `nvidia.com/gpu` device plugin rather than DRA.

This makes hardware matching look and feel like DRA, end to end:

* `InferenceClass` advertises a pool's node's list of devices, like a DRA `ResourceSlice`
* `ModelDeployment` specifies a list of device requests, like a DRA `ResourceClaim`
* The scheduler uses these to match deployments to pools

An `InferenceClass` describes a pool's hardware as DRA devices. Each has a `driver`, a `count`, typed `attributes` and `capacity`, and a `claim` discriminator: `DRA` for hardware a real driver exposes and claims at admission, `Synthetic` for hardware that matters for placement but has no driver yet (an InfiniBand fabric, say).

```yaml
kind: InferenceClass
metadata:
  name: eks-l4-1x-g6
spec:
  provisioning:
    provider: EKS
    eks: { instanceType: g6.xlarge, accelerator: { type: nvidia-l4, count: 1 } }
  devices:
  - name: gpu
    claim: DRA
    driver: gpu.nvidia.com
    deviceClassName: gpu.nvidia.com
    count: 1
    attributes:
      architecture: { string: Ada Lovelace }
    capacity:
      memory: { value: "23034Mi" }   # real usable VRAM, not the nominal 24GB
```

A `ModelDeployment` requests devices in DRA CEL. The scheduler matches each request against a pool's devices and pins the replica to one that satisfies them. When a `ModelDeployment` requests a device the `InferenceClass` marks as `claim: DRA`, that request is passed through as the real workload's (i.e. Deployment's or LWS's) DRA `ResourceClaim`.

```yaml
kind: ModelDeployment
metadata: { name: qwen-demo, namespace: ml-team }
spec:
  replicas: 1
  nodeSelector:
    devices:
    - name: gpu
      count: 1
      selectors:
      - cel: device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("4Gi")) >= 0
  workers:
    topology: { tensor: 1 }
    template:
      spec:
        containers:
        - name: engine
          image: vllm/vllm-openai:v0.7.3
          args: ["--model=Qwen/Qwen2.5-0.5B-Instruct"]
```

`nodeSelector` is required, and at least one request must resolve to a claimable device.

A `ModelDeployment` fans out to `spec.replicas`, spread across clusters and packed onto fewer only when capacity forces it. A replica's identity is a `(cluster, index)` pair so co-located replicas don't collide. A replica is never re-homed: if its cluster goes away it's replaced, like a Pod whose node is gone. Editing the `nodeSelector` rolls replicas whose pinned pool no longer matches.

EKS is bumped to 1.36, where DRA is GA.

The `cel`, `quantity`, and `semver` modules reimplement the DRA device-selector CEL surface on the pure-Python celpy evaluator, since base CEL and celpy don't ship the Kubernetes extensions. They're matched against the upstream libraries (`k8s.io/apiserver/pkg/cel/library`, `blang/semver`, `resource.Quantity`), with test tables mirroring the upstream suites. The divergences celpy can't reach are documented at the top of `cel.py`.

Validated end to end on a provisioned EKS 1.36 cluster: the DRA driver published an L4 as a ResourceSlice, a `ModelDeployment` with a GPU `nodeSelector` scheduled a replica, DRA allocated the L4 to the serving pod, and the pod served Qwen2.5-0.5B through the gateway.

I have:


functions/compose-inference-cluster/function/fn.py
```diff
@@ -207,7 +207,7 @@ def compose_eks(self, eks):
         backend_secrets = self.resolve_eks_backend_secrets(eks_ready, backend_exists)
         if backend_secrets or backend_exists:
             if backend_secrets:
-                self.compose_kserve_backend(backend_secrets)
+                self.compose_serving_stack(backend_secrets)
```
negz, line 219: Not strictly related to this change, but it was broken in main. I'd like to get type checking working as part of `nix flake check` - I think it'd have caught this since `self.compose_kserve_backend` was no longer defined.

functions/compose-model-deployment/function/scheduling.py
```diff
@@ -81,153 +145,495 @@ def _cluster_ready(cluster: icv1alpha1.InferenceCluster) -> bool:
...
+    fails any request or matches only synthetic devices.
+
+    A request matches a pool device when the device has enough UNCONSUMED count
+    to cover the request and every selector evaluates true against that device.
+    Each resolved DRA request becomes a distinct DeviceRequest in one
+    ResourceClaim, and DRA allocates distinct devices per request, so a device's
+    count is consumed as requests claim it: two requests cannot both be satisfied
+    by the same single-count device, and N requests against one device must fit
+    within that device's count. Without this accounting the scheduler would place
+    a replica onto a node DRA can't actually satisfy.
```
negz: Our scheduler can race if there's two MDs being scheduled at the same time. The failure mode is that it could overcommit an IC. We build the capacity ledger from observed state, which could be stale if we're racing with the same function serving another MD. That'll only be a problem if the IC is close to full.

I think we should fix this post-v0.1, so I'll raise a tracking issue. I'm thinking we could have the retain phase  notice oversubscribed clusters and drop non-running replicas so they get rescheduled.

functions/compose-model-deployment/function/quantity.py
```diff
@@ -0,0 +1,194 @@
...
+        raise ValueError(f"invalid quantity: {original!r}")
+    # A trailing dot ("5.") is valid upstream but not for Decimal; drop it.
+    if num.endswith("."):
+        num = num[:-1]
+    return decimal.Decimal(num)
+
+
+def _round_nano(value: decimal.Decimal) -> decimal.Decimal:
+    """Round up to nano scale, matching resource.ParseQuantity (inf.RoundUp)."""
+    return value.quantize(_NANO, rounding=decimal.ROUND_UP)
```
dennis-upbound: `_round_nano` quantizes to `1e-9` under python's default 28-digit decimal context, so any quantity that needs >28 digits at nano scale throws `decimal.InvalidOperation` instead of parsing. threshold lands right above int64 — `1Ei`/`8Ei` parse, `10Ei`/`100Ei`/`256E` blow up. all three are valid `resource.Quantity` values, so this diverges from the upstream parity the module is built around, and it surfaces as `InvalidOperation` rather than the `ValueError` the docstring promises. inside a selector the broad `except` in `matches()` swallows it as a silent non-match, and `isQuantity("256E")` returns false where upstream returns true.

GPU memory is ~1e10 so the demo path never touches this, but it's a real gap in a module whose whole job is matching k8s. fix is a local context bump:

```python
def _round_nano(value):
    with decimal.localcontext() as ctx:
        ctx.prec = 60
        return value.quantize(_NANO, rounding=decimal.ROUND_UP)
```

`test_quantity.py` mirrors `quantity_test.go` but has no ≥Exa row, which is why it slipped — worth adding `quantity("256E")` / `10Ei` to lock it.

functions/compose-model-deployment/function/scheduling.py
```diff
@@ -81,153 +145,495 @@ def _cluster_ready(cluster: icv1alpha1.InferenceCluster) -> bool:
...
+def _device_satisfies(device, programs: list[cel.Program]) -> bool:
+    """Whether a pool device satisfies every selector (all ANDed)."""
+    # by_alias keeps the DRA wire names (bool/int, not the generated bool_/int_
+    # Python attribute names) so the CEL activation sees device.attributes the
+    # way DRA selectors expect.
+    raw = device.model_dump(by_alias=True, exclude_none=True)
+    return all(p.matches(raw) for p in programs)
+
+
+def _match_pool(pool, requests: list[_CompiledRequest]) -> list[DeviceRequest] | None:
```
dennis-upbound, line 193: `_match_pool` assigns requests to devices greedily in nodeSelector order — first device that matches and has the count gets consumed, then we move on. the documented multi-device cases are safe: a `gpu` request (`gpu.nvidia.com`) and a `nic` request (`nic.nvidia.com`) read different attribute domains, so their match-sets are disjoint and order can't starve either. the false-negative needs two requests whose match-sets *overlap* — e.g. a pool exposing two GPU device-kinds where a broad request (`A||B`) and a narrow one (`B`) both match one device. then listing the broad one first eats the device the narrow one needs and we reject the pool, control-plane-side, even though a valid assignment exists. repro: `[broad, narrow]` → no match, `[narrow, broad]` → match.

niche today. but since the device-list shape is explicitly sold (Alternatives considered) as the way to express multi-device requirements, the docstring shouldn't present the count accounting as sound — can we add a line that assignment is greedy in request order and overlapping requests can falsely reject a satisfiable pool? exact fix is a small bipartite match whenever overlapping selectors become a real case.

functions/compose-model-deployment/function/fn.py
```diff
@@ -236,7 +283,7 @@ def write_status(self, matched):
         replicas_ready = sum(
```
dennis-upbound, line 283: `write_status` and `derive_replicas_ready` both walk `matched` and re-run the same `get_condition(... "Ready")` lookup per replica to count ready ones — two passes computing the identical number. minor, but they can drift if one changes and the other doesn't. compute `replicas_ready` once and hand it to both?

functions/compose-model-replica/function/backends/native.py
```diff
@@ -12,7 +12,7 @@
 from function.backends import base
 
 # Namespace for serving workloads on remote clusters.
-_REMOTE_NAMESPACE = "default"
+_REMOTE_NAMESPACE = base.REMOTE_NAMESPACE
 
 # Port the engine serves the OpenAI-compatible API on.
 _ENGINE_PORT = 8000
```
dennis-upbound: `_ENGINE_PORT = 8000` and `_LABEL_SERVING = "modelplane.ai/serving"` are defined identically here and in `llmd.py`, and the Service/HTTPRoute shape around them is basically copy-pasted too. the port especially is a contract — the engine serves 8000, ModelEndpoint URLs assume it — so it shouldn't live in two places that can silently disagree. lift both into `base.py` next to `REMOTE_NAMESPACE`?

functions/compose-model-replica/function/backends/base.py
```diff
@@ -81,6 +81,64 @@ def apply_cache_args(args: list[str], replica: v1alpha1.ModelReplica, engine) ->
...
+# reference individual requests within the claim.
+_POD_CLAIM_NAME = "devices"
+
+# CEL readiness query matching workloads whose all-replicas-available signal is
+# an Available=True condition. Both a Deployment and a LeaderWorkerSet publish
+# this condition when their desired replicas are up; neither publishes a Ready
+# condition, so provider-kubernetes' DeriveFromObject policy (which only checks
+# a Ready condition) can never mark them ready. The has() guard keeps the query
+# false (not erroring) before the workload first writes status.conditions.
+AVAILABLE_CEL = (
```
dennis-upbound, line 114: the comment says both a Deployment and a LeaderWorkerSet publish `Available=True`, and we lean on that for the only readiness signal (DeriveFromCelQuery). Deployment definitely does. did the LWS path actually get observed going Ready end-to-end, or is the LWS `Available` condition inferred? if LWS doesn't publish it the multi-node replica never marks ready and the deployment hangs in `ModelStarting` forever — just want to make sure that's verified and not assumed, since the e2e in the description reads like the single-replica native path.

design/design.md
```diff
@@ -650,24 +687,25 @@ The fleet scheduler picks `(InferenceCluster, pool)` for each ModelReplica:
...
-3. **Derive physical shape** from `workers.topology`: nodes per worker, GPUs per
-   node.
-4. **Check capacity.** Does the pool have enough available nodes? Available =
+2. **Filter pools** by evaluating each `nodeSelector.devices` request against the
+   pool's InferenceClass devices. A pool matches a request when it has a device
+   whose `count` covers the request and whose `driver`, `attributes`, and
+   `capacity` satisfy every selector. A pool matches the deployment when it
+   satisfies every request.
+3. **Check capacity.** Does the pool have enough available nodes for one replica
+   (`pipeline * data / dataLocal` from `workers.topology`)? Available =
```
dennis-upbound: this says the gate needs enough nodes "for one replica (`pipeline * data / dataLocal`)", but that formula is nodes-per-*worker* — the table at line 441 labels it exactly that. `topology_shape` (scheduling.py:124) computes nodes_per_replica = nodes_per_worker * `workers.count`, i.e. `pipeline * count` in v0.1 (data/dataLocal aren't read yet). for the disagg decode example above (`count: 3`, `pipeline: 2`) the code gates on 6 nodes/replica; this formula gives 2. can we multiply by `workers.count` here (and at line 448) so the per-replica figure matches what the scheduler actually consumes?

dennis-upbound (commented):
Went deep on the DRA matching path — ran the compose-model-deployment tests and reproduced the two correctness items below against celpy. Solid overall; notes inline. The quantity and match-pool ones are the substantive ones, the rest are nits / a doc-vs-code mismatch.

dennis-upbound (approved):
approved with the comments

functions/compose-model-deployment/function/quantity.py
negz: Good spotting. I added a little test helper to validate against real upstream behavior, which found a few other places we could be more faithful.

functions/compose-model-deployment/function/scheduling.py
negz: I went down a whole path of exploring backtracking / bipartite but in the end I feel like this is exceedingly unlikely to ever happen so went with the docstring update.

functions/compose-model-deployment/function/fn.py
negz: I played with refactoring this but I prefer it as is. It's cheap and I'm not worried about drift.

functions/compose-model-replica/function/backends/native.py
negz: I think you introduced these, but agree. 😉 Fixed.

dennis-upbound: heh thanks!

functions/compose-model-deployment/function/quantity.py
dennis-upbound: nice. is there some sort of DRA conformance tests we can run?

functions/compose-model-replica/function/backends/base.py
negz: I didn't test the LWS scenario E2E but I did check the code and it does appear to use Available to indicate the set is ready.

Issue #28, opened 2026-04-09
Add CEL validation rules for discriminated unions


negz:
### What problem are you facing?

Modelplane's XRDs use discriminated unions throughout: an enum field determines which sibling config block is required (e.g., `backend: KServe` requires a `kserve` block). There's no validation enforcing this -- a user can set `backend: KServe` without providing a `kserve` block and the API server accepts it. The error surfaces later when the composition function can't find the expected config.

### How could Modelplane help solve your problem?

Add `x-kubernetes-validations` CEL rules to each discriminated union. The pattern is:

```yaml
x-kubernetes-validations:
- rule: "self.backend != 'KServe' || has(self.kserve)"
  message: "kserve is required when backend is KServe"
- rule: "self.backend != 'Dynamo' || has(self.dynamo)"
  message: "dynamo is required when backend is Dynamo"
```

negz:
Let's also check for anything else in the APIs that'd benefit from more CEL validation - e.g. https://github.com/modelplaneai/modelplane/pull/27#discussion_r3080690966

PR #112, opened 2026-06-08
Pre-stage model weights with a ModelCache primitive

### LINKED ISSUE #66: ModelCache v0.1 — PVC backend, multi-node
**Design lives in [`design/modelcache/`](https://github.com/modelplaneai/modelplane/blob/dennis/modelcache-design/design/modelcache/design.md) (branch `dennis/modelcache-design`).** This issue tracks v0.1 implementation.

## v0.1 scope

PVC backend, multi-node ready, no dedup. From the [design doc](https://github.com/modelplaneai/modelplane/blob/dennis/modelcache-design/design/modelcache/design.md):

- `ModelCache` CRD with artifact discriminator (`Weights`, `Tokenizer`, `Bytes`)
- Sources: `huggingFace`, `s3`, `http`, `inline`, `configMap`
- `PVC` (RWX) storage backend with one-shot prefetch Job (absorbs [#61](https://github.com/modelplaneai/modelplane/issues/61))
- `replication: AllMatchingClusters` — one RWX PVC per matching cluster, shared across all LWS gang pods
- `clusterSelector.matchLabels` for cluster filtering
- Mount path intrinsic to the cache; deployments reference by name via `caches: [{ name }]`
- Scheduling gated on per-cluster cache `Ready` condition
- Fail-fast when a target cluster has no RWX storage class on `InferenceCluster.spec.storage.storageClassName`
- Pluggable storage backend pattern shared with [#72 KVOffloadTier](https://github.com/modelplaneai/modelplane/issues/72)

## Out of scope (tracked separately)

- `LoraAdapter` / `Engine` artifact kinds → v0.2
- `ContentAddressed` backend (Modal-style tiered cache + lazy loading) → v0.2
- Cross-deployment / cross-tenant dedup → v0.2
- `gcs` / `azure` / `oci` / `pvc-clone` sources → v0.2
- `AllMatchingNodes` replication mode → v0.2
- Substrate unification [#72](https://github.com/modelplaneai/modelplane/issues/72) → v0.3

Roadmap detail in the design doc [§ v0.2](https://github.com/modelplaneai/modelplane/blob/dennis/modelcache-design/design/modelcache/design.md#v02--content-addressed-backend-lazy-loading-full-artifact-taxonomy) and [§ v0.3](https://github.com/modelplaneai/modelplane/blob/dennis/modelcache-design/design/modelcache/design.md#v03--substrate-unification-architectural-option).

## Examples

Nine (ModelCache + ModelDeployment) examples in [`design/modelcache/examples/`](https://github.com/modelplaneai/modelplane/tree/dennis/modelcache-design/design/modelcache/examples): single-cluster basic, multi-node TensorPipeline gang, multi-cluster replication, separate tokenizer, private S3, opaque `Bytes` kind, plus three v0.2 previews.

## References

- [Design doc](https://github.com/modelplaneai/modelplane/blob/dennis/modelcache-design/design/modelcache/design.md)
- [Examples](https://github.com/modelplaneai/modelplane/tree/dennis/modelcache-design/design/modelcache/examples)
- [#61](https://github.com/modelplaneai/modelplane/issues/61) (closed) — RWX PVC mechanism
- [#72](https://github.com/modelplaneai/modelplane/issues/72)
- [PR #75](https://github.com/modelplaneai/modelplane/pull/75) — `engine.env` + `imagePullSecrets`; ModelCache rides on those for credential-bearing sources

dennis-upbound:
Fixes #66.

Serving pods fetched model weights from HuggingFace at every boot. For frontier models that's tens to hundreds of GB pulled on every pod start, and in a multi-node LeaderWorkerSet gang each pod pulls independently — so an 8-pod gang downloads the same 600GB eight times before it can serve. There was no way to stage weights once and share them.

This adds `ModelCache`, a v0.1 primitive that pre-stages a HuggingFace model onto a per-cluster ReadWriteMany PVC via a one-shot hydration Job, so serving pods mount the weights instead of downloading them. A deployment opts in with `spec.modelCacheRef`; the cache PVC mounts at `/mnt/models` on every serving pod, shared across the whole gang, and the weights are downloaded once per cluster and read N times.

The `compose-model-cache` function fans a `ModelCache` out to every matched `InferenceCluster`: a RWX PVC sized from the source, plus a hydration Job that runs `hf download` into it. Hydration is re-run safe via a completion marker, so an interrupted download resumes rather than serving truncated weights. Per-cluster phase (Pending/Hydrating/Ready/Failed) and an `x/y` ready summary are reported on the cache's status.

`modelCacheRef` now threads from `ModelDeployment` through to each `ModelReplica`, and the native and llm-d backends mount the cache. The mount is engine-agnostic — it lands on the native Deployment pod and on both the leader and worker templates of an llm-d gang. `--model=/mnt/models` is injected only for the turnkey vLLM path; a bring-your-own engine like SGLang sets its own `--model-path`, so injection is skipped when the engine supplies its own command.

On GKE the `modelplane-rwx` storage class self-provisions: `compose-gke-cluster` enables the Filestore API and `compose-inference-cluster` composes a VPC-pinned Filestore StorageClass. On EKS the cache works against an admin-provided EFS StorageClass (`modelplane-rwx-efs`); auto-provisioning EFS is a separate follow-up.

Before, a deployment had to bake the model into the engine args and re-pull every boot:

```yaml
spec:
  workers:
    template:
      spec:
        containers:
        - name: engine
          args: ["--model=Qwen/Qwen3-0.6B"]   # fetched from HF on every pod start
```

After, the cache is staged once and mounted:

```yaml
spec:
  modelCacheRef:
    name: qwen
  workers:
    template:
      spec:
        containers:
        - name: engine
          args: ["--model=/mnt/models"]       # read from the mounted PVC
```

Scope is locked to the HuggingFace source plus a Modelplane-managed PVC, matching the merged XRD. EKS works against an admin-provided EFS StorageClass; auto-provisioning EFS is a separate follow-up.

How it was validated so far: unit tests for every changed function (the cache function's PVC/Job/status/conditions, the deployment→replica propagation, and the backend mounts including the vLLM-inject vs SGLang-skip split). Live cluster validation is pending.

I have: 

- [x] Read and followed Modelplane's contribution process.
- [x] ~Run `nix flake check` and made sure it passes.~ Per-function unit tests pass and ruff lint/format is clean locally; the full sandboxed `nix flake check` runs in CI on this PR.
- [x] Added or updated tests covering the composition function changes.


docs/concepts.md
```diff
@@ -155,6 +162,23 @@ args should reference this path (e.g. `--model=/mnt/models` for vLLM).
...
+  StorageClass and enables `file.googleapis.com`. Nothing for the admin to do.
+- **EKS:** bring-your-own for v0.1. On the workload cluster the admin must
+  install the `aws-efs-csi-driver` EKS add-on (with an IRSA role bound to
+  `AmazonEFSCSIDriverPolicy`); create an EFS file system with a mount target in
+  each node subnet and a security group allowing inbound NFS (2049) from the
+  node security group; and create a StorageClass named `modelplane-rwx-efs`
+  with `provisioner: efs.csi.aws.com`, `provisioningMode: efs-ap`, and
+  `fileSystemId: <fs-id>`. Set `eks.cache.storageClassName` if the admin's
+  class has a different name. EFS is elastic, so the cache's `sizeGiB` is
+  informational on EKS — the PVC API still requires a size, but EFS ignores it.
```
negz, line 181: Is there a path to do this automatically (post-v0.1)? If so please raise a tracking issue.

examples/cache/qwen-cached.yaml
```diff
@@ -0,0 +1,42 @@
...
+#
+# The ModelCache stages the weights on workload-cluster storage once; the
+# ModelDeployment references it and mounts it at /mnt/models on its pod.
+apiVersion: modelplane.ai/v1alpha1
+kind: ModelCache
+metadata:
+  name: qwen
+  namespace: ml-team
+spec:
+  source:
```
negz: I missed in https://github.com/modelplaneai/modelplane/pull/84: I think we should add a required discriminator field  now, even though we only support HF. So something like:

```yaml
spec:
  source: HuggingFace
  huggingFace:
    ...
```

I prefer required discriminator fields because it forces the user to be explicit about what they want, and makes the code that handles the API simpler. We don't need to e.g. try to infer what the source is based on what fields (`spec.huggingFace` etc) are set. We can't add a required field later though - so I like to start with one just in case.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+        self.rsp = rsp
+        self.xr = v1alpha1.ModelCache(**resource.struct_to_dict(req.observed.composite.resource))
+        self.clusters: list[icv1alpha1.InferenceCluster] = []
+
+    def compose(self):
+        # The XRD can't yet enforce "exactly one source" (no CEL union rule —
+        # issue #28), so a ModelCache with an empty/unknown source reaches us.
+        # Fail fast with a clear condition rather than NPE in _hf_hydration.
+        # This is also the seam where future sources (s3/http/inline) plug in:
+        # extend _source_supported() and dispatch in _job_manifest().
```
negz: I think if we add a required discriminator field we _could_ add CEL to validate that the user set the relevant object right? In which case we wouldn't need this code-level defense.

Note https://github.com/modelplaneai/modelplane/issues/28 is a TODO to enforce via CEL - not tracking that we _can't_ do that today.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+    def resolve_inputs(self) -> bool:
+        """Require all InferenceClusters matching the (optional) selector.
+
+        Returns False when Crossplane hasn't resolved the requirement yet;
+        Crossplane re-calls the function once it's available. A resolved-but-
+        empty match flows through (match_clusters() -> NoClusters condition).
+        """
+        match_labels = {LABEL_KEY_CLUSTER: LABEL_VALUE_CLUSTER}
+        if self.xr.spec.clusterSelector and self.xr.spec.clusterSelector.matchLabels:
+            match_labels.update({k: str(v) for k, v in self.xr.spec.clusterSelector.matchLabels.items()})
```
negz, line 168: I don't think either the design or any of the examples shows how to use `spec.clusterSelector` for a ModelCache. It'd be good to show an example, and to document what happens when you omit it (which I presume is it matches all clusters).

functions/compose-inference-cluster/function/fn.py
```diff
@@ -166,6 +167,15 @@ def compose_gke(self, gke):
         if gke_ready and kubeconfig:
             self.compose_cluster_provider_config(kubeconfig.name, kubeconfig.key, sa_key)
 
+        if gke_ready and kubeconfig:
```
negz: This looks like the same as line 167 above right? Can we merge the two blocks?

functions/compose-inference-cluster/function/fn.py
```diff
@@ -166,6 +167,15 @@ def compose_gke(self, gke):
...
+            provider_config = resource.child_name(self.xr.metadata.name, "cluster-kubeconfig")
```
negz: I think this'll work (assuming this is how we name the kubeconfig when we create it) but it might be more future proof to read the observed providerconfig's name. That way we'll have the right name here if we change how we name it in future.

(Confusingly `kubeconfig.name` above seems to actually be the kubeconfig's secret's name?)

functions/compose-inference-cluster/function/fn.py
```diff
@@ -286,6 +296,57 @@ def compose_cluster_provider_config(self, kubeconfig_name, kubeconfig_key, sa_ke
...
+        VPC name carries a provider-generated suffix, so it CANNOT be derived
+        from the XR name — pinning to the bare name fails with "network does not
+        exist" (verified live on GKE). None until the GKECluster observes its
+        network; the StorageClass is gated on this being present.
+        """
+        observed = self.req.observed.resources.get("gke-cluster")
+        if not observed:
+            return None
+        manifest = resource.struct_to_dict(observed.resource)
+        return manifest.get("status", {}).get("network", {}).get("name") or None
```
negz: I'd prefer to use the Pedantic models when we can - i.e. deserialize this into a GKECluster rather than an opaque dict.

functions/compose-model-deployment/function/fn.py
```diff
@@ -169,6 +169,12 @@ def compose_replicas(self, matched):
         for cluster_info in matched:
             replica_key = f"replica-{cluster_info.name}"
 
+            spec_kwargs = {"clusterName": cluster_info.name, "workers": workers}
+            if self.xr.spec.modelCacheRef:
+                spec_kwargs["modelCacheRef"] = mrv1alpha1.ModelCacheRef(
+                    name=self.xr.spec.modelCacheRef.name,
+                )
```
negz: Does using kwargs here weaken the type checking we'd get from using `ModelReplica`? 🤔 Could we instead form the ModelReplica model outside of `resource.update` and set `replica.spec.modelCacheRef = ModelCacheRef(name=...)` only if the MD XR has one?

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
+"""Compose a ModelCache.
+
+Stages a HuggingFace model onto a ReadWriteMany PVC on every matched
+InferenceCluster via a one-shot hydration Job. Pods that reference the
+cache (ModelDeployment.spec.modelCacheRef -> ModelReplica) mount the PVC
+at /mnt/models, so weights are downloaded once per cluster and read N
+times by every pod in an LWS gang.
+
+v0.1 surface (locked to the merged XRD): source `huggingFace` only,
+Modelplane-managed RWX PVC, replication to all matching clusters.
```
negz: Might be worth dropping this - I could see us forgetting to clean it up later.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+# compose-model-deployment's cluster matching exactly. negz's PR #51 removes
+# this workaround (bare ResourceSelector once `up` ships Crossplane >=2.2.1 and
+# function-sdk-python grows a require_all helper) — when it lands, migrate this
+# matcher alongside compose-model-deployment's, not separately.
+LABEL_KEY_CLUSTER = "modelplane.ai/cluster"
+LABEL_VALUE_CLUSTER = "true"
+
+# Hydration container. python:3.11-slim has pip; we install huggingface_hub
+# at runtime. A Modelplane-owned image with the tool preinstalled is a
+# follow-up.
```
negz: Please raise a tracking issue for this too.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+# emptiness check, which had the partial-download false-skip bug.
+_HYDRATED_MARKER = f"{HYDRATION_MOUNT}/.modelplane-hydrated"
+_SKIP_IF_HYDRATED = f"if [ -f {_HYDRATED_MARKER} ]; then echo 'already hydrated, skipping'; exit 0; fi; "
+
+
+def _hf_hydration(hf) -> tuple[list[dict], str]:
+    """Return (env, shell command) for a HuggingFace source.
+
+    huggingface-hub 1.x deprecated `huggingface-cli` (and dropped the `[cli]`
+    extra), so the Job previously died at install. Use `hf download` directly.
```
negz: I think we can drop this context - it seems like the agent narrating a change from a previous approach it tried that never actually saw the light of day.

Interestingly my agents want to do this all the time too. I thought I added something to try stop it in CONTRIBUTING.md but I now can't find it.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+                forProvider=k8sobjv1alpha1.ForProvider(manifest=manifest),
+            ),
+        )
+
+    # --- naming (must stay in sync with backends/base.cache_pvc_name) ---
+    # Namespace-qualified so caches of the same name from different Modelplane
+    # namespaces don't collide in the workload cluster's `default` namespace
+    # (Nic's recurring #99 collision concern).
+    def _pvc_name(self) -> str:
+        return f"modelcache-{self.xr.metadata.namespace}-{self.xr.metadata.name}"[:63]
```
negz: Could we use `child_name` here (and for `_job_name`)? iirc it handles the truncation intelligently.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+            return PHASE_HYDRATING
+        return PHASE_PENDING
+
+    def _observed_status(self, key: str) -> dict:
+        """Remote resource status echoed back under Object.status.atProvider.manifest.status."""
+        observed = self.req.observed.resources.get(key)
+        if not observed:
+            return {}
+        d = resource.struct_to_dict(observed.resource)
+        return d.get("status", {}).get("atProvider", {}).get("manifest", {}).get("status", {}) or {}
```
negz: Would be good to use the Pedantic model here too.

functions/compose-model-cache/function/fn.py
```diff
@@ -0,0 +1,413 @@
...
+                                "name": "artifact",
+                                "persistentVolumeClaim": {"claimName": self._pvc_name()},
+                            }
+                        ],
+                    },
+                },
+            },
+        }
+
+    def derive_cluster_phase(self, cluster_name: str) -> str:
```
negz, line 285: FWIW the `Object` MR has a few options for deriving its Ready condition from the resource it wraps - e.g. all conditions true, CEL conditional, etc. That could simplify the work being done here to just checking if the relevant Objects are ready.

functions/compose-model-cache/tests/test_fn.py
```diff
@@ -0,0 +1,204 @@
+"""Tests for the compose-model-cache function."""
```
negz, line 1: I'd like to follow the same pattern as the rest of the function tests - table of Cases testing at the request/response level.

negz (commented):
Thanks @dennis-upbound!

 really like the PR description. I found it way easier to read. Did you get an agent to generate it using the new guidance in CONTRIBUTING.md?

docs/concepts.md
dennis-upbound: Filed #114 to track auto-provisioning EFS on EKS.

examples/cache/qwen-cached.yaml
dennis-upbound: Agreed — and it lines up with the `InferenceCluster.spec.cluster.source` pattern we already use. `source` is now a required enum with a sibling `huggingFace` object, plus a CEL rule requiring the matching object (which let me drop the runtime guard below).

functions/compose-model-cache/function/fn.py
dennis-upbound: Done — added the CEL `has(self.huggingFace)` rule alongside the discriminator, so the code-level defense is gone. Thanks for the #28 clarification; I'd read it as "can't yet".

functions/compose-model-cache/function/fn.py
dennis-upbound: Added a commented `clusterSelector` example and documented that omitting it stages the cache on every matched cluster.

functions/compose-inference-cluster/function/fn.py
dennis-upbound: Merged.

functions/compose-inference-cluster/function/fn.py
dennis-upbound: Switched to reading the observed ProviderConfig's name so it survives a naming change, and renamed the local — you're right that `kubeconfig.name` was actually the secret's name.

functions/compose-inference-cluster/function/fn.py
dennis-upbound: Done — deserializing into `GKECluster` and reading `status.network.name` typed.

functions/compose-model-deployment/function/fn.py
dennis-upbound: Reworked to build the typed `ModelReplica` and set `spec.modelCacheRef` only when the deployment has one.

functions/compose-model-cache/function/fn.py
dennis-upbound: Dropped — agreed, a point-in-time note that'd rot.

functions/compose-model-cache/function/fn.py
dennis-upbound: Filed #115.

functions/compose-model-cache/function/fn.py
dennis-upbound: Switched both to `child_name`. It's a contract with `backends/base.cache_pvc_name` (the serving side mounts the PVC by this exact name), so I changed both together.

functions/compose-model-cache/function/fn.py
dennis-upbound: Done — deserialize the `Object` envelope via the model; the inner `manifest.status` stays a dict since it's the wrapped resource's freeform status.

functions/compose-model-cache/function/fn.py
dennis-upbound: Nice — set `DeriveFromCelQuery` on the PVC (`status.phase == "Bound"`) and Job (`Complete` condition) Objects so each derives its own Ready. Kept a thin status read for the per-cluster `Hydrating`/`Failed` phase we surface in `status.clusters[]`.

functions/compose-model-cache/tests/test_fn.py
dennis-upbound: Converted to the table-of-Cases request/response golden pattern.

dennis-upbound:
> Thanks @dennis-upbound!
> 
> really like the PR description. I found it way easier to read. Did you get an agent to generate it using the new guidance in CONTRIBUTING.md?

yep! thanks for setting up the guidance. I had to still clean a little llm fluff but overall not bad

functions/compose-model-cache/function/fn.py
dennis-upbound: dropped - added to memories so it doesn't narrate intra-MR stuff like this

negz (approved):

Issue #34, opened 2026-04-14
Support prefill/decode disaggregation


negz:
### What problem are you facing?

Prefill is compute-bound, decode is memory-bandwidth-bound. Running them on the same GPU pool means optimizing for neither. Modelplane has no way to express disaggregated serving today.

### How could Modelplane help solve your problem?

LLM inference has two phases. Prefill processes all the input tokens in parallel. It's compute-bound and determines time-to-first-token. Decode generates output tokens one at a time. It's memory-bandwidth-bound and determines inter-token latency. When both phases share a GPU pool, a prefill burst interferes with in-flight decodes, causing unpredictable tail latency. You can't independently tune parallelism for each phase.

Disaggregation runs them on separate pools. A prefill instance processes the prompt, then transfers its KV cache (the intermediate state it computed for every input token) to a decode instance over RDMA. This lets you use different parallelism strategies per phase. For example TP=1 prefill replicas for throughput, TP=4 decode replicas for latency. llm-d's benchmarks show 40-50% lower end-to-end latency for workloads with long input sequences (high input:output token ratio). It's not always the right choice. Short prompts and small models don't benefit enough to justify the KV cache transfer overhead. But for large models with long contexts it's a significant win.

KServe (which Modelplane uses as its inference stack) already supports this. An optional `prefill` section on `LLMInferenceService` tells the controller to create separate Deployments for prefill and decode. KServe v0.18 ships decode templates that include the llm-d routing sidecar (which coordinates KV cache transfer via NIXL), and supports per-phase autoscaling via WVA.

I think the natural place to express disaggregation in Modelplane is split across two resources: engine configuration on ClusterModel (or Model), and the topology choice plus replica counts on ModelDeployment. The platform team declares that disagg is *available* for a model on certain hardware. The ML team *activates* it for their deployment.

The serving profile gains a `topology` discriminator, a `prefill` block with its own engine spec and per-pod resources, and optional per-profile `resources` for the decode side. Each phase's engine spec is complete and independent, with no merging and no shared base args. This follows the existing serving profile philosophy where each profile is a complete, tested configuration.

```yaml
apiVersion: modelplane.ai/v1alpha1
kind: ClusterModel
metadata:
  name: llama-405b
spec:
  model:
    name: meta-llama/Llama-3.1-405B-Instruct
  source: HuggingFace
  huggingFace:
    repo: meta-llama/Llama-3.1-405B-Instruct
  resources:
    vram: "810Gi"

  serving:
[...]

PR #116, opened 2026-06-09
Prefill/decode disaggregation design

dennis-upbound:
Towards #34.

Issue #34 asks for prefill/decode disaggregation, but its original sketch predates the KServe drop and describes it through KServe's `LLMInferenceService.prefill` section. There's no current design or user-facing documentation for how disaggregation fits the backend-dispatcher architecture.

This adds two docs, no code:

- `design/disaggregation.md` specifies disaggregation on the current architecture: a self-contained `prefill` block on the deployment, KV transfer via the engine's `--kv-transfer-config` and NIXL, routing through the same swappable EPP on a GAIE `InferencePool` that unified serving uses, the correctness constraints to enforce as matching matures, and when disaggregation is worth it. It parallels `design/modelcache.md`, with an Alternatives considered section covering the KServe sketch, a bespoke proxy, a routing discriminator, Modelplane choosing per-role hardware, and an implicit prefill:decode ratio.
- `docs/concepts.md` gains a Disaggregated Serving section covering the prefill block, the compute-bound vs bandwidth-bound split, KV transfer, and the ModelCache and co-location requirements. It also updates the ModelCache source bullet to the shipped `source` enum shape.

It also carries one unrelated CI fix (`Skip the checklist requirement on draft PRs`), bundled here because it's what surfaced the failure on this draft. The `checklist-completed` job ran `require-checklist-action` on every `pull_request` event, so a WIP draft that hadn't filled out the template checklist failed the check. The commit gates the job on the PR not being a draft and adds the `ready_for_review` trigger, so it's skipped while a PR is a draft and runs once it's marked ready. It's a self-contained commit and can be split into its own PR if you'd rather not bundle it.

I have: 

- [x] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~ Docs and CI-config change; the sandboxed `check` job passes in CI on this PR.
- [x] ~Added or updated tests covering any composition function changes.~ No composition function changes.

### DIFF
--- a/design/disaggregation.md
+++ b/design/disaggregation.md
@@ -167,13 +167,15 @@ at `pipeline: 1`, because disaggregation needs cross-pod coordination regardless
 of the per-role topology.
 
 **Gateway.** `InferencePool` as an `HTTPRoute` backend needs a GAIE-conformant
-gateway; core Envoy Gateway — which ServingStack installs today for plain
-`HTTPRoute → Service` routing — does not serve it. Of the conformant options
-(Envoy AI Gateway, Istio, kgateway), **Modelplane chooses Envoy AI Gateway**: it
-layers on the Envoy Gateway data plane ServingStack already runs (an additive
-controller + CRDs, not a gateway swap), so it's the lowest-friction change and
-leaves existing plain routes untouched — where Istio or kgateway would replace
-the gateway. Unified serving keeps its plain `Service` route for now; the
+gateway. ServingStack installs **Envoy AI Gateway** as the default — it's
+Envoy-based, purpose-built for LLM traffic, and serves `InferencePool` v1 — so
+disaggregation works out of the box on a Modelplane-provisioned cluster. But the
+gateway is a swappable seam, like the EPP: any GAIE-conformant gateway works
+(Istio, kgateway, GKE Gateway, …), so a customer can plug in their own — for a
+BYO cluster that already runs one, or simply a different preference. (Plain
+Envoy Gateway is not conformant on its own; `InferencePool` support comes from
+the AI Gateway layer, which is why we install that rather than bare Envoy
+Gateway.) Unified serving keeps its plain `Service` route for now; the
 `InferencePool`/EPP path is used by disaggregated serving (and can later carry
 KV-/load-aware routing for unified serving too).
 
--- a/design/disaggregation.md
+++ b/design/disaggregation.md
@@ -21,6 +21,14 @@ generates the output. Modelplane expresses this with a `prefill` block on the
 deployment: the top-level `workers` is the decode (or unified) role, and adding
 a `prefill` block makes the deployment disaggregated.
 
+It pays off for large models under load with strict TTFT/ITL targets, long
+context, and a fast interconnect, where the two phases' loads are large and
+skewed enough to tune separately; for small models, short context, or low
+traffic the KV-transfer overhead outweighs the benefit and aggregated serving
+(optionally with chunked prefill) is simpler. The choice is the operator's:
+Modelplane serves unified by default and disaggregates only when a `prefill`
+block is set.
+
 ```yaml
 apiVersion: modelplane.ai/v1alpha1
 kind: ModelDeployment
@@ -31,7 +39,7 @@ spec:
   replicas: 1
   modelCacheRef:
     name: llama-405b
-  # Top-level workers: the decode role.
+  # Top-level workers: the decode role (memory-bandwidth-bound).
   workers:
     count: 3
     topology:
@@ -44,7 +52,19 @@ spec:
           args:
           - "--model=/mnt/models"
           - '--kv-transfer-config={"kv_connector":"NixlConnector","kv_role":"kv_consumer"}'
-  # Prefill role. Self-contained.
+  # Decode's hardware. nodeSelector is required (a list of DRA device requests);
+  # here a high-VRAM GPU plus the InfiniBand fabric the KV transfer needs.
+  nodeSelector:
+    devices:
+    - name: gpu
+      count: 8
+      selectors:
+      - cel: device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("141Gi")) >= 0
+    - name: nic
+      count: 8
+      selectors:
+      - cel: device.attributes["nic.nvidia.com"].linkType == "infiniband"
+  # Prefill role. Self-contained, with its own (compute-bound, single-GPU) hardware.
   prefill:
     workers:
       count: 5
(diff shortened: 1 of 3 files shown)


design/disaggregation.md
```diff
@@ -0,0 +1,182 @@
...
+inter-token latency. Run on the same pods, a prefill burst stalls in-flight
+decodes and neither phase can be tuned independently.
+
+Disaggregation runs the two phases as separate pod sets. A prefill instance
+processes the prompt and transfers its KV cache to a decode instance, which
+generates the output. Modelplane expresses this with a `prefill` block on the
+deployment: the top-level `workers` is the decode (or unified) role, and adding
+a `prefill` block makes the deployment disaggregated.
+
+```yaml
```
negz, line 32: Nit: This is invalid with https://github.com/modelplaneai/modelplane/pull/101 merged. `nodeSelector` is now required. (I need to update README too.)

design/disaggregation.md
```diff
@@ -0,0 +1,182 @@
...
+These are documented now and enforced as the matching and validation surfaces
+mature.
+
+- **Co-location.** A replica's prefill and decode must be schedulable on one
+  InferenceCluster. The fleet scheduler rejects the deployment if no matched
+  cluster can host both roles.
+- **Interconnect.** KV transfer needs NVLink within a node or RDMA/InfiniBand
+  across nodes; over PCIe or ethernet it bottlenecks. It is required as a
+  cluster or pool capability (e.g. `networkInterNode`) and matched the same way
+  as other hardware requirements.
```
negz: This'd be a `Synthetic` device with #101 merged.

design/disaggregation.md
```diff
@@ -0,0 +1,182 @@
...
+
+## When to use
+
+Disaggregation pays off for large models under load with strict TTFT and ITL
+targets, long context, and a fast interconnect, where prefill and decode load
+are large enough and skewed enough to tune separately. For small models, short
+context, or low traffic, the KV-transfer overhead outweighs the benefit;
+aggregated serving, optionally with chunked prefill, is simpler and usually
+faster. The decision is the operator's. Modelplane serves unified by default and
+disaggregates only when a `prefill` block is set.
```
negz: Is this useful for the design doc? If so, should we fold it into background?

design/disaggregation.md
```diff
@@ -0,0 +1,182 @@
...
+serving. Reusing it means one routing component for both unified and
+disaggregated serving rather than a disaggregation-only proxy.
+
+### A routing discriminator instead of a template
+
+The EPP could be selected by a `picker` enum. Instead it is a curated PodSpec
+subset (`routing.template`), the same shape and owner as the engine, defaulting
+to the llm-d EPP and overridable by image and args. This avoids a discriminator
+for a component that is really just a container, matching the engine convention
+and design.md's preference against gratuitous discriminators.
```
negz: I don't follow this alternative. Maybe an API sketch would help?

design/disaggregation.md
negz: Also showing that both blocks have their own selectors would probably be illustrative.

design/disaggregation.md
```diff
@@ -0,0 +1,182 @@
...
+
+Disaggregation pays off for large models under load with strict TTFT and ITL
+targets, long context, and a fast interconnect, where prefill and decode load
+are large enough and skewed enough to tune separately. For small models, short
+context, or low traffic, the KV-transfer overhead outweighs the benefit;
+aggregated serving, optionally with chunked prefill, is simpler and usually
+faster. The decision is the operator's. Modelplane serves unified by default and
+disaggregates only when a `prefill` block is set.
+
+## Alternatives considered
```
negz: Another alternative that comes to mind: could you just write two MDs? One for prefill and the other for decode?

It feels really close to what's proposed, and I like that it just reuses existing primitives. However I think what's proposed is better:

* With one MD the scheduler has everything it needs to make sure prefill and decode workers are collocated sensibly. With two MDs either the author has to get _real_ crafty with their nodeSelectors, or we add co-scheduling hints to MD (and the scheduler probably requires all MDs).
* With one MD a disaggregated MD is still conceptually "a model deployment". An MD that's doing prefill only doesn't feel like it is.

negz (approved):

design/disaggregation.md
dennis-upbound: Fixed — the example now shows the required `nodeSelector` on both the decode and prefill roles.

dennis-upbound: Done — both roles now carry their own device selectors (decode: high-VRAM GPU + IB; prefill: single GPU + IB).

design/disaggregation.md
dennis-upbound: Done — the InfiniBand NIC is now a `claim: Synthetic` device, in both the example and the Interconnect constraint.

design/disaggregation.md
dennis-upbound: Folded the "when to use" guidance into the Summary as background and dropped the standalone section.

design/disaggregation.md
dennis-upbound: Added an API sketch contrasting the `picker` enum with `routing.template`, so the discriminator-vs-template tradeoff is concrete.

design/disaggregation.md
dennis-upbound: Added a "Two ModelDeployments" alternative capturing your reasoning — agreed the single MD is better: the scheduler can co-locate the roles, and a prefill-only MD is not conceptually a model deployment.

design/disaggregation.md
```diff
@@ -0,0 +1,281 @@
...
+
+```yaml
+spec:
+  routing:
+    template:
+      spec:
+        containers:
+        - name: epp
+          image: ghcr.io/llm-d/llm-d-inference-scheduler:v0.8.0   # default
+          args: ["--config-file=/config/epp.yaml"]                # override to tune scorers
```
negz: So for prefill decode we have the usual top level stuff, and `prefill`, and `routing`?

I think this sketch belongs in the main body of the design, not in an alternative, and could use a little more detail. It'd be useful to show it in context of a full deployment (that'd answer my question above).

Some other thoughts:

* I'm a little wary of defaulting it if omitted. It's convenient but asymmetrical with the other blocks, where we don't e.g. guess an engine config if you omit it.
* Where does the EPP pod run? (Does it need its own node selector etc etc?)

dennis-upbound: Good points — all addressed:

- **Shape:** yes — top-level `workers` (decode), `prefill`, and `routing`. Moved the sketch into the main body: the example up top is now a full deployment showing all three, and the Routing section describes `routing.template` directly. Trimmed the alternative to just the discriminator-vs-template rationale.
- **Defaulting:** agreed, dropped it. `routing` is now required when `prefill` is set (no guessed EPP), and both `workers.count` and `prefill.workers.count` are explicit (no default ratio) — symmetrical with how the engine is specified, per your point.
- **Where the EPP runs:** added — it’s a lightweight CPU Deployment (watches the pods, scores requests; no model, no GPU), on ordinary nodes in the serving namespace, no special `nodeSelector`, one per `InferencePool`. Its pod shape comes from `routing.template`.

Issue #117, opened 2026-06-09
docs: release readiness


tr0njavolta:
- [x] Capture reference material
- [x] Create vercel/netlify deployment
- [ ] Create Algolia application
- [ ] Introduction/overview rewrite (currently the README for the project)
- [ ] Create docs README
- [ ] Review assets for brand conformity
- [ ] Attribution and references to Crossplane/Upbound/CNCF/Linux Foundation
- [ ] Create versioning scheme and rules
- [ ] Add manifests/examples directory hosting
- [ ] Add AI-forward features (llms.txt, Copy MD for AI, MCP guides?) #126 contains some of this
- [ ] Fix operator rendering on CRD viewer
- [ ] Syntax highlighting is wonky

tr0njavolta:
Closing in favor of https://github.com/orgs/modelplaneai/projects/7

PR #122, opened 2026-06-10
docs: Update with CRD reference

tr0njavolta:

This PR adds a CRD reference for the APIs in the root of this project. It mirrors the Crossplane API reference with some UI changes.

It provides an example and returns spec & status in tabs rather than inline. This approach should help visualizing objects while still having context within the CRD.

Fixes #

I have: 

- [ ] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~

### DIFF
--- a/docs/data/apis/examples/inferenceclasses.yaml
+++ b/docs/data/apis/examples/inferenceclasses.yaml
@@ -1,0 +1,19 @@
+manifest: |
+  apiVersion: modelplane.ai/v1alpha1
+  kind: InferenceClass
+  metadata:
+    name: h100-80gb-8
+  spec:
+    description: "8x H100 80GB (A3 / p5)"
+    resources:
+      gpu:
+        count: 8
+        memory: "80Gi"
+    provisioning:
+      provider: GKE
+      gke:
+        machineType: a3-highgpu-8g
+        diskSizeGb: 500
+        accelerator:
+          type: nvidia-h100-80gb
+          count: 8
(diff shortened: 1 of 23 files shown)


tr0njavolta:
Preview in my fork https://modelplane-docs-preview-git-docs-upboundio.vercel.app/reference/

docs/data/apis/examples/inferenceclasses.yaml
```diff
@@ -0,0 +1,19 @@
+manifest: |
```
negz, line 1: Would it be possible to:

1. Have these live at `examples/` or `docs/examples`
2. Make them regular YAML files (not an opaque string in a YAML `manifest` key)

If we could do that then we could use one set of examples for the reference docs and elsewhere (e.g. folks browsing the repo). They'd become files folks could `kubectl apply -f http://...`.

negz:
@tr0njavolta It looks like these work on your branch preview but not on the preview Vercel linked. Any idea why?

tr0njavolta:
@negz The vercel deployment was initially looking at the docs directory at the root and the CRD generator needed to mount the api/ directory above. The new preview has a new mount point where I manually moved the files.

negz (approved):
Whoops I wrote this comment earlier but didn't hit submit.

Not a blocker - we could refactor later.

docs/data/apis/examples/inferenceclasses.yaml
negz: It'd also be nice if these YAML files had syntax highlighting, but also not a blocker.

tr0njavolta: I'll add manifest hosting to #117

PR #107, opened 2026-06-05
docs: Adds foundational docs site

tr0njavolta:
### Description of your changes

This PR builds a Hugo framework docs site that mirrors the Crossplane docs theme and setup. All assets were taken from the website/artwork repos and the current content was taken from the modelplane repo.

Notable changes from the Crossplane docs include fonts, image assets, removing versioning (for now), CNCF and Linux Foundation footer removal, Socials links removal, and Navbar links removal. All can be re-added before launch.

Fixes #

I have: 

- [ ] ~Read and followed Modelplane's [contribution process](https://github.com/modelplaneai/modelplane/blob/main/CONTRIBUTING.md).~
- [ ] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~
- [ ] ~Added or updated tests covering any composition function changes.~
- [X] Signed off every commit with `git commit -s`.

PR #123, opened 2026-06-10
Build and deploy the docs site with Nix

negz:

#107 added the docs site. This is the follow-up that wires it into Nix, so it builds and deploys the same way the rest of the project does.

The day-to-day is two commands:

- `nix run .#docs-serve` — preview locally with live reload at http://localhost:1313
- `nix build .#docs` — build the production site into `result/`

`nix flake check` now builds the site too, so a broken site fails CI like anything else. You don't need to install Hugo, Node, or the PostCSS plugins; Nix provides them at pinned versions. If you've never used Nix, the [CONTRIBUTING.md](https://github.com/modelplaneai/modelplane/blob/main/CONTRIBUTING.md) setup applies unchanged — `nix develop`, or `./nix.sh` if you don't have Nix locally.

A third command, `nix run .#docs-generate`, rebuilds the committed webpack JavaScript bundle; you only need it after changing files under `docs/utils/webpack/src/`.

Vercel builds the site with the same `.#docs` derivation that CI verifies: it installs Nix into its Amazon Linux build image and runs the build, so the deployed site is exactly what `nix flake check` checks. `docs/vercel-build.sh` does the install-and-build and `vercel.json` points the build at it. Vercel's GitHub app keeps driving deploys, including previews on fork PRs.

The second commit bumps the PostCSS dependencies, which were several majors behind as inherited from crossplane/docs, so Dependabot doesn't immediately flag them.

I have:


tr0njavolta (approved):
I was able to run and build everything per the contributing guide. Might be nice for new nix users or users only contributing to the docs to know to run this in the project root, not the docs directory but that's a minor nit.

tr0njavolta (approved):

PR #137, opened 2026-06-11
Propose and implement unopinionated ModelDeployment engines

### LINKED ISSUE #52: Replica topology and capability-based scheduling

Modelplane can't deploy multi-node models. Kimi K2, DeepSeek V3, Llama 4 Behemoth, and similar frontier models require deployment shapes the current schema can't express — multiple nodes coordinating via high-bandwidth interconnect, structured parallelism strategy, hardware capability requirements (FP8 support, NVLink, IB-400g+). The deploy function divides total VRAM by per-GPU memory to compute GPU count. For Kimi K2 this gives a number but not a deployable configuration. There's no way to tell the scheduler that 16 GPUs need to be 2 nodes of 8 with TP=8 + PP=2, not 1 node of 16 (which doesn't exist) or 16 nodes of 1 (wrong sharding).

Frontier models share a deployment shape. Kimi K2 (1T/32B MoE) needs 16 GPUs across 2 H200 nodes connected by IB-Quantum-2, running TP=8 + PP=2 with expert parallelism enabled, FP8 quantization, 141Gi per GPU. DeepSeek V3 has similar requirements. Llama 405B at full precision spans multiple nodes. Multi-node deployment with structured parallelism and hardware-specific requirements is the production pattern for serious models, not an edge case.

KServe `LLMInferenceService` already supports this. The `parallelism` block expresses tensor and pipeline parallelism structurally. The `workerNodeSize` field tells KServe to compose a multi-pod deployment across nodes via LeaderWorkerSet. Each pod gets its share of GPUs; KServe handles the Ray coordination. We just don't expose any of this through the Modelplane API today.

The API needs three things: a way to express the shape of a replica, a way to declare what hardware capabilities pools have, and a matching algorithm that puts them together correctly.

I think the natural place for shape is a `replicaTopology` block on the serving profile. The block describes one independent serving instance — what hardware it needs, how it shards across nodes, what placement constraints apply. The current `resources.vram` field is removed (subsumed by the topology block).

```yaml
kind: ClusterModel
metadata:
  name: kimi-k2-instruct
spec:
  model:
    name: moonshotai/Kimi-K2-Instruct
  source: HuggingFace
  huggingFace:
    repo: moonshotai/Kimi-K2-Instruct

  environmentSelector:
    matchLabels:
      modelplane.ai/tier: production

  serving:
  - name: vllm-h200-multinode
    replicaTopology:
      nodes: 2
      gpusPerNode: 8
      parallelism:
        tensor: 8
        pipeline: 2
        expert: enabled
      requires:
        memoryPerGpu: "141Gi"
        interconnect: nvlink
        multiNodeBandwidth: ib-400g-or-better
        precisionSupport: ["fp8"]
        nvlinkDomainSize: 8
      placement:
        podsAcrossNodes: required
        replicasAcrossNodes: preferred
    engine:
      name: vLLM
      image: vllm/vllm-openai:v0.8.0
      args:
      - "--tensor-parallel-size=8"
      - "--pipeline-parallel-size=2"
      - "--enable-expert-parallel"
      - "--quantization=fp8"
      - "--max-model-len=65536"
      - "--gpu-memory-utilization=0.90"
      - "--enable-prefix-caching"
      - "--max-num-seqs=256"
      - "--distributed-executor-backend=ray"
```

The `replicaTopology` block has clear sub-fields. `nodes` and `gpusPerNode` define the deployment shape — KServe's `workerNodeSize` comes from `nodes`, and the per-pod GPU resource limit comes from `gpusPerNode`. `parallelism` populates KServe's structured parallelism block; this duplicates information that also appears in engine args, but the duplication is unavoidable because KServe needs structured fields for LeaderWorkerSet composition while vLLM needs the args. The serving profile author keeps them consistent.

`requires` is a list of capability constraints. The deploy function matches these against pool capabilities. Numeric constraints use ≥ comparison (`memoryPerGpu: 141Gi` means at least 141Gi). Enum constraints use ordered comparison (`multiNodeBandwidth: ib-400g-or-better`). List constraints use superset comparison (`precisionSupport: [fp8]` means the pool must support FP8 among its supported formats). These comparisons are simple enough to verify by inspection; v0.1 doesn't try to handle multi-dimensional capability comparison or "or-better" beyond a single ordered enum.

`placement` controls pod and replica spread. `podsAcrossNodes: required` is correctness for multi-node replicas — the parallelism doesn't work if both pods land on the same node. `replicasAcrossNodes: preferred` is HA — multiple replicas should land on different nodes if capacity allows, but can pack if needed.

For the InferenceEnvironment side, node pools need a `capabilities` block declaring what their hardware supports. The platform engineer provisioning pools knows this; today's schema doesn't have a place to put it.

```yaml
kind: InferenceEnvironment
metadata:
  name: prod-coreweave-us-east
  labels:
    modelplane.ai/region: us-east
    modelplane.ai/tier: production
spec:
  cluster:
    source: Existing
    existing:
      secretRef:
        name: cw-cluster-kubeconfig
      provider: coreweave
      instanceType: HGX-H200-IB

  nodePools:
  - name: medium-models
    acceleratorType: nvidia-h100-80gb
    acceleratorCount: 8
    nodeCount: 0
    maxNodeCount: 8
    labels:
      modelplane.ai/pool: medium
    capabilities:
      memoryPerGpu: "80Gi"
      interconnect: nvlink
      precisionSupport: ["fp16", "bf16", "fp8"]
      multiNodeCapable: false
      nvlinkDomainSize: 8

  - name: frontier-multinode
    acceleratorType: nvidia-h200-141gb
    acceleratorCount: 8
    nodeCount: 0
    maxNodeCount: 4
    labels:
      modelplane.ai/pool: frontier
    capabilities:
      memoryPerGpu: "141Gi"
      interconnect: nvlink
      precisionSupport: ["fp16", "bf16", "fp8"]
      multiNodeCapable: true
      interNodeBandwidth: ib-400g
      nvlinkDomainSize: 8
```

The capabilities block is the hardware truth-teller. Pool memory per GPU, interconnect type, supported precisions, multi-node networking. The `multiNodeCapable` field is a discriminator — pools without it fail multi-node profile matching even if they have the right GPU type, because their networking isn't configured for it.

`provider` and `instanceType` on the cluster are infrastructure context. They don't participate in matching at v0.1, but they enable richer matching later (e.g., recipe-based validation that says "this configuration is validated on Coreweave HGX-H200-IB"). Operators authoring their own InferenceEnvironments can leave these blank.

Profile matching gains capability filtering. The deploy function walks profiles in order; for each profile, walks pools in order; first pool whose `capabilities` satisfy the profile's `replicaTopology.requires` and has capacity for `nodes × gpusPerNode` allocation units wins.

```
For each environment matching ClusterModel.environmentSelector:
  For each profile in ClusterModel.serving:
    For each pool in environment.nodePools:
      If pool.capabilities satisfies profile.replicaTopology.requires:
        If pool has capacity for profile.replicaTopology.nodes nodes
           with profile.replicaTopology.gpusPerNode GPUs each:
          Match found.
          Compose ModelPlacement with this profile and pool.
          Generate LLMInferenceService:
            replicas: from ModelDeployment scaling
            workerNodeSize: profile.replicaTopology.nodes
            parallelism: profile.replicaTopology.parallelism
            template.nodeSelector: pool.labels
            template.affinity: from profile.replicaTopology.placement
            container.args: profile.engine.args
            container.resources.limits[nvidia.com/gpu]: profile.replicaTopology.gpusPerNode
```

The matching algorithm is deterministic given declaration order. Operators control behavior by ordering profiles (priority) and pools (preference). Two pools matching the same profile in the same environment is resolved by pool declaration order; v0.1 doesn't try to be clever about cost-aware or capability-conservative selection. v0.2+ can add this.

For Kimi K2 specifically, the matching cascade looks like: filter environments by `tier: production` (the production environment matches), walk profiles (only `vllm-h200-multinode`), walk pools (medium-models fails on multiNodeCapable=false; frontier-multinode satisfies all requirements), match found, compose LLMInferenceService with `workerNodeSize: 2`, `parallelism: {tensor: 8, pipeline: 2}`, anti-affinity required across nodes, GPU limit 8 per pod, nodeSelector targeting the frontier pool's labels.

The composed LLMInferenceService:

```yaml
apiVersion: serving.kserve.io/v1alpha1
kind: LLMInferenceService
metadata:
  name: kimi-k2-instruct
  namespace: ml-team-research
spec:
  model:
    uri: hf://moonshotai/Kimi-K2-Instruct
  replicas: 1
  workerNodeSize: 2
  parallelism:
    tensor: 8
    pipeline: 2
  template:
    spec:
      nodeSelector:
        modelplane.ai/pool: frontier
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchLabels:
                app: kimi-k2-instruct
            topologyKey: kubernetes.io/hostname
      containers:
      - name: server
        image: vllm/vllm-openai:v0.8.0
        args:
        - --model=moonshotai/Kimi-K2-Instruct
        - --tensor-parallel-size=8
        - --pipeline-parallel-size=2
        - --enable-expert-parallel
        - --quantization=fp8
        - --max-model-len=65536
        - --gpu-memory-utilization=0.90
        - --enable-prefix-caching
        - --max-num-seqs=256
        - --distributed-executor-backend=ray
        resources:
          limits:
            nvidia.com/gpu: 8
```

KServe creates a LeaderWorkerSet with 2 worker pods per replica. The cluster autoscaler provisions H200 nodes from the frontier pool. Pods schedule one per node (anti-affinity required). vLLM starts in each pod, Ray sets up communication over IB, model loads with TP=8 within each node and PP=2 across the two. LLMInferenceService becomes ready.

The same API handles single-node deployments. A Gemma 3 27B profile sets `nodes: 1, gpusPerNode: 1, requires.memoryPerGpu: 40Gi, requires.precisionSupport: [fp8]`. Match against a medium-models pool (H100, supports FP8, has 80Gi available). Compose LLMInferenceService with `workerNodeSize: 1` (no LeaderWorkerSet, just a regular Deployment), GPU limit 1, nodeSelector for the medium pool. Single-node deployments are the special case where multi-node machinery isn't needed; the schema accommodates both uniformly.

This proposal sketches one shape but the design space is meaningfully bigger than what I've laid out, and getting it wrong is expensive — schema decisions made here will be hard to change once operators have ClusterModels in production. This needs a full design that accounts for the different topology patterns and the tradeoffs between them.

negz:

Fixes #52.

Today `spec.workers.topology` describes a deployment with parallelism axes (`tensor`, `pipeline`, `data`, `dataLocal`). Each axis derives and injects engine-specific flags - breaking the pass-through property Modelplane relies on everywhere else, and creating two sources of truth (the user writes the same flags in `args`).

I propose we describe a deployment by its shape instead. `spec.engines` is an array of inference engines, each a `Standalone` member or a `Leader` and one or more `Worker` members. A member carries its own `nodeSelector` and engine template; an engine may be stamped out a fixed number of times with `copies`.

A small model on a single GPU shows the API at its simplest:

```yaml
kind: ModelDeployment
metadata:
  name: qwen3-8b
  namespace: ml-team
spec:
  replicas: 1
  clusterSelector:
    matchLabels:
      modelplane.ai/tier: production
  engines:
  - name: qwen3-8b
    members:
    - role: Standalone
      nodeSelector:
        devices:
        - name: gpu
          count: 1
          selectors:
          - cel: device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("16Gi")) >= 0
      template:
        spec:
          containers:
          - name: engine
            image: vllm/vllm-openai:v0.11.0
            args:
            - --model=Qwen/Qwen2.5-7B-Instruct
```

This PR ports today's functionality onto the new shape. The scheduler co-schedules a replica's engines onto one cluster, each on a pool that satisfies its members' selectors. `compose-model-replica` composes a workload per engine - a Deployment for a Standalone, a LeaderWorkerSet for a gang - fronted by one Service and HTTPRoute. Modelplane injects no engine flags; the only injection is `MODELPLANE_LEADER_ADDRESS`, a backend-neutral alias for the gang leader's address.

Prefill/decode disaggregation (`spec.serving` and `PrefillDecode`) is designed here but not implemented.

Both paths were validated end to end on EKS L4 GPUs: a Standalone engine on one GPU, and a Leader/Worker gang serving Qwen2.5-7B pipeline-parallel across two single-GPU nodes, with completions flowing through the control-plane gateway. The multi-node example is the validated manifest. The run also filed #139, #140, and #141 for pre-existing issues it surfaced.

I have: 


design/unopinionated-deployments.md
```diff
@@ -0,0 +1,692 @@
...
+(`spec.serving`). Modelplane composes a ModelReplica per `spec.replicas` and the
+fleet scheduler places each on a cluster.
+
+Under this design inference engine configuration is opaque to Modelplane, and
+engines are deployed as flexibly shaped groups.
+
+As a result of these choices:
+
+- Modelplane can broadly and automatically support new inference engines.
+- Modelplane can broadly and automatically support new inference toplogies.
```
dennis-upbound: repeated

dennis-upbound (approved):
i like it!

examples/deployment/generic.yaml
```diff
@@ -0,0 +1,163 @@
...
+  #
+  # This example disaggregates serving into two 2-node engines: a prefill gang
+  # (compute-bound) and a decode gang (bandwidth-bound). Each gang is TP=8 per
+  # node, PP=2 across two nodes. Within each gang the leader runs the Ray head
+  # and serves; the follower joins it.
+  workers:
+
+  # Prefill gang.
+  - name: prefill-leader
+    gang: prefill
```
dennis-upbound: needs revving based on latest design?

examples/deployment/generic.yaml
```diff
@@ -0,0 +1,163 @@
+apiVersion: modelplane.ai/v1alpha1
```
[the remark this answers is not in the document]
negz: This manifest was committed by mistake - I'll remove it.

dennis-upbound (approved):

dennis-upbound:
Two things worth pinning while `serving` is still fluid, both about how far it stretches:

**EPD is coming, and `serving.disaggregation` is 2-role-shaped.** The block names exactly `prefillGroupName` + `decodeGroupName`, and the role-labeling/routing that hangs off it assumes two phases. But encode-prefill-decode (a third "encode" stage for multimodal — vision/audio encoders on their own GPUs) is already shipping: SGLang has elastic encoder disaggregation ([LMSYS, Jan 2026](https://www.lmsys.org/blog/2026-01-12-epd/)), llm-d-inference-scheduler is extending P/D→EPD ([#608](https://github.com/llm-d/llm-d-inference-scheduler/issues/608)), and the routing sidecar already takes an `--ec-connector` for the encoder→prefiller hop. `workers[]` is N groups so the *shape* holds, but the 2-named-group block would need a third role (or a phases list) to express EPD. Cheaper to leave the door open now than to version the block later.

(More #34 than #137, flagging anyway: KV transfer is shifting from direct prefill→decode to tiered KV *stores* — Mooncake/LMCache, now NIXL backends ([Mooncake](https://github.com/kvcache-ai/Mooncake/), [LMCache](https://github.com/lmcache/lmcache)). The connector's a user flag, but a shared store is a stateful component with no home in `workers[]`.)

examples/deployment/generic.yaml
negz: I had meant to delete this - will do that.

design/unopinionated-deployments.md
negz: The last word is intentionally different in each bullet.

negz:
@dennis-upbound Good point, I did read a little about EPD while working on this. The good part is the encode stage should model cleanly under `workers` as a third group. Maybe we switch from `Unified/Disaggregated` to an enum that describes how it's disaggregated? So:

```yaml
spec:
  serving:
    mode: DisaggregatedPD
    pd:
      # ...
```

```yaml
spec:
  serving:
    mode: DisaggregatedEPD
    epd:
      # ...
```

design/unopinionated-deployments.md
```diff
@@ -0,0 +1,731 @@
...
+
+- `name`: identifies the group.
+- `replicas`: how many copies of this group to run per `ModelReplica`.
+- `members`: the group's pods. A group is either a single `Standalone` member or
+  a `Leader` and a `Worker`; no other combination is valid.
+
+Each member has:
+
+- `role`: `Standalone` (default), `Leader`, or `Worker`.
+- `count`: follower pods, for a `Worker` only. Each follower is one pod on one
```
negz: Given this can only apply to workers I want to nest it:

```yaml
members:
- role: Leader
  nodeSelector: {...}
  template: {...}
- role: Worker
  worker:
    count: 2  # Or nodes?
  nodeSelector: {...}
  template: {...}
```

design/unopinionated-deployments.md
```diff
@@ -0,0 +1,731 @@
...
+  node, so `count` is also the number of follower nodes. Defaults to 1.
+- `nodeSelector`: the per-node device request — what devices the member's engine
+  pod needs. Its GPU `count` is the GPUs per node.
```
negz: I think I should move `nodeSelector` to be a group level concern.

It's equivalent for `Standalone` members. For `Leader`/`Worker` members I don't think you ever really want them to land on different pools, and I doubt you ever want them to have heterogenous hardware.

design/unopinionated-deployments.md
negz: Done.

design/unopinionated-deployments.md
negz: Done

docs/content/concepts.md
```diff
@@ -103,60 +103,59 @@ clusters, which Modelplane assumes are solely for its use.
...
-`workers` becomes the decode role and `prefill` is its own self-contained role,
-with its own `workers.count`, `topology`, `template`, and `nodeSelector`, so
-each phase can land on a different GPU class through the usual capability
-matching. The prefill engine transfers its KV cache to a decode engine over
-NIXL, configured through the engine's `--kv-transfer-config`. Like multi-node
-serving, a disaggregated deployment requires a [ModelCache](#modelcache); both
-roles mount the same PVC and stay co-located on one cluster, since KV transfer
-needs a fast interconnect (NVLink within a node, RDMA across nodes).
+When a model is too large to fit on one node's GPUs, make an engine a gang:
+give it a `Leader` member and a `Worker` member spanning `worker.nodes` nodes.
```
negz: This phrasing feels off. It's conceptually worker _members_ despite appearing in the array once. `worker.nodes` is really a convenience so you don't have to write out N identical worker member entries in the array.

examples/deployment/model-deployment-multinode.yaml
```diff
@@ -1,48 +1,99 @@
-# A ModelDeployment using multi-node tensor + pipeline parallelism.
+# A ModelDeployment serving one model across two nodes.
 #
-# pipeline > 1 runs the model across multiple nodes, each with `tensor`
-# GPUs. The scheduler picks a cluster with a pool that has at least
-# `tensor` GPUs per node and at least `pipeline` available nodes.
-# Modelplane composes a LeaderWorkerSet-backed llm-d serving instance on
-# the remote cluster.
+# An engine with a Leader and a Worker spans multiple nodes: the gang size is
+# the leader plus the Worker's nodes. The scheduler picks a cluster with a
```
negz: Phrasing feels off here too. It's not a leader and a multi-node worker, but a leader and `worker.nodes` worth of workers.

apis/modeldeployments/definition.yaml
```diff
@@ -44,254 +44,341 @@ spec:
...
+                                            minLength: 1
+                                            maxLength: 10240
+                          worker:
+                            type: object
+                            description: >-
+                              Settings for a Worker member only. No schema
+                              default - the apiserver applies defaults before
+                              CEL validation, so defaulting this object would
+                              inject it onto Standalone and Leader members and
+                              trip the rule that forbids it.
```
negz: This note about defaulting doesn't seem relevant to the end user.

docs/content/_index.md
```diff
@@ -47,7 +55,8 @@ Modelplane draws a clear boundary between two teams.
 organizational metadata via labels on clusters: tier, region, provider.
 
 **ML teams** create a `ModelDeployment` carrying everything needed to serve a
-model: the worker template, hardware topology, and replica count. Modelplane
+model: the inference engines (their members' templates and device requests)
```
negz: I'd just say "their templates and device requests here" - gloss over the fact that an engine can have multiple members at this level.

docs/content/concepts.md
```diff
@@ -103,60 +103,62 @@ clusters, which Modelplane assumes are solely for its use.
...
-2. Derives the physical shape from `workers.topology`: GPUs per node (tensor)
-   and nodes per worker (pipeline, default 1).
-3. Matches each `nodeSelector` device request against a candidate pool's
-   InferenceClass devices, gated on the pool having enough available nodes, and
-   pins the replica to a pool that satisfies every request.
-4. Creates a `ModelReplica` for each selected cluster, carrying the resolved
-   `claim: DRA` requests so the replica forms a DRA `ResourceClaim`.
+2. Derives each member's node cost: a Standalone or Leader is one pod, a
+   Worker is its `worker.nodes` pods, times the engine's `copies`. A member
+   that claims no devices costs no nodes.
```
negz: Here and elsewhere: conceptually a `role: Worker` member with multiple nodes is workers (plural), not a worker.

functions/compose-model-deployment/function/fn.py
```diff
@@ -217,15 +202,55 @@ def compose_replicas(self, matched):
...
+        nodeSelector, or only synthetic devices matched - carries only its pool
+        pin; the scheduler guarantees at least one member of every engine
+        carries requests.
+        """
+        members = []
+        for member, mp in zip(engine.members, placement.members, strict=True):
+            replica_member = mrv1alpha1.Member(
+                role=member.role,
+                nodePoolName=mp.pool,
+                template=mrv1alpha1.Template.model_validate(member.template.model_dump(exclude_none=True)),
```
negz: Is `exclude_none` definitely correct? Typically in the SDK we use `exclude_unset`.

dennis-upbound:
`_place_engine_split` splits an engine's members across pools when no single pool fits them all, and the `_place_engine` comment calls that "silently degrade interconnect." For a multi-node gang I think it's worse than degrade.

A Leader+Worker engine doing TP/PP/EP talks over its pool's fabric — NVLink within a node, InfiniBand within a pool. Put the leader on one pool and the worker on another and that fabric isn't there: cross-pool tensor parallel never forms and the gang just sits NotReady. The split hands back a placement that can't run, not a slower one, with no signal saying why.

Would it be safer to fail closed? If an engine has more than one claiming member and no single pool fits it, reject with InsufficientCapacity instead of splitting. A false reject is recoverable — add capacity, or size the gang to a pool — but a gang split across fabrics hangs silently. The claimless-coordinator ride-along you describe is fine as-is (it stays on a sibling's pool, one gang one fabric); it's specifically splitting members that each claim their own GPUs across pools that I'd guard.

dennis-upbound:
Looks great!

PR #138, opened 2026-06-12
Implement unopinionated ModelDeployment workers

negz:

Implements the design proposed in #137.

`ModelDeployment.spec.workers` modeled a topology of tensor/pipeline axes, from which Modelplane derived engine flags and a Ray bootstrap. That coupled it to per-engine knowledge, left the parallelism flags written in two places, and couldn't express data or expert parallelism.

This replaces topology with an array of worker groups, each a Standalone member or a Leader plus one or more Workers. The member carries its nodeSelector and engine template; parallelism and KV transfer live in the engine's commands and args, which pass through verbatim. The only injection is `MODELPLANE_LEADER_ADDRESS`, a backend-neutral alias for the gang leader's address. The scheduler co-schedules a replica's groups onto one cluster, each on a satisfying pool, and compose-model-replica composes a Deployment or LeaderWorkerSet per group behind one Service and HTTPRoute.

This ports today's functionality onto the new shape. Prefill/decode disaggregation (`spec.serving`) is left for a follow-up; unified serving is the only behavior.

I have:


negz:
Rolled into #137 alongside the design it implements, as separate commits.

PR #142, opened 2026-06-13
Disaggregated serving on the unopinionated engine shape

dennis-upbound:
Towards #34.

PR #137 reshaped `spec.workers` into named engine groups and composes each engine's workload plus a unified Service/HTTPRoute, but deferred disaggregated prefill/decode routing. This adds it on that shape.

A deployment opts in with `serving.mode: PrefillDecode` and marks each engine's `phase` as `Prefill` or `Decode` (CEL-checked: exactly two engines, one of each phase, and `phase` set only under `PrefillDecode`). This follows the merged design — the engine carries its own phase rather than `serving` naming engines, so a phase can't dangle. The scheduler copies `serving` and the phases onto the ModelReplica. The replica backend, when disaggregated, finds the prefill and decode engines by phase, role-labels their serving pods (`llm-d.ai/role`), fronts the decode engine with the pd-sidecar (the sidecar takes port 8000 and forwards to the engine on its own `--port`, with `--secure-proxy=false` for the HTTP gateway path), replaces the unified Service with an `InferencePool` over both engines, and points the HTTPRoute at the pool behind a hardcoded endpoint picker.

Routing an HTTPRoute to an `InferencePool` backendRef needs the serving cluster to run the substrate #137 stripped: the Envoy AI Gateway and the Gateway API Inference Extension, with Envoy Gateway delegating `InferencePool` resolution to the AI Gateway's ext-proc server. So compose-serving-stack installs the AI Gateway (CRDs and controller) and the GAIE InferencePool CRD, and restores Envoy Gateway's `extensionManager`. Envoy Gateway moves from v1.3.0 to v1.8.1, the released version the AI Gateway is tested against and the first whose chart bundles the `ListenerSet` CRD the newer gateway requires; v1.3.0 rejected `InferencePool` backendRefs outright (`InvalidKind`).

Engine flags, including the NixlConnector `--kv-transfer-config`, stay the user's per #137; this injects none. The routing layer is one module (`backends/disagg.py`) applied as a post-process over #137's per-engine workloads, so the unified path is untouched.

Validated end to end on a GKE cluster (L4s): the deployment schedules both engines through DRA; both vLLM engines serve and bring up NixlConnector (`NIXL is available`, `kv_role=kv_both`, from the user's flags); the endpoint picker reconciles the InferencePool over both endpoints; and a request through the gateway to the `InferencePool` backendRef (`ResolvedRefs=True` under v1.8.1) returns a completion from the disaggregated pair.

I have:


apis/modeldeployments/definition.yaml
```diff
@@ -26,6 +26,17 @@ spec:
...
+            # the count and the names against the mode. serving is optional and
+            # defaults to Unified, so each rule no-ops unless serving is set and
+            # Disaggregated.
+            x-kubernetes-validations:
+            - rule: "!has(self.serving) || self.serving.mode != 'Disaggregated' || size(self.engines) == 2"
+              message: "disaggregated serving requires exactly two engines (prefill and decode)"
+            - rule: "!has(self.serving) || self.serving.mode != 'Disaggregated' || self.engines.exists(e, e.name == self.serving.disaggregation.prefillGroupName)"
+              message: "serving.disaggregation.prefillGroupName must name an engine"
+            - rule: "!has(self.serving) || self.serving.mode != 'Disaggregated' || self.engines.exists(e, e.name == self.serving.disaggregation.decodeGroupName)"
+              message: "serving.disaggregation.decodeGroupName must name an engine"
```
negz: @dennis-upbound we switched this last minute sorry. We went with your suggestion of noting which worker is which using enums on the worker blocks - e.g. `phase: Decode`. Take a look at the final merged design.

apis/modeldeployments/definition.yaml
```diff
@@ -57,6 +68,49 @@ spec:
...
+                  engines. Unified (the default) fronts the engines with a
+                  Service. Disaggregated splits prefill and decode across two
+                  engines, named below, fronted by a GAIE InferencePool and an
+                  endpoint picker that sequences prefill then decode; Modelplane
+                  role-labels each engine's pods, injects the routing sidecar,
+                  and composes the picker. The gateway and picker are an
+                  implementation detail, not named here. Omitted means Unified.
+                required: [mode]
+                x-kubernetes-validations:
+                - rule: "self.mode != 'Disaggregated' || has(self.disaggregation)"
```
negz: We also went for `PrefillDecode` here (not `Disaggregated` to leave room for EPD.

apis/modeldeployments/definition.yaml
dennis-upbound: no worries will address

apis/modeldeployments/definition.yaml
dennis-upbound: cool lemme rebase!

functions/compose-model-deployment/tests/test_fn.py
```diff
@@ -855,3 +855,27 @@ async def test_compose(self) -> None:
                     json_format.MessageToDict(got),
                     "-want, +got",
                 )
+
+    async def test_serving_copied_to_replica(self) -> None:
```
negz: Shouldn't this be a case in the preceding table?

apis/modeldeployments/definition.yaml
dennis-upbound: Done. Each engine now marks its `phase` (Prefill/Decode) and the backend finds prefill/decode by phase; dropped the `serving.disaggregation` block. CEL checks exactly two engines, one of each phase, and `phase` only under `PrefillDecode`.

apis/modeldeployments/definition.yaml
dennis-upbound: Done. `mode` is `Unified | PrefillDecode` across the XRDs, generated models, backend, and tests.

functions/compose-model-deployment/tests/test_fn.py
dennis-upbound: Done. Folded into the `test_compose` table as a Case with the full-response golden, and dropped the standalone method.

apis/modeldeployments/definition.yaml
```diff
@@ -57,6 +70,27 @@ spec:
...
+                type: object
+                description: >-
+                  How the deployment is served from the cluster edge to its
+                  engines. Unified (the default) fronts the engines with a
+                  Service. PrefillDecode splits prefill and decode across two
+                  engines - each marking its phase - fronted by a GAIE
+                  InferencePool and an endpoint picker that sequences prefill
+                  then decode; Modelplane role-labels each engine's pods, injects
+                  the routing sidecar, and composes the picker. The gateway and
+                  picker are an implementation detail. Omitted means Unified.
```
negz: The tail end of this feels like fungible implementation details - not user facing API docs.

functions/compose-model-replica/function/backends/disagg.py
```diff
@@ -0,0 +1,299 @@
...
+_DECODE_ENGINE_PORT = 8001
+
+# Selector labels shared by both engines' serving pods (the InferencePool
+# matchLabels) and the per-role label the picker partitions on.
+_LABEL_ROLE = "llm-d.ai/role"
+_LABEL_INFERENCE_SERVING = "llm-d.ai/inference-serving"
+
+# EndpointPickerConfig for the disaggregated profile. The apiVersion must be the
+# GIE group the EPP binary registers (inference.networking.x-k8s.io/v1alpha1); an
+# unregistered group crash-loops the picker.
```
negz: This comment feels like the model narrating an edge case it hit. Not sure it's useful.

functions/compose-model-replica/function/backends/disagg.py
```diff
@@ -0,0 +1,299 @@
...
+        "epp-serviceaccount": base.wrap_object(provider_config, sa),
+        "epp-role": base.wrap_object(provider_config, role),
+        "epp-rolebinding": base.wrap_object(provider_config, binding),
+        "epp-config": base.wrap_object(provider_config, config),
+        "epp": base.wrap_object(provider_config, deployment, cel_query=base.AVAILABLE_CEL),
+        "epp-service": base.wrap_object(provider_config, service),
+    }
+
+
+def apply(
```
negz: Nit: Seems like disagg isn't actually a backend in the same sense as the other two. Should it be represented as some other concept?

functions/compose-model-replica/function/routing.py
```diff
@@ -0,0 +1,299 @@
...
+    serve on a different port; defaults to _DECODE_ENGINE_PORT when no --port is
+    set.
+    """
+    args = engine.get("args", [])
+    for i, a in enumerate(args):
+        if a.startswith("--port="):
+            return int(a.split("=", 1)[1])
+        if a == "--port" and i + 1 < len(args):
+            return int(args[i + 1])
+    return _DECODE_ENGINE_PORT
```
negz, line 152: Do all engines use `--port`? This flag parsing feels a bit fragile. Given we own the routing layer I wonder if we should just document that we expect the engine to listen on a specific port.

(Though I suppose we could do both - document it but also try infer it best-effort.)

functions/compose-model-replica/function/backends/disagg.py
```diff
@@ -0,0 +1,299 @@
...
+    workloads (Deployment/LWS) composed by the backends are reused as-is apart
+    from the label/sidecar post-process.
+    """
+    name = replica.metadata.name
+    prefill = next(e for e in replica.spec.engines if e.phase == "Prefill")
+    decode = next(e for e in replica.spec.engines if e.phase == "Decode")
+
+    out = dict(composed)
+    out.pop(base.SERVICE_KEY, None)
+    out.pop(base.ROUTE_KEY, None)
```
negz: I don't love that this removes the base's routing.

Could we handle the base/Unified routing concern the same way we handle it here - i.e. with an `apply` (`apply_routing`?) that adds it on. Then we wouldn't need to filter it here: we'd instead build the appropriate base backend then decorate it with the appropriate routing stack.

negz (commented):
Looks good overall @dennis-upbound. Only blocker is the comment about the new `apply` method _removing_ the base routing config. I'd prefer to start with no routing config then pick one to add.

functions/compose-model-replica/function/backends/disagg.py
dennis-upbound: Done. The backends build engines only now; a routing layer (`function/routing.py`) decorates them with the surface `serving.mode` picks — `apply` adds a Service for Unified, or the InferencePool + endpoint picker for PrefillDecode. Nothing is stripped anymore.

functions/compose-model-replica/function/backends/disagg.py
dennis-upbound: Done. Moved it out of `backends/` into `function/routing.py` as a routing strategy alongside the unified one. native/llm-d/dynamo stay the workload backends.

functions/compose-model-replica/function/routing.py
dennis-upbound: Done — did both. The decode engine is documented as expected on a fixed port (`_DECODE_ENGINE_PORT`), and we best-effort honor an explicit `--port` override rather than require the flag.

functions/compose-model-replica/function/backends/disagg.py
dennis-upbound: Done. Dropped the "crash-loops the picker" narration; the comment now just states the apiVersion is the GIE group the EPP registers.

apis/modeldeployments/definition.yaml
dennis-upbound: Done. Trimmed to user-facing intent (Unified → Service, PrefillDecode → inference-aware prefill/decode routing); dropped the InferencePool/sidecar/picker impl details.

negz (approved):

Issue #149, opened 2026-06-15
Scheduler can split a gang across interconnect fabrics


negz:
### What problem are you facing?

A multi-member engine (a `Leader` plus one or more `Worker`s) doing tensor or pipeline parallel talks over its pool's interconnect fabric — NVLink within a node, InfiniBand within a pool. The members of such a gang must share that fabric for the collective to form.

The scheduler places members per pool, but it has no concept of fabric. Pool identity is the finest grain it can reason about. So when an engine's members land on different pools, the scheduler can't tell whether those pools share a fabric or not. If they don't, the cross-pool collective never forms and the gang sits NotReady — a placement that can't run, with no signal saying why. A false reject is recoverable (add capacity, or size the gang to a pool); a gang silently scattered across fabrics is not.

This was raised by @dennis-upbound reviewing #137: https://github.com/modelplaneai/modelplane/pull/137#issuecomment-4699607997

There's a legitimate case the scheduler can't currently express safely: an operator exposes low-utilization, no-GPU nodes on the *same fabric* as the GPU nodes as a separate pool, and wants API/coordinator heads to schedule there while the GPU workers stay on the GPU pool — one fabric, two pools. Without a fabric concept the scheduler can only co-schedule these correctly by luck (e.g. exactly one GPU pool and one no-GPU pool); with two fabrics it could split the gang across them.

### How could Modelplane help solve your problem?

The scheduler should never produce a cross-fabric placement for one gang. Options considered, roughly in increasing order of capability:

1. **Reject any engine that doesn't fit a single pool** (capacity or geometry). Simplest and fully defensible given the scheduler can't reason about fabric — but it forecloses the same-fabric multi-pool case above.
2. **Model fabric explicitly.** Add a fabric / topology key to a cluster's published pools so the scheduler can constrain a split to one fabric, and allow cross-pool placement only within it. Correct and unlocks the same-fabric multi-pool case, at the cost of a status API addition the InferenceCluster source must populate.

The near-term change tracked separately removes cross-pool splitting (option 1). This issue tracks the underlying concern and the eventual fabric-aware placement (option 2) so the safe same-fabric multi-pool case isn't lost to the void.

negz:
> So when an engine's members land on different pools, the scheduler can't tell whether those pools share a fabric or not.

It also can't handle fabric that's more granular than pool level. e.g. I assume a pool that spans N zones isn't all on the same fabric.

negz:
[...]

PR #150, opened 2026-06-15
Place each engine on a single pool instead of splitting across pools

negz:

The scheduler placed an engine's members on one pool when it could, and split them across pools when no single pool fit. Splitting is unsafe, as [@dennis-upbound pointed out reviewing #137](https://github.com/modelplaneai/modelplane/pull/137#issuecomment-4699607997): the scheduler can't reason about interconnect fabric. Pool identity is the finest grain it has, so when a gang's members land on different pools it can't tell whether those pools share a fabric.

A Leader/Worker gang doing tensor or pipeline parallel talks over its pool's fabric — NVLink within a node, InfiniBand within a pool. Scatter its members across pools that don't share a fabric and the collective never forms: the gang sits NotReady with no signal saying why. Splitting doesn't degrade the interconnect, it removes it — handing back a placement that can't run rather than a slower one.

This change places every member of an engine on a single pool, or rejects the engine on that cluster. Rejection is safe and recoverable — the fill phase tries another cluster, and the deployment surfaces `InsufficientCapacity` if none fit — whereas a cross-fabric placement is a silent hang. Per-member `nodeSelector`s stay (a member can still claim different devices from the shared pool, and a claimless member rides along at zero node cost); they just no longer place members on different pools.

Co-scheduling a gang across pools that genuinely share a fabric — an operator exposing low-utilization nodes beside the GPU nodes as a separate pool, say — needs the scheduler to model fabric, which it doesn't yet. That's tracked in #149.

The change is concentrated in `_place_engine`, now reduced to "place the whole engine on the first single pool that fits, or reject." `_place_engine_split` and `_place_member` are gone. That's the part worth reviewer attention.

I have: 

### DIFF
--- a/functions/compose-model-deployment/function/scheduling.py
+++ b/functions/compose-model-deployment/function/scheduling.py
@@ -15,17 +15,18 @@ garbage-collects it, and the fill phase mints a fresh replica elsewhere to
 refill the deployment's replica count. Moving is always delete-plus-create,
 mirroring how Kubernetes treats a Pod whose node is gone.
 
-A replica is a set of engines co-scheduled onto one cluster. The unit of pool
-placement is the engine member: each member carries its own nodeSelector, so
-each is placed on a pool of that cluster that satisfies it. A gang's members
-usually want the same hardware, and they talk over the pool's fabric, so the
-scheduler prefers one pool that satisfies every member of an engine; only when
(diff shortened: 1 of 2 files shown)


dennis-upbound (approved):
lgtm

PR #152, opened 2026-06-15
Run docs Vale and htmltest as flake checks, not apps

negz:

Follow-up to #148, which added Vale prose linting and htmltest link checking for the docs as Nix apps (`nix run .#docs-vale` / `.#docs-htmltest`) run by a dedicated `docs.yml` workflow.

This PR moves both into `nix flake check`.

The repo has two ways to run tooling. **Apps** (`nix run .#...`) run on your machine with full network and filesystem access — good for things that have to reach the network or write files, like `docs-serve` or `docs-generate`. **Checks** (`nix flake check`) run in a sandbox: no network, only the inputs they declare, and they can't touch your working tree. That sandbox is the tradeoff. It's why the docs checks need a bit more plumbing than the apps did, and it's also the reason to prefer them: a check that passes for me passes for you and passes in CI, because nothing about the host can change the result.

Concretely, the move buys us:

- **One command for the whole repo.** Everything CI enforces runs under `nix flake check` — Python tests, lints, the docs site build, and now Vale and htmltest. No separate command or workflow to remember for docs.
- **Reproducibility.** No "works on my machine": the checks can't depend on what's installed, cached, or reachable.

It costs us:

- **No fast, targeted run.** With the apps gone, iterating on prose means `nix flake check`, which runs the whole suite rather than just Vale. If that friction bites we can add a thin app back alongside the check.
- **A pinned hash to maintain.** Vale downloads its style packages (Google, Microsoft, etc.) over the network, which a sandboxed check can't do. We sync them in a fixed-output derivation instead — Nix allows that one step network access in exchange for verifying the result against a hash. Those packages track "latest" upstream releases, so when one cuts a release the check fails with a hash mismatch and someone bumps `outputHash` in `nix/docs.nix` to the value Nix prints. `CONTRIBUTING.md` documents this.

One detail worth a look: htmltest checks links against the site **as laid out on disk**, so it can't run against the production build, whose links carry the `modelplane.ai/docs/` prefix. The check builds a second copy of the site with a `/` base URL for htmltest — the same base URL the old app used.

@tr0njavolta note the Vale `--minAlertLevel`: the check keeps your `warning` setting, so warnings and errors fail and suggestions don't. The app took the level as an argument, so you could opt into seeing suggestions locally (`-- --minAlertLevel=suggestion`). A check takes no arguments, so it's hardcoded to `warning` and there's no good way to surface suggestions through it. I'm on the fence about what to do here. WDYT about dropping suggestions and making rules either mandatory to follow or not?

I have:


tr0njavolta:
@negz - Dropping suggestions makes the most sense to me. The passive voice check is noisy and requires contributors wrap any "legit" passive voice instances in vale exceptions. Any reviewer would review for clarity and comprehension so I am okay with dropping altogether.

negz:
Thanks for the extra fixes @tr0njavolta!

PR #154, opened 2026-06-15
Fix deletion ordering for ProviderConfig consumers and ModelReplicas

### LINKED ISSUE #97: ServingStack Usages protect ProviderConfigs against only one Release/Object, not all
> [!NOTE]
> Updated 2026-06-15: This bug originated in `compose-kserve-backend`, which has since been removed along with KServe. The same `compose_usages` logic — and the same bug — now lives in `compose-serving-stack`. The EKS/KServe specifics below are historical; the gap is in the shared serving stack and is not cloud- or backend-specific.

### What happened?

`compose-serving-stack` composes two `Usage`s to keep its `ProviderConfig`s alive until the resources that use them are deleted, so Helm releases and provider-kubernetes Objects can uninstall from the workload cluster before losing the credentials they need. The intent, per the code comments, is "Helm ProviderConfig protected by all Releases" and "K8s ProviderConfig protected by all Objects":

[`compose-serving-stack/function/fn.py`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-serving-stack/function/fn.py) (the `compose_usages` method):

```python
# Helm ProviderConfig protected by all Releases.
usagev1beta1.Usage(spec=usagev1beta1.Spec(
    of=usagev1beta1.Of(kind="ProviderConfig", resourceRef=...pc),
    by=usagev1beta1.By(kind="Release", resourceSelector=usagev1beta1.ResourceSelector(matchControllerRef=True)),
    replayDeletion=True,
))
# K8s ProviderConfig protected by all Objects.
usagev1beta1.Usage(spec=usagev1beta1.Spec(
    of=usagev1beta1.Of(kind="ProviderConfig", resourceRef=...pc),
    by=usagev1beta1.By(kind="Object", resourceSelector=usagev1beta1.ResourceSelector(matchControllerRef=True)),
    replayDeletion=True,
))
```

A `Usage`'s `by` selector resolves to a **single** resource, not all matching ones. The Usage controller's selector resolver lists candidates, picks the first match, and stops. So `usage-helm-pc` protects the Helm `ProviderConfig` against exactly one `Release`, and `usage-k8s-pc` against exactly one `Object` - whichever the selector happened to resolve first. The other releases and objects are not holding the `ProviderConfig`.

The serving stack composes several `Release`s (cert-manager, envoy-gateway, ...) and several `Object`s (gateway-class, gateway, ...), so most of them are unprotected.

On teardown this lets the `ProviderConfig`s be deleted while releases and objects that still need them are mid-delete. Those then wedge. I originally hit it tearing down an EKS-backed cluster: the `GatewayClass` `Object` and the `gateway-helm` `Release` were left unable to finalize with `cannot get provider config: ProviderConfig.kubernetes.m.crossplane.io "...-cluster-..." not found` and the equivalent for the Helm `ProviderConfig`, because the single protecting Object/Release had already gone and let the `ProviderConfig` delete. I had to clear finalizers by hand to unblock the teardown.

This isn't cloud- or backend-specific - it's in the shared serving stack - so it affects GKE-backed and bring-your-own clusters too. It often won't surface because the one protected release/object happens to outlive the others, but it's a real ordering gap whenever it doesn't.

### How can we reproduce it?

1. Bring up any InferenceCluster whose serving stack installs more than one Helm `Release` and more than one provider-kubernetes `Object` (the default stack does).
2. Tear it down such that the `ProviderConfig`s are garbage collected around the same time as the releases/objects (e.g. observed during a foreground-cascade delete of the InferenceCluster).
3. Some of the stack's `Release`s/`Object`s wedge with `cannot get provider config: ... not found`, because only one of them was holding each `ProviderConfig` via its `Usage`.

Originally observed on the `ecks` branch with a KServe backend on EKS (Crossplane v2.3.1). The KServe backend has since been removed, but the same `compose_usages` logic carried over unchanged into `compose-serving-stack`, so the bug persists on `main` regardless of cloud or backend.

negz:

Fixes #97.

Two deletion-ordering gaps let Crossplane tear down a resource before its dependents finish cleaning up, wedging them with finalizers that need clearing by hand. Both come down to using Usages correctly across the namespaced/cluster-scoped boundary.

**ProviderConfig consumers (#97).** compose-serving-stack used two Usages with bare `matchControllerRef` `by` selectors to hold its ProviderConfigs alive until the Releases and Objects using them are gone. But a Usage's `by` resolves to a *single* resource, so each ProviderConfig was protected against only one consumer while the stack installs several. On teardown a ProviderConfig could be deleted out from under the rest, wedging them with `cannot get provider config: ... not found`. This adds a generic `compose-usages` function that runs as a late pipeline step: for every Release or Object referencing a ProviderConfig it stamps a unique label and composes one Usage per consumer, so each one holds the ProviderConfig until it's gone. It carries no serving-stack knowledge — any pipeline gets correct protection by running it last. compose-serving-stack keeps its separate Envoy Gateway ordering Usages.

**ModelReplicas vs their InferenceCluster.** Deleting an InferenceCluster out from under running ModelReplicas stranded them. A Usage can't express this directly: the cluster is cluster-scoped and replicas are namespaced, so a ClusterUsage's `by` can't reach a replica and a namespaced Usage can't name the cluster as its `of`. Instead, compose-inference-cluster protects the cluster with a reason-only ClusterUsage (no `by`), which blocks deletion until the Usage itself is gone. It requires the ModelReplicas labelled for this cluster across all namespaces and composes the ClusterUsage while any exist; when the last goes, it stops composing it and the cluster becomes deletable. The guard runs ahead of compose()'s early returns so a transient (e.g. classes briefly unresolved) doesn't drop it.

### Validation

End to end on an EKS-backed InferenceCluster (Qwen2.5-0.5B on a g6.xlarge L4):

- **#97:** one `usage-pc-*` Usage composed per Release/Object referencing each ProviderConfig; on teardown the ProviderConfigs outlived their consumers, then released.
- **ModelReplica guard:** deploying the model composed the ClusterUsage; `kubectl delete inferencecluster` was rejected by the `nousages.protection.crossplane.io` webhook; deleting the ModelDeployment dropped the guard and the cluster's pending delete replayed to completion.

Worth flagging: `replayDeletion` replays with the original attempt's propagation policy, which the webhook defaults to `Background`. A bare `kubectl delete inferencecluster` thus replays as a background delete — the teardown the docs warn against, which in my run leaked an ELB security group that blocked the VPC. `--cascade=foreground` replays cleanly. Not introduced here, but the guard makes foreground the path to take.

I have:

### DIFF
--- a/apis/servingstacks/composition.yaml
+++ b/apis/servingstacks/composition.yaml
@@ -11,3 +11,6 @@ spec:
   - functionRef:
       name: modelplane-modelplanecompose-serving-stack
     step: compose-serving-stack
+  - functionRef:
+      name: modelplane-modelplanecompose-usages
+    step: compose-usages
--- a/crossplane-project.yaml
+++ b/crossplane-project.yaml
@@ -60,6 +60,10 @@ spec:
     tarball:
       name: compose-model-service
       pathPrefix: _output/functions/compose-model-service
+  - source: Tarball
+    tarball:
+      name: compose-usages
+      pathPrefix: _output/functions/compose-usages
   dependencies:
   - type: crd
     git:
--- a/flake.nix
+++ b/flake.nix
@@ -67,6 +67,7 @@
         "compose-model-endpoint"
         "compose-model-replica"
         "compose-model-service"
+        "compose-usages"
       ];
 
       supportedSystems = [
--- a/functions/compose-inference-cluster/function/fn.py
+++ b/functions/compose-inference-cluster/function/fn.py
@@ -28,6 +28,7 @@ from models.io.crossplane.m.kubernetes.clusterproviderconfig import (
     v1alpha1 as k8scpcv1alpha1,
 )
 from models.io.crossplane.m.kubernetes.object import v1alpha1 as k8sobjv1alpha1
+from models.io.crossplane.protection.clusterusage import v1beta1 as clusterusagev1beta1
 from models.io.crossplane.protection.usage import v1beta1 as usagev1beta1
 from models.io.k8s.apimachinery.pkg.apis.meta import v1 as metav1
 
@@ -50,6 +51,15 @@ CONDITION_REASON_INVALID_NODE_POOL = "InvalidNodePool"
 
 # Composed resource key for the backend XR.
 BACKEND_RESOURCE_KEY = "serving-stack"
+
+# Composed resource key for the ClusterUsage that blocks the InferenceCluster's
+# deletion while ModelReplicas are scheduled to it.
+_REPLICA_GUARD_RESOURCE_KEY = "usage-replicas"
+
+# Label stamped on ModelReplicas by compose-model-deployment, carrying the name
+# of the InferenceCluster the replica is scheduled to. Kept in sync with that
+# function's _LABEL_CLUSTER.
+_LABEL_CLUSTER = "modelplane.ai/cluster"
 
 # Secret types that couple compose-gke-cluster (writer) to this function
 # (reader) and compose-serving-stack (reader).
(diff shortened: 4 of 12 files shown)


dennis-upbound (approved):
lgtm

PR #156, opened 2026-06-15
Latch ModelCache readiness so the hydration Job can be cleaned promptly

### LINKED ISSUE #140: ModelCache deletion hangs: completed hydration Job pod outlives the Job and pins the PVC
### What happened?

Deleting a ModelCache left its PVC stuck `Terminating` on the workload cluster. The PVC's `pvc-protection` finalizer wouldn't release because the cache's hydration Job pod - long since `Completed` - still existed and referenced the PVC. Deleting the completed pod by hand released the PVC.

The hydration Job already sets `ttlSecondsAfterFinished: 3600` ([`functions/compose-model-cache/function/fn.py:255`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-cache/function/fn.py#L255)), but that doesn't cover this case: the cache was deleted well within the hour, and when provider-kubernetes deleted the Job, the completed pod survived it. The `batch/v1` Job API's default deletion propagation is `orphanDependents` (kubectl overrides this to background cascade; generic API clients don't), so deleting the Job object orphans its pods.

Observed during E2E testing of #138 on EKS with the `modelplane-rwx-efs` StorageClass.

### How can we reproduce it?

1. Create a ModelCache and wait for hydration to complete (Job pod `Completed`).
2. Delete the ModelCache within an hour of hydration finishing.
3. On the workload cluster: the cache PVC stays `Terminating`; the hydration pod still exists and mounts it.
4. Delete the pod; the PVC deletes immediately.

Modelplane `v0.1.0-dev` (branch `topologicality`), EKS, Kubernetes 1.36, provider-kubernetes >=1.2.0, EFS CSI RWX storage.

dennis-upbound:
Fixes #140.

Deleting a ModelCache left its PVC stuck `Terminating`. The completed hydration Job pod still referenced the PVC, and `pvc-protection` won't release it while a pod references it. provider-kubernetes deletes the Job with the `batch/v1` default (orphan) propagation, so the pod is orphaned rather than cleaned and pins the PVC indefinitely.

The fix latches readiness and hands the Job's cleanup to its own TTL controller. A cluster's `Ready` phase is recorded in the XR status, so the cache stays `Ready` once hydration has succeeded even after the Job is gone. Once `Ready`, the Job is dropped from the composition — safe because hydration is idempotent (a marker file short-circuits a re-run, so a flap that re-adds the Job is a cheap skip, not a re-download). The Job carries every management policy except `Delete`, so dropping it **orphans** the external Job rather than deleting it: its own TTL controller then deletes it *with* its pods (a real cascade), instead of Crossplane orphaning the pod. The TTL drops to 180s so the PVC isn't pinned for long.

Worst case — deleting a cache within the TTL of hydration finishing — the PVC sits `Terminating` until the Job's TTL fires, then deletes cleanly: bounded by the TTL rather than the previous permanent hang.

A follow-up worth noting: the TTL has to sit above provider-kubernetes' observe interval so we catch the Job's success before it's cleaned. Enabling provider-kubernetes `--enable-watches` (a `DeploymentRuntimeConfig`, same shape we already use for provider-helm) + `watch: true` would make observation event-driven and let the TTL shrink further. Left out here on purpose: `watch` is alpha/not-for-production and `--enable-watches` is cross-cutting, so it's a deliberate call, not a one-function bugfix.

I have:

### DIFF
--- a/functions/compose-model-cache/function/fn.py
+++ b/functions/compose-model-cache/function/fn.py
@@ -50,6 +50,21 @@ REMOTE_NS = "default"
 HYDRATION_IMAGE = "python:3.11-slim"
 HYDRATION_MOUNT = "/mnt/artifact"
 
+# ttlSecondsAfterFinished governs how long the completed Job (and its
+# PVC-pinning pod) lingers before its TTL controller cascade-deletes both. Keep
+# it short so a just-deleted cache's PVC isn't pinned for long - but above
+# provider-kubernetes' observe interval, so the function still catches the Job's
+# success (to latch Ready and drop the Job) before the TTL fires.
+_JOB_TTL_SECONDS = 180
+
+# Every management policy except Delete. Once a cluster is Ready the Job is
+# dropped from the composition; without Delete, dropping orphans the external Job
+# instead of deleting it, so its own TTL controller cascade-cleans the completed
+# pod that pins the PVC - whereas Crossplane's delete would orphan that pod. (No
+# "all but Delete" shorthand exists, and deletionPolicy: Orphan is ignored once
+# managementPolicies is on.) Re-adding the Job after a flap is a cheap skip.
+_JOB_MANAGEMENT = ["Observe", "Create", "Update", "LateInitialize"]
+
 # Per-source default RWX storage class, mirroring the InferenceCluster XRD
 # defaults (GKE/Existing -> Filestore-backed modelplane-rwx; EKS -> EFS-backed
 # modelplane-rwx-efs). Used only when a cluster omits its cache block entirely:
@@ -142,9 +157,13 @@ class Composer:
         if not self.resolve_inputs():
             return
         matched = self.match_clusters()
-        for cluster in matched:
-            self.compose_cluster_resources(cluster)
+        # Derive each cluster's phase first (from observed state), then compose:
+        # a hydrated cluster's Job is composed Observe-only so Crossplane doesn't
+        # recreate it after the TTL controller cleans it.
         per_cluster_phase = [(c.metadata.name, self.derive_cluster_phase(c.metadata.name)) for c in matched]
+        phase_by_name = dict(per_cluster_phase)
+        for cluster in matched:
+            self.compose_cluster_resources(cluster, phase_by_name[cluster.metadata.name])
         self.mark_ready_resources(per_cluster_phase)
         self.write_status(matched, per_cluster_phase)
         self.derive_conditions(matched, per_cluster_phase)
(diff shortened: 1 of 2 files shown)


functions/compose-model-cache/function/fn.py
```diff
@@ -50,6 +50,15 @@
 HYDRATION_IMAGE = "python:3.11-slim"
 HYDRATION_MOUNT = "/mnt/artifact"
 
+# The completed hydration pod pins the cache PVC until the pod itself is deleted.
+# provider-kubernetes deletes the Job with the batch/v1 default (orphan)
+# propagation, which leaves the pod behind, so only the Job's own TTL controller
+# cascade-cleans it. Keep the TTL short so a deleted cache's PVC isn't pinned for
+# long - but above provider-kubernetes' observe interval, so the function still
+# catches the Job's success before the Job is cleaned. Once a cluster is hydrated
+# the Job flips to Observe (compose_cluster_resources) so it isn't recreated.
```
negz: Making sure I understand: can we not just stop composing (i.e. proactively delete) the Job once we know hydration has succeeded? I think you're saying this won't work because Jobs orphan their Pods unless you delete with a specific policy, which Crossplane doesn't do?

dennis-upbound: right. unless there's something in Crossplane I'm missing?

negz: No, I think what you've done here is the best option.

functions/compose-model-cache/function/fn.py
```diff
@@ -142,9 +151,13 @@ def compose(self):
         if not self.resolve_inputs():
             return
         matched = self.match_clusters()
-        for cluster in matched:
-            self.compose_cluster_resources(cluster)
+        # Derive each cluster's phase first (from observed state), then compose:
+        # a hydrated cluster's Job is composed Observe-only so Crossplane doesn't
+        # recreate it after the TTL controller cleans it.
```
negz, line 162: Will this leave the observed resource in an error state?

I'm wondering if there's some way we can safely stop composing the Job at all once it's no longer needed. I imagine that'd be racy?

dennis-upbound: Reworked to address this. Instead of the Observe-flip (which would have left exactly that erroring ghost MR), we now **drop** the Job once `Ready` — which is the "safely stop composing it once it's no longer needed" you were after. The enabler is the Job's management policy excluding `Delete`: dropping then *orphans* the external Job instead of deleting it, so its own TTL controller cascade-cleans the completed pod, rather than Crossplane orphaning it. No ghost MR. Dropping is safe because hydration is idempotent (a flap re-adds a cheap marker-skip, not a re-download). It relies only on no-Delete-orphans-on-removal + the Job TTL cascade, so there's nothing here that needs an e2e to de-risk.

negz: Ah, nice fix!

negz (approved):

PR #159, opened 2026-06-16
Fix the GAIE inferencepool install in the serving stack

### LINKED ISSUE #157: Serving stack pulls the inferencepool chart from ghcr.io, which denies anonymous pulls
### What happened?

`compose-serving-stack` installs the Gateway API Inference Extension (GAIE) `inferencepool` Helm chart as part of the serving stack. The Release never installs: provider-helm can't pull the chart because the configured registry denies anonymous pulls.

```
create failed: failed to install release: failed to pull chart: failed to perform "FetchReference" on source: GET "https://ghcr.io/v2/kubernetes-sigs/gateway-api-inference-extension/charts/inferencepool/manifests/v1.0.1": GET "https://ghcr.io/token?scope=repository%3Akubernetes-sigs%2Fgateway-api-inference-extension%2Fcharts%2Finferencepool%3Apull&service=ghcr.io": response status code 403: denied: requested access to the resource is denied
```

The chart is pulled from `oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts`, set in [`compose-serving-stack/function/fn.py:60`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-serving-stack/function/fn.py#L60):

```python
_GAIE_CHART = "inferencepool"
_GAIE_REPO = "oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts"
_GAIE_VERSION = "v1.0.1"
```

That ghcr path requires authentication — its token endpoint returns `403 denied` (and the tags endpoint returns `401 authentication required`) for anonymous clients. The serving stack composes the Release with no `pullSecretRef`, so the pull is anonymous and fails.

GAIE publishes the chart for public consumption on `registry.k8s.io`, not ghcr. The upstream [getting-started guide](https://gateway-api-inference-extension.sigs.k8s.io/guides/) installs it from `oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool`, which pulls anonymously:

```
$ helm pull oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool --version v1.0.1
Pulled: registry.k8s.io/gateway-api-inference-extension/charts/inferencepool:v1.0.1
Digest: sha256:301b913dbff1d75017db0962b621e6780777dcb658475df60d1c6b5b84ee1635
```

The `inferencepool` Release is always composed, so its failure pins `BackendReady=False` on the ServingStack and `Ready=False` on the InferenceCluster indefinitely. The scheduler in `compose-model-deployment` gates on `InferenceCluster Ready=True` ([`scheduling.py:289`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-deployment/function/scheduling.py#L289)), so no model can be deployed to an affected cluster. Every other release and object in the stack reaches Ready; only this one fails.

Introduced by #142 (commit 891d51e5), which added the GAIE install. Switching `_GAIE_REPO` to `oci://registry.k8s.io/gateway-api-inference-extension/charts` looks like the fix; the version and chart name are otherwise correct.

### How can we reproduce it?

1. Bring up any InferenceCluster (EKS or GKE). The serving stack installs the `inferencepool` Release.
2. `kubectl get release.helm.m.crossplane.io -A` — the `inferencepool` release stays `SYNCED=False` with the 403 above.
3. The InferenceCluster never reaches `Ready=True`, and `kubectl apply` of a ModelDeployment never schedules.

Direct confirmation of the registry difference:

```bash
# Configured (ghcr): denied
helm pull oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts/inferencepool --version v1.0.1
# Error: ... response status code 403: denied

# Upstream public (registry.k8s.io): works
helm pull oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool --version v1.0.1
# Pulled: registry.k8s.io/gateway-api-inference-extension/charts/inferencepool:v1.0.1
```

Modelplane version: `main` (also on a branch built from `main`), commit 891d51e5 / `v0.1.0-dev` build.

* Crossplane: v2.3.2
* Inference backend: serving stack (GAIE inferencepool chart)
* Workload cluster: EKS (`g6.xlarge`, 1x L4, us-west-2) — not cloud-specific; same code path on GKE
* Kubernetes: control plane kind v1.34.0; EKS cluster default
* provider-helm v1.2.5, provider-kubernetes v1.2.5

### LINKED ISSUE #158: Serving stack can't install the inferencepool chart for GAIE CRDs: chart ships no CRDs and requires modelServers.matchLabels
### What happened?

`compose-serving-stack` installs the Gateway API Inference Extension (GAIE) `inferencepool` Helm chart to obtain the InferencePool CRD. The Release fails to install because the chart renders a running InferencePool instance, not CRDs, and requires values the serving stack doesn't pass:

```
create failed: failed to install release: execution error at (inferencepool/templates/inferencepool.yaml:22:3): .Values.inferencePool.modelServers.matchLabels is required
```

The intent, per [`compose-serving-stack/function/fn.py:496`](https://github.com/modelplaneai/modelplane/blob/main/functions/compose-serving-stack/function/fn.py#L496) (`compose_gaie_crds`), is to install the chart purely to get the InferencePool CRD onto the workload cluster, "equivalent to applying that release's manifests.yaml". It composes the Release with no values:

```python
resource.update(
    self.rsp.desired.resources["gaie-crds"],
    _helm_release(
        chart=_GAIE_CHART,        # "inferencepool"
        repo=_GAIE_REPO,          # oci://registry.k8s.io/gateway-api-inference-extension/charts
        version=_GAIE_VERSION,    # v1.0.1
        namespace=_GAIE_NAMESPACE,
        provider_config=_pc_name(self.xr),
    ),
)
```

Two things are wrong with using this chart as a CRD source:

1. The `inferencepool` chart ships **no CRDs**. It has no `crds/` directory; its `templates/` render a full running InferencePool deployment - the InferencePool resource itself, the endpoint picker (EPP) Deployment, ServiceAccount, RBAC, Service, and ServiceMonitor. Installing it does not install the `inference.networking.k8s.io` CRDs.

2. Because the templates render an actual InferencePool, the chart **requires** `inferencePool.modelServers.matchLabels` (templated at `inferencepool/templates/inferencepool.yaml:12`), which a CRD-only install has no sensible value for. With no values set, the render fails outright.

Upstream installs the GAIE CRDs separately from the release's `manifests.yaml`, and installs the `inferencepool` chart once per InferencePool with `--set inferencePool.modelServers.matchLabels.app=<pool>` (see the [getting-started guide](https://gateway-api-inference-extension.sigs.k8s.io/guides/)). The chart is a per-pool runtime install, not a CRD bundle.

The `inferencepool` Release is always composed, so its failure pins `BackendReady=False` on the ServingStack and `Ready=False` on the InferenceCluster, which blocks model scheduling. This surfaces only after the registry is corrected (#157); with the wrong registry the pull fails first.

Fixing this likely means dropping the chart-as-CRD-source approach and installing the GAIE CRDs from the upstream `manifests.yaml` instead (the codebase has no remote-manifest pattern for provider-kubernetes Objects today, which the code comment notes). If a running InferencePool is actually wanted here, it needs the required per-pool values - but that's a different model from "install the CRDs once per cluster".

### How can we reproduce it?

1. Bring up any InferenceCluster (EKS or GKE) with the `inferencepool` chart pointed at a reachable registry (i.e. with #157 applied).
2. `kubectl get release.helm.m.crossplane.io -A` - the `gaie-crds` / inferencepool Release stays `SYNCED=False` with the `modelServers.matchLabels is required` error above.
3. The InferenceCluster never reaches `Ready=True`; no model can be scheduled.

Confirm the chart ships no CRDs and what it renders:

```bash
helm pull oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool --version v1.0.1 --untar --untardir /tmp/gaie
ls /tmp/gaie/inferencepool/crds/        # no such directory
ls /tmp/gaie/inferencepool/templates/   # inferencepool.yaml, epp-deployment.yaml, rbac.yaml, ...
```

Modelplane version: `main` (and a branch built from `main`), GAIE introduced by #142 (commit 891d51e5), with the #157 registry fix applied.

* Crossplane: v2.3.2
* Inference backend: serving stack (GAIE inferencepool chart v1.0.1)
* Workload cluster: EKS (`g6.xlarge`, 1x L4, us-west-2) - not cloud-specific; same code path on GKE
* Kubernetes: control plane kind v1.34.0
* provider-helm v1.2.5, provider-kubernetes v1.2.5

negz:
Fixes #157.
Fixes #158.

The serving stack's Gateway API Inference Extension (GAIE) install, added in #142, can't complete, so a freshly provisioned InferenceCluster never reaches `Ready` and no model can be scheduled onto it. Three problems, found while bringing up an EKS-backed cluster end to end.

First, the `inferencepool` chart was pulled from `oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts`, which denies anonymous pulls — its token endpoint returns 403. provider-helm can't fetch the chart, so the Release never installs. GAIE publishes the chart for public consumption on `registry.k8s.io`, which serves it anonymously, so this points the repo there.

Second, with the registry corrected, the install still fails: the `inferencepool` chart ships no CRDs. It renders a running InferencePool instance plus its endpoint picker and requires `inferencePool.modelServers.matchLabels`, which the serving stack doesn't set, so the render errors out. The CRDs the serving stack actually wants live in the upstream release's `manifests.yaml`. This vendors them into the function and applies them as provider-kubernetes Objects on the remote cluster, the same way `compose-inference-gateway` installs the Gateway API CRDs that the Traefik chart likewise doesn't ship. The upstream manifests are a live-cluster export carrying a top-level `status` and `metadata.creationTimestamp`; those are stripped from the vendored copy (with a header note for re-vendoring) so the Object doesn't re-apply every reconcile and trip the composite's watch circuit breaker.

Third, the `ai-gateway-crds` and `ai-gateway` releases were composed but never marked ready — they weren't in `mark_readiness`'s list. So even with every release and object healthy on the cluster, the ServingStack's readiness aggregation waited on them forever and the InferenceCluster never reached `Ready`. This adds them to the list.

The first two unmask each other in sequence (the registry fix reveals the chart-content problem); the third was hidden behind both. Each maps to its own commit.

Validated end to end on an EKS-backed InferenceCluster: with all three fixes the GAIE CRDs install and settle, the serving stack converges, and the InferenceCluster reaches `Ready=True`.

I have:

### DIFF
--- a/functions/compose-serving-stack/function/fn.py
+++ b/functions/compose-serving-stack/function/fn.py
@@ -57,7 +57,7 @@ _AI_GATEWAY_CONTROLLER_PORT = 1063
 # Gateway API Inference Extension (GAIE) constants. Provides the InferencePool
 # CRD that disaggregated replicas front their decode endpoints with.
 _GAIE_CHART = "inferencepool"
-_GAIE_REPO = "oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts"
+_GAIE_REPO = "oci://registry.k8s.io/gateway-api-inference-extension/charts"
 _GAIE_VERSION = "v1.0.1"
 _GAIE_NAMESPACE = "gateway-api-inference-extension"
 
--- a/functions/compose-serving-stack/function/fn.py
+++ b/functions/compose-serving-stack/function/fn.py
@@ -13,7 +13,10 @@ Usage resources protect ProviderConfigs from premature deletion during
 teardown, ensuring Helm releases can uninstall before losing connectivity.
 """
 
+import pathlib
+
 import grpc
+import yaml
 from crossplane.function import logging, resource, response
 from crossplane.function.proto.v1 import run_function_pb2 as fnv1
 from crossplane.function.proto.v1 import run_function_pb2_grpc as grpcv1
@@ -54,12 +57,21 @@ _AI_GATEWAY_VERSION = "v0.7.0"
 _AI_GATEWAY_CONTROLLER_FQDN = f"ai-gateway-controller.{_AI_GATEWAY_NAMESPACE}.svc.cluster.local"
 _AI_GATEWAY_CONTROLLER_PORT = 1063
 
-# Gateway API Inference Extension (GAIE) constants. Provides the InferencePool
-# CRD that disaggregated replicas front their decode endpoints with.
-_GAIE_CHART = "inferencepool"
-_GAIE_REPO = "oci://registry.k8s.io/gateway-api-inference-extension/charts"
-_GAIE_VERSION = "v1.0.1"
-_GAIE_NAMESPACE = "gateway-api-inference-extension"
+
+# Gateway API Inference Extension (GAIE) CRDs, providing the InferencePool that
+# disaggregated replicas front their decode endpoints with. Vendored from the
+# upstream release's manifests.yaml.
+_HERE = pathlib.Path(__file__).parent
+_GAIE_CRDS = [
+    doc
+    for doc in yaml.safe_load_all((_HERE / "gaie_crds.yaml").read_text())
+    if doc and doc.get("kind") == "CustomResourceDefinition"
+]
+
+
+def _gaie_crd_key(doc: dict) -> str:
+    """Stable composed-resource key for a GAIE CRD."""
+    return f"gaie-crd-{doc['metadata']['name']}"
 
 
 def _helm_release(
@@ -539,29 +551,21 @@ class Composer:
         )
 
     def compose_gaie_crds(self):
-        """Compose the Gateway API Inference Extension (GAIE) CRDs. Gated on the
-        same ProviderConfigs as Envoy Gateway.
-
-        Installs the InferencePool CRD that disaggregated replicas front their
-        decode endpoints with. The codebase has no remote-manifest pattern for
-        provider-kubernetes Objects, so the CRDs come from the upstream Helm
-        chart published by kubernetes-sigs - equivalent to applying that
-        release's manifests.yaml.
+        """Compose the Gateway API Inference Extension (GAIE) CRDs as
+        provider-kubernetes Objects on the remote cluster. Gated on the same
+        ProviderConfigs as Envoy Gateway.
         """
         pc_observed = self.provider_configs_observed()
-        if not (pc_observed or "gaie-crds" in self.req.observed.resources):
-            return
-
-        resource.update(
-            self.rsp.desired.resources["gaie-crds"],
-            _helm_release(
-                chart=_GAIE_CHART,
-                repo=_GAIE_REPO,
-                version=_GAIE_VERSION,
-                namespace=_GAIE_NAMESPACE,
-                provider_config=_pc_name(self.xr),
-            ),
-        )
+        for doc in _GAIE_CRDS:
+            key = _gaie_crd_key(doc)
+            if not (pc_observed or key in self.req.observed.resources):
+                continue
+            resource.update(
+                self.rsp.desired.resources[key],
+                _k8s_object(_pc_name(self.xr), doc),
+            )
+            if resource.get_condition(self.req.observed.resources.get(key), "Ready").status == "True":
+                self.rsp.desired.resources[key].ready = fnv1.READY_TRUE
 
     def compose_prometheus(self):
         """Compose the kube-prometheus-stack. Gated on ProviderConfigs being
--- a/functions/compose-serving-stack/function/fn.py
+++ b/functions/compose-serving-stack/function/fn.py
@@ -759,6 +759,8 @@ class Composer:
         condition_ready = [
             "cert-manager",
             "envoy-gateway",
+            "ai-gateway-crds",
+            "ai-gateway",
             "prometheus",
             "leader-worker-set",
             "node-feature-discovery",
--- a/functions/compose-serving-stack/function/gaie_crds.yaml
+++ b/functions/compose-serving-stack/function/gaie_crds.yaml
@@ -1,0 +1,1007 @@
+# Gateway API Inference Extension (GAIE) CRDs, vendored from the v1.0.1 release:
+# https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.0.1/manifests.yaml
+#
+# When re-vendoring (bump the release in the URL above), strip each CRD's
+# top-level status and metadata.creationTimestamp. The upstream manifests carry
+# these server-owned fields; left in, the provider-kubernetes Object that applies
+# a CRD re-applies every reconcile (observed never matches desired), flooding
+# watch events and tripping the composite's watch circuit breaker.
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    inference.networking.k8s.io/bundle-version: v1.0.1
+  name: inferenceobjectives.inference.networking.x-k8s.io
+spec:
+  group: inference.networking.x-k8s.io
+  names:
+    kind: InferenceObjective
+    listKind: InferenceObjectiveList
+    plural: inferenceobjectives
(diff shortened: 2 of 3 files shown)


dennis-upbound (approved):
thanks for doing this. hope this works

functions/compose-serving-stack/function/fn.py
```diff
@@ -54,12 +57,21 @@
...
-# Gateway API Inference Extension (GAIE) constants. Provides the InferencePool
-# CRD that disaggregated replicas front their decode endpoints with.
-_GAIE_CHART = "inferencepool"
-_GAIE_REPO = "oci://ghcr.io/kubernetes-sigs/gateway-api-inference-extension/charts"
-_GAIE_VERSION = "v1.0.1"
-_GAIE_NAMESPACE = "gateway-api-inference-extension"
+# Gateway API Inference Extension (GAIE) CRDs, providing the InferencePool that
+# disaggregated replicas front their decode endpoints with. Vendored from the
+# upstream release's manifests.yaml; the inferencepool Helm chart renders a
+# running InferencePool instance, not the CRDs, so we apply them directly.
```
negz: Here and below we should drop this info about what the Helm chart does. I don't think there's any reason this is interesting to the user - it's the agent narrating a change.

negz: Done — dropped the chart narration in the `_GAIE_CRDS` comment and the `compose_gaie_crds` docstring. Kept the why on `_sanitize_crd` (the watch-circuit-breaker churn), since that's non-obvious.

negz: Also took your follow-up: dropped the runtime _sanitize_crd guard, strip the server-owned fields in the vendored gaie_crds.yaml only, and left a note in its header telling maintainers to strip top-level status and metadata.creationTimestamp when re-vendoring (with the watch-circuit-breaker reason). Removed the chart narration from the YAML header too.

functions/compose-serving-stack/function/gaie_crds.yaml
```diff
@@ -0,0 +1,1007 @@
+# Gateway API Inference Extension (GAIE) CRDs, vendored from upstream:
+# https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.0.1/manifests.yaml
+#
+# When re-vendoring, strip each CRD's top-level status and
+# metadata.creationTimestamp. The upstream manifests carry these server-owned
+# fields; left in, the provider-kubernetes Object that applies a CRD re-applies
+# every reconcile (observed never matches desired), flooding watch events and
+# tripping the composite's watch circuit breaker. Keep in sync with _GAIE_VERSION.
```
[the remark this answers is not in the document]
negz: Fixed — the header no longer references `_GAIE_VERSION` (which was removed with the Helm-release approach). It now points maintainers at the release URL and says to bump it when re-vendoring.

functions/compose-serving-stack/function/fn.py
```diff
@@ -54,12 +57,21 @@
...
+
+# Gateway API Inference Extension (GAIE) CRDs, providing the InferencePool that
+# disaggregated replicas front their decode endpoints with. Vendored from the
+# upstream release's manifests.yaml.
+_HERE = pathlib.Path(__file__).parent
+_GAIE_CRDS = [
+    doc
+    for doc in yaml.safe_load_all((_HERE / "gaie_crds.yaml").read_text())
+    if doc and doc.get("kind") == "CustomResourceDefinition"
+]
```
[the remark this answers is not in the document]
negz: We deliberately went with strip-only and no load-time guard, per review discussion above: the vendored `gaie_crds.yaml` is committed already stripped of `status`/`creationTimestamp`, with a header note telling maintainers to strip them when re-vendoring. The PR description was stale (it still mentioned the guard); I've updated it to match.

PR #164, opened 2026-06-16
Auto-provision EFS RWX storage for ModelCache on EKS

dennis-upbound:
Fixes #114.

On GKE, Modelplane auto-provisions the `modelplane-rwx` Filestore StorageClass, so a `ModelCache` PVC binds with no admin setup. On EKS, RWX storage was bring-your-own: the admin had to install the EFS CSI driver, create an EFS filesystem with mount targets and a security group, wire IAM, and create the `modelplane-rwx-efs` StorageClass. This brings EKS to parity.

`compose-eks-cluster` now provisions the EFS substrate on every EKS cluster (like GKE enables the Filestore API): an Elastic-throughput filesystem, a mount target per node subnet, an NFS security group, and the EFS CSI driver. The driver's IAM role is bound through **EKS Pod Identity** — the `eks-pod-identity-agent` addon plus a `PodIdentityAssociation` — rather than IRSA, so there's no OIDC provider, thumbprint, or trust-policy plumbing (and it leaves a reusable Pod Identity foundation for future addons). The filesystem id is surfaced on `EKSCluster.status.efsFileSystemId`, the way `GKECluster` surfaces its network name.

`compose-inference-cluster` reads that id and composes the `modelplane-rwx-efs` StorageClass on the workload cluster, gated on the id being present (it can't pin to a bare name) — mirroring the existing GKE Filestore StorageClass. An admin-provided `storageClassName` is left unmanaged.

Provisioning is unconditional per EKS cluster: an empty EFS filesystem costs ~nothing, and gating on "a cache exists" would chicken-and-egg with the StorageClass. Throughput is Elastic (hardcoded) so a cold cache being hydrated then read by many pods never throttles. FSx for Lustre (the issue's future note) and configurable throughput are out of scope.

Design notes: `design/eks-efs-modelcache.md`.

Validated on a live EKS cluster in `us-west-2`: `compose-eks-cluster` provisions the full EFS substrate — the filesystem reaches `available`, with mount targets, the NFS security group, the EFS-CSI IAM role + policy attachment, the `eks-pod-identity-agent` and `aws-efs-csi-driver` addons, and the `PodIdentityAssociation` — and `efsFileSystemId` surfaces on `EKSCluster.status`. `compose-inference-cluster`'s StorageClass gate reads that id and is satisfied; the StorageClass itself composes once the cluster's addons finish bootstrapping, and its goldens cover the composed `Object`.

I have:

[...]

PR #176, opened 2026-06-17
Move ModelCache storage provisioning into the cluster XRs

negz:

Follows on from #164, which brought EKS to parity with GKE by auto-provisioning the EFS substrate for ModelCache RWX storage.

ModelCache needs a ReadWriteMany StorageClass on each workload cluster. The cloud-specific substrate behind that class — EFS on EKS, Filestore on GKE — is provisioned by `compose-eks-cluster` and `compose-gke-cluster`, but the StorageClass that fronts it was composed a layer up, in `compose-inference-cluster`. To do that, the cluster XRs surfaced their substrate identifiers (`EKSCluster.status.efsFileSystemId`, `GKECluster.status.network.name`) so the InferenceCluster could read them back and pin a provider-kubernetes `Object` to them.

This PR moves the StorageClass down to live alongside the substrate it depends on. `compose-eks-cluster` and `compose-gke-cluster` now compose the StorageClass `Object` directly, through the provider-kubernetes `ProviderConfig` they already maintain for their own cluster, and report the effective class name on a new `status.cache.storageClassName`. The InferenceCluster relays that name up to its own `status.cache`, and `compose-model-cache` reads it from there to target the cache PVC. The substrate identifiers no longer have a reader, so they're dropped from the cluster XRs' status.

As a result the InferenceCluster composes only the cluster XR, a ServingStack, a ClusterProviderConfig, and the Usages that guard deletion — the storage resources move out entirely.

On a Modelplane-provisioned cluster the StorageClass can't pre-exist: the cluster isn't there until `compose-eks-cluster` or `compose-gke-cluster` creates it. So a provisioned cluster always gets the Modelplane-managed class (`modelplane-rwx` on GKE, `modelplane-rwx-efs` on EKS) and reports it on `status.cache.storageClassName`. The bring-your-own knob, `cache.storageClassName`, now applies only to the `Existing` source, where the admin owns the cluster and can supply a class.

I have:


dennis-upbound:
ah yes this is a lot cleaner! thanks!

negz:
Just validated this E2E on an EKS cluster - ModelCache still works as before.

Issue #187, opened 2026-06-18
Use the modelplane Cachix cache to speed up CI builds


negz:
### What problem are you facing?

CI builds everything from source with Nix on every run. The `check` and `push-package` jobs in `.github/workflows/ci.yml` run `nix flake check` and `nix build`/`nix run` without any binary cache, so each run rebuilds derivations that haven't changed. This makes CI slower than it needs to be.

Crossplane solves this with a [Cachix](https://app.cachix.org/cache/crossplane) binary cache wired into its CI (see Crossplane's `.github/workflows/ci.yml`, which runs `cachix/cachix-action` after `install-nix-action`). We've reserved the `modelplane` cache at https://app.cachix.org/cache/modelplane to do the same.

This is blocked on taking the repo public. We want to keep it as secret as possible until then, and we don't want a private Cachix cache because public caches are free. So this is a fast-follow for when the repo goes public.

### How could Modelplane help solve your problem?

Wire the `modelplane` Cachix cache into CI, mirroring how Crossplane does it:

1. Add a `CACHIX_AUTH_TOKEN` secret to the repo (a write token from the `modelplane` Cachix cache).
2. After the `Install Nix` step in each Nix job in `.github/workflows/ci.yml`, add a Cachix step:

```yaml
- name: Setup Cachix
  uses: cachix/cachix-action@v17
  with:
    name: modelplane
    authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
```

This pushes build outputs to the cache on `main`/`workflow_dispatch` runs and pulls from it on PRs, so unchanged derivations are fetched rather than rebuilt.

**Blocked on:** taking the repo public. Don't enable this until then — a public Cachix cache would expose the build graph while the repo is still meant to be secret.

negz:
We may not need this given https://github.com/modelplaneai/modelplane/pull/190 enables the Determinate "Magic Nix Cache", which caches builds to GHA Cache. Last time I looked this approach was broken, but looks like they fixed it.

negz:
Magic Nix Cache seems to be working well enough.

PR #188, opened 2026-06-18
Free disk space before building and pushing the package

negz:

The `push-package` job fails with `No space left on device` while building the package — see [run 27733270322](https://github.com/modelplaneai/modelplane/actions/runs/27733270322). The `ubuntu-24.04` runner ships with several preinstalled toolchains we don't use, leaving too little free disk for the Nix build.

This adds a `Cleanup Disk` step (using [`jlumbroso/free-disk-space`](https://github.com/jlumbroso/free-disk-space), the same action crossplane/crossplane uses) before checkout. It removes the Android, .NET, and Haskell runtimes and the runner tool cache, reclaiming over 20GB. The job builds an OCI package with Nix and doesn't use Docker images, swap, or apt packages, so those slower options stay disabled.

I have:

- [ ] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~ This only changes a GitHub Actions workflow.
- [ ] ~Added or updated tests covering any composition function changes.~ No composition function changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,19 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
+      # Building and pushing the package needs more disk than the runner has
+      # free by default, so reclaim space we don't use.
+      - name: Cleanup Disk
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+        with:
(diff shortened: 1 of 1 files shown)


negz:
https://github.com/modelplaneai/modelplane/issues/187 may also help in future by avoiding some build dependencies. For example today we're compiling the crossplane CLI on each run.

Issue #186, opened 2026-06-18
ModelCache footprint can silently diverge from where replicas are scheduled


pluna:
### What problem are you facing?

A `ModelCache` stages a model's weights onto the `InferenceCluster`s its `clusterSelector` matches (or every cluster, when the selector is omitted). A `ModelDeployment` that wants those weights points at the cache by name
(`modelCacheRef`), and the fleet scheduler then places each replica on whichever matching cluster has capacity.

Nothing keeps the cache's footprint aligned with where replicas actually land. The cache's `clusterSelector` and the deployment's placement are decided independently, and the scheduler picks a replica's cluster based on capacity, which I cannot fully predict. If a replica lands on a cluster the cache did not stage to (my preferred clusters were full and the scheduler spilled elsewhere, or I narrowed the cache's selector to save storage), the cache's PVC simply is not there. The pod cannot mount it and the replica is stuck. This only shows up at runtime as a volume-mount failure on the pod, not as anything I can see on the `ModelDeployment` or `ModelCache` at apply time.

Today the only safe option is to stage the cache to every cluster the deployment could ever land on. For a large model that means paying storage on clusters I may never deploy to. Narrowing the cache to control that cost reintroduces the
misalignment risk. I am forced to choose between wasted storage and a silent runtime failure.

"Stage everywhere to be safe" also has a security cost. When a model is gated or private I attach download credentials to the cache, and hydration runs on every matched cluster. A broad or omitted selector means that credential is used, and the private weights are written, on clusters that run none of my deployments. The token and the weights end up in each of those
workload clusters' namespace, on infrastructure other teams share, outside my own namespace. So the workaround quietly widens who can reach my credentials and my model.

There is also no lifecycle coupling. A `ModelCache` is created and deleted by hand, independent of the deployments that use it. Nothing warns me that a cache is still referenced before I delete it (deleting one out from under a live
deployment removes its PVC), and nothing reclaims a cache once no deployment uses it. The staged weights sit on storage on every matched cluster until someone remembers to clean them up.

### How could Modelplane help solve your problem?

Caching should be automatic and tied to the model, not a resource I create and wire up by name. The deployment already declares the model it runs, so I should not also have to author a `ModelCache` and keep its footprint and lifecycle in
sync with the deployment by hand. Functionally:

- I declare the model on my deployment and stop managing caching. Modelplane  stages the weights on whatever cluster the scheduler places a replica on, and  only there. A cluster that has not served this model before pays the download
[...]

PR #189, opened 2026-06-18
Constrain ModelDeployment placement to its ModelCache footprint

negz:

Relates to #186.

A `ModelCache` stages a model's weights onto the `InferenceCluster`s its `clusterSelector` matches. A `ModelDeployment` that references the cache via `spec.modelCacheRef` has its placement decided independently — the scheduler filters candidate clusters only by the deployment's own `clusterSelector` and ignores the cache entirely. A replica can land on a cluster the cache never staged to, where its PVC doesn't exist, and the pod gets stuck on a volume mount that never resolves. Nothing surfaces this at apply time on the `ModelDeployment` or `ModelCache`; it shows up only at runtime.

This makes `compose-model-deployment` read the referenced `ModelCache` and intersect its `clusterSelector` with the deployment's own when filtering the candidate clusters for *new* replicas. Both are `matchLabels` selectors, so the candidate set is their AND, and a new replica can't schedule to a cluster outside the cache's footprint.

The constraint applies to new placement only. The scheduler already retains existing replicas on their pinned cluster; it gains a `fill` flag that gates only the placement of *new* replicas, leaving retain untouched. Losing the cache therefore can't evict or re-place a running replica — dropping a narrowing input only ever broadens the candidate set. Since the cache is read only when weights are first staged, deleting it out from under a live deployment is a no-op for the replicas already running: they're retained, and `fill` is held off only for new ones.

When the referenced cache can't be read, the footprint is unknown, so the function holds off placing new replicas while still retaining the ones it has. It reports this on a `ModelCacheResolved` condition, kept separate from `ReplicasScheduled` so a deployment can be scheduled (retaining replicas) and waiting on its cache at the same time. A cache Crossplane hasn't fetched yet is transient and reported with a condition alone; a cache that resolved to nothing — deleted, or a typo'd ref — also emits a warning event. `get_required_resource` returns `None` for both of those cases, so a small `Resolution` helper classifies a required resource as unresolved / absent / present; I've opened crossplane/function-sdk-python#217 to upstream it, with a `TODO` to drop the local copy once it lands.

#186 raises a broader question: whether `ModelCache` should remain an explicit resource at all, or whether caching should be derived from placement so the footprint can't be authored out of sync in the first place. That's a conceptual and architectural change, and not one to rush into v0.1. This PR doesn't foreclose it — it's the small, safe change that removes the silent-failure footgun now and buys us time to align on how `ModelCache` should work without a runtime sharp edge in the meantime.

I have:

### DIFF
--- a/docs/content/models/model-cache.md
+++ b/docs/content/models/model-cache.md
@@ -29,7 +29,11 @@ Each cache has:
   (`Dragonfly` for P2P distribution, `OCI` for NIM-style bundled artifacts).
 - An optional **clusterSelector** to scope replication. Omitting
   `spec.clusterSelector` stages the cache on every matched cluster; setting
-  `matchLabels` restricts it to clusters carrying those labels.
+  `matchLabels` restricts it to clusters carrying those labels. A
+  `ModelDeployment` that references the cache places *new* replicas only onto
+  clusters within this footprint, so narrowing the selector also narrows where
+  replicas can land - a replica never schedules to a cluster the cache didn't
+  stage to. Replicas already running are left where they are.
 
 The cache mounts at `/mnt/models` on every consuming pod; engine container args
 should reference this path (`--model=/mnt/models` for vLLM).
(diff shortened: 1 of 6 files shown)


dennis-upbound (approved):
this makes sense

PR #200, opened 2026-06-18
docs: Restructure Get Started

tr0njavolta:

This PR breaks the get started guide into 2 parts - one for a simple single cluster/model deployment to show the workflow and then moves into scaling and selecting hardware from the CEL selector rather than deploying a fleet right off the bat.

This is based on discussion with @dennis-upbound on probable common scenarios. Initial guide mimics a test/dev deployment to show a working model and then moves to fleet level deployments with larger hardware based on the growing resource needs.

I have 2 todo comments still here to explain the Modelplane Crossplane configuration deployment (based on positioning) and explaining the CEL selector in the Model Deployment back to the inference class.

Fixes #

I have: 


dennis-upbound (approved):
Looks good! I'll take another pass later after I verify the demo also

PR #203, opened 2026-06-18
Make the NVIDIA DRA driver work on GKE

### LINKED ISSUE #202: Serving stack DRA driver never schedules on GKE Standard (system-critical priority blocked by ResourceQuota)
## Problem

On GKE Standard, the serving stack's NVIDIA DRA driver daemonset never schedules, so no `ResourceSlice` is ever published, every `claim: DRA` GPU request fails to allocate, and **no model can serve** — `ModelReplica` pods sit `Pending` indefinitely. The `ServingStack` still reports `Ready=True`, which hides the failure.

This is the GKE equivalent of "nothing works": a fresh GKE cluster comes up, the serving stack installs cleanly, but the DRA-based scheduling path is dead on arrival.

## Root cause

`compose-serving-stack` installs the `dra-driver-nvidia-gpu` chart (v0.4.0), whose `dra-driver-nvidia-gpu-kubelet-plugin` daemonset runs at `priorityClassName: system-node-critical`.

GKE Standard restricts `system-node-critical` / `system-cluster-critical` priority pods: they are only admitted in a namespace that has a `ResourceQuota` explicitly permitting them (GKE auto-injects a `gcp-critical-pods` quota into its own managed namespaces like `kube-system`, `gke-managed-system`; user namespaces get nothing). The DRA driver lands in its own `dra-driver-nvidia-gpu` namespace with no such quota, so pod creation is rejected:

```
DaemonSet dra-driver-nvidia-gpu-kubelet-plugin
  Desired: 1   Current: 0   Ready: 0
Events:
  Warning  FailedCreate  (x18)  daemonset-controller
    Error creating: insufficient quota to match these scopes:
      [{PriorityClass In [system-node-critical system-cluster-critical]}]
```

Downstream symptom on the workload nodes (A100 present, node untainted, GPU allocatable):

```
kubectl get resourceslices            # -> none
ModelReplica pod: Pending
  FailedScheduling: 0/N nodes are available: 1 cannot allocate all claims, ...
  NotTriggerScaleUp: ... cannot allocate all claims
```

## Reproduction

1. Provision a GKE Standard `InferenceCluster` with a GPU node pool (e.g. `a2-highgpu-1g`) and let the serving stack install.
2. Deploy a `ModelDeployment` whose GPU device uses `claim: DRA`.
3. Observe: serving stack `Ready=True`, but `kubectl -n dra-driver-nvidia-gpu get ds` shows `CURRENT 0`, `kubectl get resourceslices` is empty, and the replica pod stays `Pending` with `cannot allocate all claims`.

Confirmed on two fresh GKE Standard clusters (`us-east1`, `us-west1`), chart `dra-driver-nvidia-gpu:0.4.0`.

## Workaround (manual, live-only)

Create a permitting quota in the driver namespace on each workload cluster, then the daemonset schedules and publishes slices:

```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: allow-critical-pods
  namespace: dra-driver-nvidia-gpu
spec:
  hard:
    pods: "1000"
  scopeSelector:
    matchExpressions:
    - operator: In
      scopeName: PriorityClass
      values: [system-node-critical, system-cluster-critical]
```

## Proposed fix

`compose-serving-stack` should make the DRA driver schedulable on GKE out of the box. Options:

1. **Compose the permitting `ResourceQuota`** in the DRA driver's namespace (scoped to the two system priority classes) alongside the Helm release — mirrors what GKE does for its managed namespaces. Cloud-agnostic and safe.
2. **Drop the `system-node-critical` priority** for the DRA driver kubelet-plugin on GKE (use a high but non-system PriorityClass), via chart values.

Option 1 is the smaller, more portable change.

## Secondary issue (readiness gap)

The `ServingStack` reports `Ready=True` while the DRA driver can't schedule. Readiness should reflect the DRA driver actually running / publishing a `ResourceSlice`, so a broken GPU path doesn't look healthy.

dennis-upbound:

On GKE, no GPU model can serve: the NVIDIA DRA driver never starts, so it never publishes `ResourceSlice`s, and every `claim: DRA` request stays `Pending` with `cannot allocate all claims`. Two independent GKE-specific defects cause this:

**1. Priority quota.** The driver's `kubelet-plugin` daemonset runs at `system-node-critical` priority. GKE only admits `system-node-critical` / `system-cluster-critical` pods in a namespace that has a `ResourceQuota` permitting those priority classes (it injects a `gcp-critical-pods` quota into its own managed namespaces, not into the driver's). So the daemonset is rejected:

```
FailedCreate: insufficient quota to match these scopes:
  [{PriorityClass In [system-node-critical system-cluster-critical]}]
```

**2. Driver root.** GKE installs the NVIDIA driver under `/home/kubernetes/bin/nvidia`, but the chart defaults `NVIDIA_DRIVER_ROOT` to `/`, so the kubelet plugin's init container can't find `nvidia-smi` / `libnvidia-ml` and never starts.

### Approach

**Quota — composed unconditionally.** `compose-serving-stack` lays the permitting `ResourceQuota` down in the driver's namespace on every cluster. We only *know* GKE needs it, but it only **grants** headroom for those two priority classes and constrains nothing, so it's harmless where they aren't restricted (EKS, self-managed). (Per review: simpler than inferring it from a cloud signal.)

**Driver root — cluster-driven.** This one is genuinely cloud-specific (setting it on EKS would be wrong), so the cluster composition decides and the serving stack only consumes:
- `ServingStack` gains `spec.nvidiaDriverRoot`, defaulting to `/`.
- `compose-inference-cluster` sets it to `/home/kubernetes/bin/nvidia` for provisioned **GKE** clusters; EKS / existing keep the default.
- `compose-serving-stack` passes a non-default value through to the DRA driver chart.

No "where am I running" branching inside ServingStack.

### How verified

- Unit: `compose-inference-cluster` asserts the GKE-path ServingStack carries `nvidiaDriverRoot`; `compose-serving-stack` asserts a non-default root produces the chart override and the quota is composed on the default root too (no override). `pytest` green, `ruff` clean. Full `nix flake check` in CI.
- Live on GKE (earlier, via the equivalent manual config): the DRA driver starts, publishes a `ResourceSlice`, and a `claim: DRA` model schedules and serves.

Closes #202.

### I have:

- [x] Read and followed Modelplane's contribution process.
- [ ] ~Run `nix flake check` locally~ — ran the affected function tests + ruff + format; full check runs in CI.
- [x] Added or updated tests covering the composition change.


apis/servingstacks/definition.yaml
```diff
@@ -76,6 +76,20 @@ spec:
...
+                description: >-
+                  Host path where the NVIDIA driver is installed, passed to the
+                  DRA driver as nvidiaDriverRoot. Defaults to / (the upstream
+                  default), which suits EKS and self-managed clusters. Set it for
+                  platforms that install the driver elsewhere — GKE uses
+                  /home/kubernetes/bin/nvidia. A non-default value also makes the
+                  serving stack compose a ResourceQuota permitting the DRA
+                  driver's system-critical pods, which GKE requires. The cluster
+                  composition sets this; the serving stack never inspects its own
+                  cloud.
```
negz, line 91: Nit: inferring that we need this ResourceQuota because we set *any* non-default feels a bit magic.

I suggest we just lay down the ResourceQuota everywhere and add a comment noting we only *know* GKE needs it but don't expect it to hurt any other cluster.

negz (approved):

PR #235, opened 2026-06-21
docs: Versioning

### LINKED ISSUE #130: docs: Create versioning scheme
Versioning should be branch based as opposed to directory based

tr0njavolta:
Add docs versioning via release-branch subdomains

Each minor release gets its own subdomain (v0.1.docs.modelplane.ai, v0.2.docs.modelplane.ai, …) served from the existing Vercel project using branch domains. No new Vercel project is needed per release — just assign the subdomain to the release branch in the dashboard and set HUGO_BASEURL.

A wildcard CNAME (*.docs.modelplane.ai) covers every future release automatically, so there's no DNS change per release either.

What this adds:

- Version dropdown on every page of release builds, reading from docs/data/versions.yaml. Hidden on main — there's nothing to switch between from the root.
- docs/data/versions.yaml with the v0.1 entry pointing at https://v0.1.docs.modelplane.ai.
- docs/vercel-build.sh updated to accept HUGO_BASEURL in production, so release branch projects set their subdomain URL as an environment variable with no code change.
- CONTRIBUTING.md documents the per-release steps and the one-time DNS setup.

What to do after this merges: see CONTRIBUTING.md § Versioning 

Fixes #130 

I have: 


CONTRIBUTING.md
```diff
@@ -420,6 +420,48 @@ rather than the dev version it derives from git metadata on ordinary runs.
...
+   ```
+4. Update `docs/data/versions.yaml` on `main`:
+   ```yaml
+   versions:
+     - version: "0.1"
+       url: "https://docs.modelplane.ai/v0.1"
+     - version: "main"
+       url: ""
+   ```
+5. Set `latest = "0.1"` in `docs/hugo.toml` on `main`.
```
negz: It'd be nice to avoid this process, but I'm not sure we can. The best alternative I could find was a pattern some folks (apparently) use where:

* Each release branch of the docs site is its own Vercel project
* There's a central Vercel project that routes to them all as if they were external sites

I'm on the fence about which approach is better. I imagine juggling a whole bunch of Vercel projects would be a different kinda pain.

tr0njavolta: I would be okay with the multi Vercel project approach if we plan on limiting the number of versions for docs we want to serve.

negz (approved):
The release process LGTM. If it ends up being a pain to maintain multiple Vercel sites we can try the original approach of building all the branches -> an artifact on main.

negz (approved):

PR #243, opened 2026-06-22
Add a Supported Providers page to the docs

bassam:

The docs never answered "where can I run Modelplane, and where is it headed?" in one place. Cloud and neocloud support was scattered across the platform pages, the FAQ ("a few hyperscalers and neoclouds"), and per-cloud tabs, which undersold both what ships today and how far the Crossplane provider ecosystem reaches.

This adds a Supported Providers page under the platform section. It frames support as three states a provider can be in: Modelplane provisions the cluster (GKE and EKS today), you bring your own conformant cluster with `source: Existing`, or a Crossplane provider exists that marks where native provisioning can land next. A single alphabetical table (deliberately neutral, no ranking) covers the major clouds and a curated set of GPU-specialist neoclouds. An Accelerators column shows the silicon each offers as compact vendor pills (NVIDIA, AMD, TPU, Trainium, Ascend) via a new `accel` shortcode that borrows the website's pill treatment and monochrome logos, and the Crossplane column links each provider repo through a GitHub icon (`repolink` shortcode). An on-premises and bare-metal note covers DGX BasePOD/SuperPOD and bare-metal Kubernetes: supported today via bring-your-own, with provisioning through NVIDIA Base Command Manager or other bare-metal provisioners on the roadmap.

Building the five-column matrix surfaced a pre-existing theme bug: `.table-responsive` set `overflow: hidden`, which overrode Bootstrap's `overflow-x: auto`, so any table wider than the viewport crammed its columns and clipped instead of scrolling. The wrapper scrolls again, and the table render hook tags many-column tables so they keep their natural width and fit the measure on desktop while still scrolling on phones.

The FAQ also gains a direct "Does Modelplane support NVIDIA?" entry covering the stack — GPUs bound through DRA and CEL selectors, NIM microservices and the TensorRT-LLM engine running as engine containers, weights and NIM-style artifacts from NVIDIA NGC, and NVIDIA Dynamo in the installed inference stack with deeper integration on the roadmap — and the FAQ and the Register a Cluster page link to the new page.

The change is documentation only; no composition functions changed.

I have:

- [ ] ~~Added or updated tests covering any composition function changes.~~


negz (approved):
LGTM overall. I like the idea.

I do notice the table's too wide to render the Crossplane provider support on my 14" MacBook Pro. I have to scroll to the side to see it. Not a dealbreaker.

Also, I wonder if we need to distinguish cloud from Neocloud? Does it matter WRT this list?

PR #264, opened 2026-06-23
Add an example for collecting engine metrics

pluna:
The examples show how to serve a model but not how to observe one. An operator who wants an engine's Prometheus metrics has no page that shows the wiring.

This adds that page. vLLM exposes `/metrics` on its serving port with no extra flag, and Modelplane already runs a Prometheus on each workload cluster with open `PodMonitor` discovery, so scraping the engine is a `PodMonitor` selecting the engine pods plus a `port-forward`. The engine container port is unnamed, so the `PodMonitor` matches it by number with `targetPort`. A per-shape section covers the deltas: SGLang needs `--enable-metrics`, leader/worker is scraped on the leader only, and prefill/decode splits into two role-labelled engines with the decode engine on `8001` behind the routing sidecar.

A 0.5B Qwen model on a single L4 is the subject, chosen as the cheapest thing to serve; the platform layer reuses the existing `gke-l4-1x-g2` class and `gke-l4-single` cluster. I validated it end to end on a GKE L4: the deployment served a completion, and the example's `PodMonitor` made `vllm:num_requests_running` show up in the in-cluster Prometheus. The per-shape notes for SGLang, leader/worker, and prefill/decode are grounded in the composition functions, not run end to end.

`accept.txt` gains `sidecar` for the Vale spelling check.

I have:

### DIFF
--- a/docs/content/examples/collecting-engine-metrics.md
+++ b/docs/content/examples/collecting-engine-metrics.md
@@ -12,7 +12,7 @@ namespaces, so scraping the engine is a `PodMonitor` plus a `port-forward`. The
 model is only the subject; the same wiring fits any engine, with the SGLang,
 leader/worker, and prefill/decode differences noted at the end.
 
-This was run end to end on GKE; the `InferenceClass` and `ModelDeployment` are the
+This was run end to end on GKE. The `InferenceClass` and `ModelDeployment` are the
 exact manifests from that run, and the `PodMonitor` below scraped this deployment.
 Apply the platform side first, then the ML side. The GKE `InferenceCluster`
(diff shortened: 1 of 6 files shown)


dennis-upbound (approved):
nice!

tr0njavolta:
@pluna the Vercel build is returning ```The deployment was blocked because the commit email pablo@upbound.io could not be matched to a GitHub account. Ensure your git email matches your GitHub account.``` for this and https://github.com/modelplaneai/modelplane/pull/265. Would you mind checking out https://vercel.com/docs/deployments/troubleshoot-project-collaboration#account-configuration to fix?

PR #274, opened 2026-06-23
Redirect the docs root to the latest release

negz:

`docs.modelplane.ai` is built from `main` and served the dev docs at its root, so readers landing on the bare domain saw unreleased docs instead of the latest release. The version dropdown's own comment already assumed the root redirects to latest — a behavior that was never wired up.

This makes the apex redirect to the latest release while keeping the dev docs browsable on their own subdomain:

| URL | Serves |
|---|---|
| `docs.modelplane.ai` | redirect to the latest release |
| `main.docs.modelplane.ai` | the dev (`main`) docs |
| `vX-Y.docs.modelplane.ai` | each minor release |

The apex and `main.docs` builds both come from `main` and are both `version == "main"`; what tells them apart is the baseURL. The home page redirects only when the build's baseURL isn't `main`'s own subdomain — both URLs read from `versions.yaml`. The redirect target is the latest release's URL, keyed by `params.latest` in `hugo.toml`, so the dropdown and the redirect share one source of truth. The redirect uses the same `window.location.replace` + `noscript` meta-refresh that `list.html` already uses for section index pages.

`versions.yaml` now lists every version, including `main`, each with a real URL, so the dropdown renders on every build (its old guard skipped `main`). `latest` moves from `"main"` to `"0.1"`.

`versions.yaml` and the home/dropdown templates must stay byte-identical across branches so every build offers the same switcher. The matching release-side change is in modelplaneai/modelplane#275.

**Vercel setup this depends on:** add `main.docs.modelplane.ai` to the `modelplane-docs` project on the `main` branch, with a Production env var `HUGO_BASEURL=https://main.docs.modelplane.ai/` scoped to `main`. Without it the apex still redirects correctly, but `main.docs` won't resolve.

I have:

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -22,22 +22,47 @@ release from it and run the workflow against the new tag.
 
 ## Versioning the docs
 
-Docs are versioned at the minor level. Each minor release is served from its own
-subdomain (`v0-1.docs.modelplane.ai`, `v0-2.docs.modelplane.ai`, …) using a single
-Vercel project with one branch domain per release. The `main` branch always serves
-the latest docs at `docs.modelplane.ai`. Patch releases push to the existing release
-branch and need no changes to `main`.
+Docs are versioned at the minor level, each version on its own subdomain, served
+by a single Vercel project with one branch domain per build:
+
+- `docs.modelplane.ai` — the canonical apex. Its home page redirects to the
+  latest release, so a reader who types the bare domain lands on the latest docs.
+- `vX-Y.docs.modelplane.ai` — each minor release (`v0-1`, `v0-2`, …), built from
+  its `release-X.Y` branch.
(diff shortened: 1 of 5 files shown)


negz:
Both valid — fixed in the latest push.

- index.html: the redirect condition now also requires `$mainURL` to be non-empty, so a missing/empty `main` entry in versions.yaml makes the dev-docs build fail open (serve content) rather than redirect.
- version-dropdown-menu.html: restored the defensive `$versions` check alongside the length test so an unset data file can't error `len`.

#275 carries the byte-identical template changes.

tr0njavolta:
Had to move from branches in per environment to release-per-project. The v0-1 domain is tracking `release-0.1`

tr0njavolta (approved):

PR #280, opened 2026-06-23
docs: Update analytics string for prod environments only

tr0njavolta:
os.Getenv "VERCEL_ENV" always returns "" in a pure Nix build. Replace it with hasPrefix site.BaseURL "https://docs.modelplane.ai", which is the only URL that build produces.

Fixes #

I have: 


bassam:
Will this work with version branches?

tr0njavolta:
Updated with a string contains statement to capture the domain for prod environments

bassam (approved):

PR #307, opened 2026-06-24
Add a speculative-decoding serving example for Qwen3-8B

pluna:

The Examples page covers single-engine, multi-node, and disaggregated serving, but nothing for speculative decoding, a common way to cut single-GPU decode latency. This adds an example for n-gram (prompt-lookup) speculation, which proposes tokens by matching the prompt and so runs as one `Standalone` vLLM engine with no draft model or extra cached weights.

The page serves Qwen3-8B on a single L4 and benchmarks the speculative engine against an identical baseline without `--speculative-config`. On a copy-heavy edit workload the speculative engine reaches 2.4 times the output-token throughput at about half the time per output token, with a 65% draft acceptance rate. I ran it end to end on GKE: the cluster provisioned, the deployment served a valid completion, and the benchmark numbers in the page come from that run.

I have:

- [ ] ~~Added or updated tests covering any composition function changes.~~ No composition functions changed; this is docs only.

### DIFF
--- a/docs/content/examples/qwen3-8b-speculative-decoding.md
+++ b/docs/content/examples/qwen3-8b-speculative-decoding.md
@@ -1,0 +1,68 @@
+---
+title: Qwen3-8B speculative decoding
+weight: 60
+description: An 8.2B dense chat model on a single L4 with n-gram speculative decoding.
+---
+<!-- vale write-good.Passive = NO -->
+This example sets up n-gram (prompt-lookup) speculative decoding for Qwen3-8B on a
+single NVIDIA L4. On copy-heavy output, editing a pasted code block where most
+output tokens are copied from the prompt, it roughly doubles decode throughput and
+halves the time per output token:
+
+| Metric | Without speculation | With n-gram speculation |
+|---|---|---|
+| Output token throughput (tok/s) | 16.10 | 39.01 |
+| Mean TPOT (ms/token) | 60.20 | 24.21 |
+
+Measured on a single L4 (`vllm/vllm-openai:v0.23.0`, Qwen3-8B, 30 copy-heavy
+prompts at concurrency 1) against the same model without `--speculative-config`;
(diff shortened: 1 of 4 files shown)


docs/manifests/examples/qwen3-8b-speculative-decoding/model-service-baseline.yaml
```diff
@@ -0,0 +1,14 @@
+# Exposes the qwen3-8b-base deployment as an OpenAI-compatible URL, so the
+# benchmark can hit the baseline engine the same way it hits the speculative one.
+# Read its address from status.address:
+#   kubectl get ms qwen3-8b-base -n ml-team -o jsonpath='{.status.address}'
```
negz: Does this duplicate the existing example?

Do we need to include everything required here to actually run the experiment, or should we just include the final result? I would prefer the latter.

pluna: Fixed

docs/manifests/examples/qwen3-8b-speculative-decoding/model-deployment-baseline.yaml
```diff
@@ -0,0 +1,38 @@
+# No-speculation baseline for the benchmark: Qwen3-8B on a single L4 without
```
negz: @pluna I think this should be removed too. 😄 Once this is gone we can merge.

pluna: @negz The Qwen3-8B example has a different deployment configuration. Existing example has tool-call-parser and reasoning-parser args, which I didn't want to have in the speculative decoding, so I had to create a new one for base and spec to have only one difference --speculative-config.

dennis-upbound (approved):

PR #314, opened 2026-06-25
Name each serving-stack Helm release after its chart

### LINKED ISSUE #215: ServingStack NVIDIA DRA driver fails on Cilium clusters: service-account name exceeds 63-char CiliumIdentity label limit
### What happened?

On a cluster whose CNI is **Cilium** (seen on Nebius Managed Kubernetes, but it applies to any Cilium cluster), the ServingStack's **NVIDIA DRA driver kubelet-plugin pod never starts**, so the InferenceCluster's `BackendReady` condition stays `False` forever and no GPU is ever bound. Everything else in the serving stack (cert-manager, Envoy Gateway, LeaderWorkerSet, Prometheus, NFD) comes up fine; the `gpu.nvidia.com` DeviceClass even registers — but the kubelet plugin is stuck `Pending` and publishes **zero ResourceSlices**.

### Root cause

The pod can't get a network sandbox:

```
Warning  FailedCreatePodSandBox  kubelet  Failed to create pod sandbox: rpc error: code = Unknown
desc = failed to setup network for sandbox "...": plugin type="cilium-cni" failed (add):
unable to create endpoint: Cilium API client timeout exceeded
```

The cilium agent log on the node shows why:

```
level=warning msg="Key allocation attempt failed" error="unable to allocate ID 36359 for key
[... k8s:io.cilium.k8s.policy.serviceaccount=nebius-llama-23b5b4e31ad4-dra-driver-nvidia-gpu-service-account-kubeletplugin ...]:
CiliumIdentity.cilium.io \"36359\" is invalid: metadata.labels: Invalid value:
\"nebius-llama-23b5b4e31ad4-dra-driver-nvidia-gpu-service-account-kubeletplugin\":
must be no more than 63 characters" subsys=allocator
```

Cilium derives a `CiliumIdentity` for every pod and copies the pod's **service-account name** into the **label value** `io.cilium.k8s.policy.serviceaccount`. Kubernetes [label values are capped at 63 characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). The DRA driver's service account is named:

```
<helm-release>-dra-driver-nvidia-gpu-service-account-kubeletplugin   # 76 chars
```

where `<helm-release>` is `<inferencecluster-name>-<hash>` (here `nebius-llama-23b5b4e31ad4`). The name exceeds 63 chars, so Cilium rejects the identity → the pod never gets an endpoint → never schedules → DRA never functions.

The chart's own suffix is already 51 chars (`-dra-driver-nvidia-gpu-service-account-kubeletplugin`); with the release-hash prefix it exceeds 63 **regardless of how short the InferenceCluster name is**, so it can't be worked around from the API — it needs a `fullnameOverride` / short `serviceAccount.name` on the release.

**Offending code:** `functions/compose-serving-stack/function/fn.py:606` (`compose_dra_driver`) builds the `dra-driver-nvidia-gpu` Helm release (`fn.py:~635`) without overriding the chart's fullname / serviceAccount, so the SA inherits the long composed release name.

On EKS/GKE this is harmless — their CNIs don't key a per-pod identity on the SA name — which is why it hasn't shown up before.

### Upstream status (Cilium)

This is a known Cilium bug: [cilium/cilium#16579](https://github.com/cilium/cilium/issues/16579). It was fixed in [cilium/cilium#39552](https://github.com/cilium/cilium/pull/39552), which **removes the `io.cilium.k8s.policy.serviceaccount` label from `CiliumIdentity` entirely** (the label wasn't actually consumed by Cilium), rather than truncating or hashing the name.

The upstream fix does **not** resolve this issue for us in practice:

- It was merged to `main` (2025-05-20) and **deliberately not backported**, so it only exists in Cilium builds cut from main after that date.
- The affected clusters run **managed Cilium** (e.g. Nebius), where we don't choose the Cilium build and can't reconfigure identity-relevant labels.

So we can't rely on the upstream fix reaching the clusters that hit this. The fix has to live on the modelplane side.

### Fix

In `compose_dra_driver`, pass the chart a short, fixed `fullnameOverride` (and/or `serviceAccount.name`, e.g. `dra-gpu`) so the kubelet-plugin SA stays well under 63 chars independent of the InferenceCluster name.

**Recommendation / general rule:** any name that may be reused as a Kubernetes **label value** or **DNS-1123 label** must stay **≤63 characters**, with headroom for suffixes that charts/controllers append. The ServiceAccount object itself permits up to 253 chars, which is why creation succeeds — but anything that copies the name into a 63-char field (as Cilium does) then fails downstream. This is a recurring ecosystem footgun, not Cilium-specific (e.g. [aws/aws-cdk#23643](https://github.com/aws/aws-cdk/issues/23643) — "serviceAccountName should not be used as a label value; it can exceed 63 characters"). Keeping generated names short by default avoids the whole class of failures across CNIs and controllers.

Refs:
- Kubernetes — [Labels and Selectors (63-char value limit)](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
- [aws/aws-cdk#23643](https://github.com/aws/aws-cdk/issues/23643)

### How can we reproduce it?

1. Have a Cilium-CNI cluster (e.g. Nebius mk8s 1.34) with a GPU node (NVIDIA driver preinstalled).
2. Put its kubeconfig in a Secret in `modelplane-system`, label the GPU node `modelplane.ai/pool=<pool>`.
3. Apply an `InferenceClass` (DRA `gpu.nvidia.com`) and an `InferenceCluster` (`source: Existing`) — using a normal-length name like `nebius-llama`:
   ```
   kubectl apply -f inference-class.yaml -f inference-cluster.yaml
   ```
4. Watch the ServingStack install on the workload cluster:
   ```
   kubectl --kubeconfig <workload> -n dra-driver-nvidia-gpu get pods
   # dra-driver-nvidia-gpu-kubelet-plugin-xxxxx   0/1   Pending  (FailedCreatePodSandBox, cilium timeout)
   kubectl --kubeconfig <workload> get resourceslices   # none
   kubectl get inferencecluster <name> -o jsonpath='{..conditions[?(@.type=="BackendReady")].status}'  # False (Installing) forever
   ```
5. Confirm in the cilium agent log: `CiliumIdentity ... must be no more than 63 characters` on the `...service-account-kubeletplugin` SA.

Modelplane version: v0.1.0-rc.1
- Crossplane: v2.3.2
- provider-helm: v1.2.0; provider-kubernetes: v1.2.1
- Workload cluster: Nebius Managed Kubernetes, Kubernetes v1.34.8, **CNI = Cilium**
- Control plane: KIND
- Inference backend: vLLM v0.7.3 (Llama 3.1 8B), single NVIDIA H200
- NVIDIA DRA driver chart: `dra-driver-nvidia-gpu` v0.4.0 (`NVIDIA_DRIVER_ROOT=/`)

dennis-upbound:

Fixes #215.

provider-helm names each Helm release after the composed resource, whose generated name is `<inferencecluster>-<hash>`. That name is long, unpredictable, and length-capped, and it flows into every chart's fullname and from there into the names of the resources the chart creates. A chart truncates its own fullname to 63 chars and then appends suffixes — the NVIDIA DRA driver's kubelet-plugin ServiceAccount is `<fullname>-service-account-kubeletplugin` — so the final name can exceed 63 even when the fullname did not. Kubernetes label values are capped at 63 chars, so any consumer that copies such a name into a label value rejects it. On Cilium clusters the per-pod CiliumIdentity keys on the ServiceAccount name, so the DRA driver's kubelet-plugin pod never gets a network sandbox, never schedules, and no GPU is ever bound.

This isn't specific to the DRA driver. Rendering every chart at a realistic release name shows the DRA driver overflows at essentially any InferenceCluster name, while `node-feature-discovery` and `lws` overflow once the name is a little longer, and the kube-prometheus-stack node-exporter subchart overflows the same way (it ignores the chart's `fullnameOverride` and uses the release name directly). The shared cause is the release name, so the fix belongs there rather than per chart.

`_helm_release` now sets `crossplane.io/external-name` to `mp-<chart>`, which provider-helm uses verbatim as the release name. Each release is named for its chart (`mp-dra-driver-nvidia-gpu`) instead of the InferenceCluster, so every derived name is short and independent of the InferenceCluster name: the budget below holds for any deployment, not only the ones tested. The `mp-` prefix reserves a release-name namespace for the releases this function owns; without it a release named `cert-manager` would adopt or collide with a `cert-manager` release a user already runs in that namespace, and `mp-cert-manager` cannot.

| Limit | Tightest case | Length |
|---|---|---|
| Helm release name ≤ 53 | `mp-node-feature-discovery` | 25 |
| ServiceAccount name ≤ 63 | `mp-dra-driver-nvidia-gpu-service-account-kubeletplugin` | 54 |

The DRA driver is the binding constraint at 54/63: it appends the longest suffix (`-service-account-kubeletplugin`, 30 chars) to a release name that already contains the chart name. Every other chart is smaller. The release name is stable across chart-version upgrades, so provider-helm upgrades in place rather than treating each version as a new release, and the composed resource keeps its generated, hashed `metadata.name`, so managed-resource uniqueness in the control plane is unchanged.

This does change release identity. On an existing install provider-helm installs fresh releases under the new `mp-<chart>` names and orphans the old `<inferencecluster>-<hash>` releases, so migrating a running stack means a teardown and reinstall. That's acceptable pre-1.0, and the Cilium clusters this fixes are non-functional today regardless.

I have:

### DIFF
--- a/functions/compose-serving-stack/function/fn.py
+++ b/functions/compose-serving-stack/function/fn.py
@@ -47,6 +47,15 @@ from models.io.k8s.apimachinery.pkg.apis.meta import v1 as metav1
 # Label key for composed resources that need deletion ordering via Usages.
 _LABEL_RESOURCE = "modelplane.ai/resource"
 
+# Annotation provider-helm reads as the Helm release name (see _helm_release).
+_EXTERNAL_NAME_ANNOTATION = "crossplane.io/external-name"
+
+# Reserved prefix for the Helm release names this function manages. Names every
+# release "mp-<chart>" so it's clearly modelplane-owned and can't adopt or
+# collide with a same-named release a user already runs in the same namespace
+# (e.g. a cluster's own "cert-manager"). See _helm_release.
+_RELEASE_NAME_PREFIX = "mp-"
+
 # CEL readiness query for the Envoy Gateway Object. The Gateway's LoadBalancer
 # address is assigned asynchronously by the controller after the Object is
 # applied. With the default SuccessfulCreate policy the Object is Ready the
@@ -131,20 +140,28 @@ def _helm_release(
     labels: dict | None = None,
     metadata_namespace: str | None = None,
 ) -> helmv1beta1.Release:
-    """Build a Helm Release targeting a remote (or local) cluster."""
-    md = None
-    if labels or metadata_namespace:
+    """Build a Helm Release targeting a remote (or local) cluster.
+
+    The crossplane.io/external-name annotation sets a fixed "mp-<chart>" release
+    name, which provider-helm uses verbatim. Left unset, the release inherits the
+    composed resource's generated "<inferencecluster>-<hash>" name; charts derive
+    resource names from it and append suffixes, so names can exceed the 63-char
+    label-value limit and break consumers like Cilium's per-pod CiliumIdentity.
+    "mp-<chart>" keeps every derived name short regardless of InferenceCluster
+    name (worst case is the DRA driver's ServiceAccount at 54 chars), is stable
+    across chart-version upgrades so provider-helm upgrades in place, and its
+    "mp-" prefix reserves a namespace for the releases this function owns. Each
+    chart runs once per workload cluster, so the name is unique. See issue #215."""
+    md = metav1.ObjectMeta(
+        annotations={_EXTERNAL_NAME_ANNOTATION: f"{_RELEASE_NAME_PREFIX}{chart}"},
         # Only set fields that are present; under exclude_unset, an explicit
         # namespace=None or labels=None would leak a null into the metadata.
-        md = metav1.ObjectMeta(
-            **({"namespace": metadata_namespace} if metadata_namespace is not None else {}),
-            **({"labels": labels} if labels is not None else {}),
-        )
+        **({"namespace": metadata_namespace} if metadata_namespace is not None else {}),
+        **({"labels": labels} if labels is not None else {}),
+    )
 
     release = helmv1beta1.Release(
-        # Only set metadata when present (see _k8s_object: avoids a null
-        # metadata leaking under exclude_unset serialization).
-        **({"metadata": md} if md is not None else {}),
+        metadata=md,
         spec=helmv1beta1.Spec(
             providerConfigRef=helmv1beta1.ProviderConfigRef(
                 kind="ProviderConfig",
--- a/functions/compose-serving-stack/tests/test_fn.py
+++ b/functions/compose-serving-stack/tests/test_fn.py
@@ -135,6 +135,7 @@ _USAGE_GATEWAY_CLASS_BY_GATEWAY = {
 _CERT_MANAGER = {
     "apiVersion": "helm.m.crossplane.io/v1beta1",
     "kind": "Release",
+    "metadata": {"annotations": {"crossplane.io/external-name": "mp-cert-manager"}},
     "spec": {
         "forProvider": {
             "chart": {
@@ -161,6 +162,7 @@ _ENVOY_GATEWAY = {
     "apiVersion": "helm.m.crossplane.io/v1beta1",
     "kind": "Release",
     "metadata": {
+        "annotations": {"crossplane.io/external-name": "mp-gateway-helm"},
         "labels": {"modelplane.ai/resource": "envoy-gateway"},
     },
     "spec": {
@@ -215,6 +217,7 @@ _ENVOY_GATEWAY = {
 _AI_GATEWAY_CRDS = {
     "apiVersion": "helm.m.crossplane.io/v1beta1",
     "kind": "Release",
+    "metadata": {"annotations": {"crossplane.io/external-name": "mp-ai-gateway-crds-helm"}},
     "spec": {
         "forProvider": {
             "chart": {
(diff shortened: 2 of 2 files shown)


functions/compose-serving-stack/function/fn.py
```diff
@@ -131,20 +140,38 @@ def _helm_release(
...
+    #215.
+
+    A fixed "mp-<chart>" name keeps every derived name well under 63 regardless
+    of the InferenceCluster name: the longest is the DRA driver's kubelet-plugin
+    ServiceAccount, "mp-dra-driver-nvidia-gpu-service-account-kubeletplugin" at
+    54 chars. Each chart runs once per workload cluster in its own namespace, so
+    "mp-<chart>" is a unique, stable release name (stable across chart-version
+    upgrades, so provider-helm upgrades in place rather than reinstalling). The
+    composed resource keeps its own generated metadata.name, unique in the
+    control plane. The "mp-" prefix is reserved for releases this function owns."""
```
negz: Nit: This information is important but the comment feels ~3x as long as it needs to be to communicate it.

functions/compose-serving-stack/tests/test_fn.py
```diff
@@ -718,6 +729,53 @@ async def test_default_driver_root_skips_override_keeps_quota(self) -> None:
         ]
         self.assertNotIn("nvidiaDriverRoot", dra_values)
 
+    async def test_release_names_short_regardless_of_inferencecluster_name(self) -> None:
```
negz: Do we need this new test? The cases above cover the behavior and would capture a regression.

negz (approved):
Thanks @dennis-upbound! Few nits but looks good overall.

PR #320, opened 2026-07-06
Set the inference gateway Service to externalTrafficPolicy Cluster

### LINKED ISSUE #316: BYO InferenceCluster on Nebius never reaches Ready: gateway LoadBalancer Service uses externalTrafficPolicy: Local
### What happened?

On a Nebius Managed Kubernetes cluster registered as a BYO `InferenceCluster`
(`source: Existing`), the serving stack installs but the cluster never reaches
`Ready`, so no model can be deployed to it.

The blocker is the inference gateway's `Service`. The serving stack creates an
Envoy Gateway `Service` of type `LoadBalancer`, and Envoy Gateway sets
`externalTrafficPolicy: Local` on it by default. Nebius's load balancer rejects
that value, so the `Service` never gets an external IP:

```
$ kubectl describe svc envoy-modelplane-system-inference-gateway-... -n envoy-gateway-system
  Type     Reason                  Message
  Warning  SyncLoadBalancerFailed  Error syncing load balancer: failed to ensure
                                    load balancer: externalTrafficPolicy: "Local"
                                    is unsupported
$ kubectl get svc ... -n envoy-gateway-system
  ... LoadBalancer 10.96.39.66 <pending> 80:31558/TCP
```

With the gateway address stuck `<pending>`, readiness fails up the chain:

- `ServingStack`: `Ready=False`, `Unready resources: gateway`
- `InferenceCluster`: `Ready=False`, `BackendReady=False`
- `ModelDeployment`: `ReplicasScheduled=False`, reason `InsufficientCapacity`
  (`0 of 1 replicas scheduled (checked 1 clusters)`)

The scheduler skips clusters that are not ready, so a deployment that targets a
Nebius cluster stays unscheduled indefinitely. Nothing in the serving
stack functions sets `externalTrafficPolicy`, so the gateway `Service` inherits
the Envoy Gateway default on every workload cluster.

Expected: a BYO `InferenceCluster` on Nebius reaches `Ready` and serves models,
the same as on GKE and EKS.

What Nebius expects: its load balancer assigns a public IP to a `LoadBalancer`
`Service` by default, with no extra configuration ([Nebius load balancer
docs](https://docs.nebius.com/kubernetes/clusters/load-balancer)). The
`SyncLoadBalancerFailed` event above shows the controller rejects only
`externalTrafficPolicy: Local`, so it accepts the Kubernetes default of
`Cluster`. The gateway should provision a working load balancer on Nebius once
its `Service` uses a value the controller accepts.

### How can we reproduce it?

1. Provision a Nebius `mk8s` cluster (Kubernetes 1.34) with a public endpoint.
2. Register it as a BYO cluster: apply an `InferenceClass` and an
   `InferenceCluster` with `spec.cluster.source: Existing` pointing at the
   cluster's kubeconfig secret.
3. Wait for the serving stack to install on the cluster.
4. Observe the `envoy-...-inference-gateway` `Service` in `envoy-gateway-system`
   stay `EXTERNAL-IP <pending>`, with the `SyncLoadBalancerFailed` event above.
5. Observe `InferenceCluster` stay `Ready=False` and any `ModelDeployment`
   targeting it report `InsufficientCapacity`.

Workaround: patch the gateway `Service` to `externalTrafficPolicy: Cluster`.
Nebius then assigns a public IP and the gateway becomes ready. The patch is not
durable, since Envoy Gateway owns the `Service`.

Modelplane version: development build off `main` (the serving stack names
releases `mp-<chart>`, which includes PR #314)
Crossplane: v2.3.2
Inference backend: vLLM 0.7.3
Cloud provider and cluster type: a Nebius Managed Kubernetes (`mk8s`) cluster with a public endpoint
Kubernetes: 1.34
Providers: provider-helm v1.2.0, provider-kubernetes v1.2.1

pluna:

Envoy Gateway defaults the managed LoadBalancer Service to `externalTrafficPolicy: Local`. Some clouds' load balancers reject that: on Nebius the Service emits `SyncLoadBalancerFailed` and never gets an external IP, so the gateway address stays pending, the ServingStack and InferenceCluster never reach Ready, and deployments targeting the cluster sit unscheduled. It was previously worked around by manually patching the Service after the fact.

This composes an `EnvoyProxy` that pins `provider.kubernetes.envoyService.externalTrafficPolicy` to `Cluster` and references it from the `GatewayClass` via `parametersRef`, the idiomatic Envoy Gateway way to configure the managed Service. `Cluster` is accepted by every cloud the provider runs on (GKE, EKS, Nebius), and the inference gateway does not need client source-IP preservation. The composed `gateway-proxy` resource is marked ready in `mark_readiness` alongside the other gateway resources so it does not hold the composite from Ready.

The change is composition-only with no cloud branching, so it behaves identically across clouds. I validated it live end to end on EKS (an L4 `g6.xlarge` cluster): the managed gateway Service came up with `externalTrafficPolicy: Cluster` and a real ELB address, the InferenceCluster reached Ready, and Qwen3-8B served completions through the gateway. The golden tests cover the composed EnvoyProxy, the GatewayClass `parametersRef`, and the readiness transition.

Fixes #316.

I have:

### DIFF
--- a/functions/compose-serving-stack/function/fn.py
+++ b/functions/compose-serving-stack/function/fn.py
@@ -727,8 +727,8 @@ class Composer:
         )
 
     def compose_gateway(self) -> None:
-        """Compose the GatewayClass and Gateway on the remote cluster. Gated on
-        ProviderConfigs being observed."""
+        """Compose the gateway namespace, EnvoyProxy, GatewayClass, and Gateway on
+        the remote cluster. Gated on ProviderConfigs being observed."""
         pc_observed = self.provider_configs_observed()
         pc = _pc_name(self.xr)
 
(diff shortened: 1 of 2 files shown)


negz (approved):

negz:
Thanks @pluna this looks good to me.

PR #322, opened 2026-07-06
docs: memory limit kind note

tr0njavolta:
Adds note to user to update their Docker memory limit to 16gb

Fixes #

I have: 

### DIFF
--- a/docs/content/getting-started/installation.md
+++ b/docs/content/getting-started/installation.md
@@ -19,8 +19,8 @@ illustration.
 
 **Make sure your container engine is configured to allow higher resource allocations.**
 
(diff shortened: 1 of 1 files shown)


negz:
Do we know we need 16GB? Would e.g. 12GB be enough? If someone has a 16GB laptop I don't think they could allocate 16GB to kind.

I ran the getting started guide through to installing the IG and found that needs ~9GB. It would be useful to test how much memory is needed for the most resource intensive stage of the guide (e.g. multiple ICs running).

negz (approved):

negz:
Thanks @tr0njavolta!

PR #342, opened 2026-07-13
Route Unified serving through an endpoint picker

### LINKED ISSUE #325: EPP is only wired for disaggregated serving, not aggregated
## What problem are you facing?

Modelplane only wires an endpoint picker (EPP + InferencePool) for disaggregated (`PrefillDecode`) serving. `Unified` serving falls through to a plain Kubernetes Service (`compose-model-replica/function/routing.py`, `apply()` gates the EPP on `serving.mode == "PrefillDecode"`).

So whenever an aggregated model has more than one interchangeable serving pod in a cluster (`copies > 1`, or co-located replicas of the same model), requests are sprayed round-robin with no prefix-cache or load awareness. Each pod recomputes prefixes another pod already has cached.

## How could Modelplane help?

Give aggregated serving the same KV-/load-aware endpoint picking that disaggregated serving already gets, whenever a model has more than one serving pod in a cluster. Open question: how the picker should be scoped (per replica vs per model-per-cluster) given co-located replicas fragment the view.

dennis-upbound:
Fixes #325.

Unified serving fronted a replica's pods with a plain Service, which round-robins: requests spread with no prefix-cache or load awareness, and each pod recomputes prefixes another already holds. Only PrefillDecode serving got the InferencePool and endpoint picker.

This gives Unified serving the same InferencePool and endpoint picker, with a single-profile picker config scoring by prefix cache and queue depth (no prefill/decode split).

Following review, it does so in **every** case rather than only when more than one pod serves the model. The picker is vestigial for a single pod — nothing to pick between — but the alternative is swapping the route's backend from a Service to a pool the moment a second pod appears, which drops in-flight requests. One path holds whether the replica serves from one pod or many. This removes the plain-Service path entirely (`base.serving_resources` and the `SERVICE_KEY` it composed); the pool selects the pods by the serving label they already carry, exactly as the multi-pod path did.

Also addresses the rest of the review:
- `_epp_config_yaml` → `_disaggregated_epp_config_yaml`, matching its `_unified_epp_config_yaml` sibling.
- Reuses the `_name()` helper the other composition functions share, in place of the inline `assert`/`metadata.name` idiom.
- Names the EPP pod annotation `modelplane.ai/epp-config-checksum` so it's clear which config the checksum tracks.

The picker is scoped per replica. Co-located replicas of one model on a cluster keep separate prefix caches; consolidating them into one picker per model per cluster has no owner in the per-replica resource model and belongs with the routing work in #34.

Validated with `nix flake check`: `test-compose-model-replica` (46 tests) and the docs-manifest validation pass. The unified routing tests assert a single serving pod and a multi-pod replica both compose the InferencePool + endpoint picker (ConfigMap, RBAC, Deployment, Service) and an HTTPRoute pointing at the pool, with no plain Service.

I have:

### DIFF
--- a/functions/compose-model-replica/function/routing.py
+++ b/functions/compose-model-replica/function/routing.py
@@ -18,7 +18,11 @@ The workload backends (native, llm-d, dynamo) compose engines only; this layer
 decorates them with the routing the replica's serving.mode selects. apply picks
 the strategy:
 
-- Unified fronts every engine's serving pods with one Service + HTTPRoute.
+- Unified fronts the serving pods with a GAIE InferencePool + endpoint picker,
+  so requests route by prefix cache and load rather than round-robin. One path
+  whether the replica serves from a single pod or many: the picker is vestigial
+  for one pod, but the alternative is swapping a Service + HTTPRoute for a pool
+  the moment a second pod appears, which would drop in-flight requests.
 - PrefillDecode (disaggregated) role-labels the two phase engines, injects the
   pd-sidecar on decode, and fronts both with a GAIE InferencePool + endpoint
   picker, routed to in place of a Service.
@@ -30,11 +34,20 @@ Engine flags, including --kv-transfer-config for the NixlConnector, are the
 user's; this layer injects none.
 """
 
+import hashlib
+
 from models.ai.modelplane.modelreplica import v1alpha1
 from models.io.crossplane.m.kubernetes.object import v1alpha1 as k8sobjv1alpha1
 from models.io.k8s.apimachinery.pkg.apis.meta import v1 as metav1
 
 from function.backends import base
+
+
+def _name(meta: metav1.ObjectMeta | None) -> str:
+    """The object's name, always set on resources read from the API server."""
+    if meta is None or meta.name is None:
+        raise ValueError("metadata.name is unexpectedly absent")
+    return meta.name
 
 
 def _namespace(meta: metav1.ObjectMeta | None) -> str:
@@ -46,6 +59,12 @@ def _namespace(meta: metav1.ObjectMeta | None) -> str:
 
 _EPP_IMAGE = "ghcr.io/llm-d/llm-d-inference-scheduler:v0.8.0"
 _SIDECAR_IMAGE = "ghcr.io/llm-d/llm-d-routing-sidecar:v0.8.0"
+
+# ConfigMap key and mount-path filename for each strategy's EPP config.
+# ConfigMap key and mount-path filename for the EPP config. One name for both
+# strategies: the ConfigMap is per-replica and carries a single strategy's
+# config, so the mount path never needs to encode which one.
+_EPP_CONFIG_FILE = "epp-config.yaml"
 
 # The pd-sidecar takes ENGINE_PORT (8000), so the decode engine listens here.
 _DECODE_ENGINE_PORT = 8001
@@ -131,12 +150,49 @@ schedulingProfiles:
     weight: 1
 """
 
+# EndpointPickerConfig for Unified serving: one profile scoring by prefix cache
+# and queue depth, with no prefill/decode split. A single schedulingProfile makes
+# the EPP fall back to its SingleProfileHandler, and prefix-cache-scorer still
+# needs approx-prefix-cache-producer to populate the prefix-match attribute it
+# scores on (see the disaggregated config above for why the producer pins
+# autoTune and blockSizeTokens). prefix-cache-scorer is weighted 2 to
+# queue-scorer's 1: cache locality dominates placement until queue depths
+# diverge, when the queue score breaks the tie. It's a starting default, not a
+# tuned one.
+_UNIFIED_EPP_CONFIG_TEMPLATE = """\
+apiVersion: inference.networking.x-k8s.io/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+- type: approx-prefix-cache-producer
+  parameters:
+    autoTune: false
+    blockSizeTokens: BLOCK_SIZE_TOKENS
+    maxPrefixBlocksToMatch: 256
+    lruCapacityPerServer: 31250
+- type: prefix-cache-scorer
+- type: queue-scorer
+- type: max-score-picker
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: max-score-picker
+  - pluginRef: prefix-cache-scorer
+    weight: 2
+  - pluginRef: queue-scorer
+    weight: 1
+"""
+
 _DEFAULT_KV_BLOCK_SIZE = 16
 
 
-def _epp_config_yaml(block_size: int) -> str:
-    """Render the EPP config with the engine's KV block size."""
+def _disaggregated_epp_config_yaml(block_size: int) -> str:
+    """Render the disaggregated EPP config with the engine's KV block size."""
     return _EPP_CONFIG_TEMPLATE.replace("BLOCK_SIZE_TOKENS", str(block_size))
+
+
+def _unified_epp_config_yaml(block_size: int) -> str:
+    """Render the unified EPP config with the engine's KV block size."""
+    return _UNIFIED_EPP_CONFIG_TEMPLATE.replace("BLOCK_SIZE_TOKENS", str(block_size))
 
 
 def _flag_value(args: list, *flags: str) -> str | None:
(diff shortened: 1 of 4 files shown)


functions/compose-model-replica/function/routing.py
```diff
@@ -47,6 +50,10 @@ def _namespace(meta: metav1.ObjectMeta | None) -> str:
 _EPP_IMAGE = "ghcr.io/llm-d/llm-d-inference-scheduler:v0.8.0"
 _SIDECAR_IMAGE = "ghcr.io/llm-d/llm-d-routing-sidecar:v0.8.0"
 
+# ConfigMap key and mount-path filename for each strategy's EPP config.
```
haarchri, line 63: Why two filenames? The ConfigMap is per-replica and only carries one strategy's config, so a single `epp-config.yaml` would serve both?!

Two names buys us a mode-dependent mount path and a ConfigMap whose data key changes if a replica ever switches serving.mode, for no benefit I can see...

functions/compose-model-replica/function/routing.py
```diff
@@ -131,14 +138,48 @@ def _namespace(meta: metav1.ObjectMeta | None) -> str:
...
+    lruCapacityPerServer: 31250
+- type: prefix-cache-scorer
+- type: queue-scorer
+- type: max-score-picker
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: max-score-picker
+  - pluginRef: prefix-cache-scorer
+    weight: 2
```
haarchri, line 180: fine as an opinionated default for now, but worth a one-line comment saying why 2:1 (cache locality dominates until queues diverge), so the next person knows it was chosen rather than inherited.

functions/compose-model-replica/function/routing.py
```diff
@@ -248,9 +333,10 @@ def _disaggregated(
     # The EPP's prefix-cache producer must chunk prefixes at the decode engine's
     # KV block size; derive it from the decode engine's flags (HACK, #179).
     block_size = _kv_block_size(_engine_args(out[decode_key]))
-    out["inference-pool"] = base.wrap_object(provider_config, _inference_pool(name))
+    selector = {"app": name, _LABEL_INFERENCE_SERVING: "true"}
```
haarchri: We now have two labelling conventions for "this pod serves the model" in one file: unified uses {LABEL_SERVING: serving_label(replica)}, disaggregated uses app + _LABEL_INFERENCE_SERVING. Since the disagg pods presumably carry the serving label too, can this move over as well? If they genuinely have to differ, a comment saying why.

functions/compose-model-replica/function/routing.py
```diff
@@ -397,11 +483,13 @@ def _http_route(replica: v1alpha1.ModelReplica, name: str) -> dict:
     }
 
 
-def _epp_objects(name: str, provider_config: str, block_size: int) -> dict[str, k8sobjv1alpha1.Object]:
+def _epp_objects(
```
haarchri: The EPP reads its config file once at startup, and the Deployment's pod template carries no hash of config_yaml. If the rendered config changes, e.g. we edit in future releases `--block-size` and `blockSizeTokens` moves with it, Crossplane updates the ConfigMap and the EPP keeps running the old config indefinitely (a mounted ConfigMap updates in place, but nothing re-reads it). Add a `checksum/config: <sha256(config_yaml)>` annotation to the pod template to trigger the rollout. Pre-existing on the disagg path, but this is now the shared function, so it's the right place?

functions/compose-model-replica/function/routing.py
dennis-upbound: Unified to a single `epp-config.yaml` for both strategies, and fixed the disagg test that read the old name.

functions/compose-model-replica/function/routing.py
dennis-upbound: Added a comment: prefix-cache is weighted 2:1 over queue because cache locality dominates placement until queue depths diverge, and it's a starting default rather than a tuned one.

functions/compose-model-replica/function/routing.py
dennis-upbound: Unified the disagg pool selector to the shared serving label too, so both strategies select the same way. The phase (role) labels stay, but they're for the EPP's prefill/decode filters, not pool membership — added a comment saying so.

functions/compose-model-replica/function/routing.py
dennis-upbound: Good catch. Added a `modelplane.ai/config-checksum` annotation (sha256 of the rendered config) on the EPP pod template so a config change rolls the Deployment, with a test pinning it. It lives in the shared `_epp_objects`, so the disagg path gets it too.

functions/compose-model-replica/function/routing.py
```diff
@@ -18,7 +18,10 @@
 decorates them with the routing the replica's serving.mode selects. apply picks
 the strategy:
 
-- Unified fronts every engine's serving pods with one Service + HTTPRoute.
+- Unified fronts one serving pod with a Service + HTTPRoute; when more than one
+  pod serves the model it fronts them with a GAIE InferencePool + endpoint
+  picker instead, so requests route by prefix cache and load rather than
+  round-robin.
```
negz: Should we just use an EPP all the time? It'd be vestigial for the single pod case, but presumably simplify the system.

If live-editing `copies` is something we want to support (and I'm not sure it is), would it be better to always use an EPP? Today if we bumped `count`, would replacing the Service + HTTPRoute with an InferencePool + EPP cause us to drop requests?

functions/compose-model-replica/function/routing.py
```diff
@@ -131,14 +142,51 @@ def _namespace(meta: metav1.ObjectMeta | None) -> str:
...
+  - pluginRef: prefix-cache-scorer
+    weight: 2
+  - pluginRef: queue-scorer
+    weight: 1
+"""
+
 _DEFAULT_KV_BLOCK_SIZE = 16
 
 
 def _epp_config_yaml(block_size: int) -> str:
```
negz: I'd prefer `_disaggregated_epp_config_yaml` since its sibling is now explicitly `_unified_epp...`.

functions/compose-model-replica/function/routing.py
```diff
@@ -203,10 +252,53 @@ def _unified(
...
+) -> dict[str, k8sobjv1alpha1.Object]:
+    """Front the replica's serving pods with an InferencePool + endpoint picker.
+
+    Selects the pods by the serving label they already carry, and routes the
+    replica's path at the pool in place of the Service. The picker scores by
+    prefix cache and queue depth (see _UNIFIED_EPP_CONFIG_TEMPLATE).
+    """
+    assert (
+        replica.metadata is not None and replica.metadata.name is not None
+    )  # set on resources read from the API server
```
negz: Please use the `_name()` (Python) function that does this in our other composition functions. Eventually I want to upstream that into the Python SDK and find-replace all occurrences in Modelplane.

functions/compose-model-replica/function/routing.py
```diff
@@ -447,7 +548,10 @@ def _epp_objects(name: str, provider_config: str, block_size: int) -> dict[str,
             "replicas": 1,
             "selector": {"matchLabels": {"app": epp}},
             "template": {
-                "metadata": {"labels": {"app": epp}},
+                "metadata": {
+                    "labels": {"app": epp},
+                    "annotations": {"modelplane.ai/config-checksum": config_checksum},
```
negz: Perhaps `epp-config-checksum` to make it more obvious what config this tracks.

functions/compose-model-replica/function/routing.py
dennis-upbound: Good call - adopted this. Unified now always fronts the pods with the InferencePool + endpoint picker, one path whether the replica serves from one pod or many. The picker is vestigial for a single pod, but it removes exactly the Service -> pool swap you flagged (which would drop in-flight requests on scale-up), and it let me delete the plain-Service path (`base.serving_resources` + `SERVICE_KEY`) entirely.

functions/compose-model-replica/function/routing.py
dennis-upbound: Done - renamed to `_disaggregated_epp_config_yaml` to match its `_unified_epp_config_yaml` sibling.

functions/compose-model-replica/function/routing.py
dennis-upbound: Done - added a `_name()` helper matching the other composition functions and used it here and in `_disaggregated`, in place of the inline assert/`metadata.name` idiom.

functions/compose-model-replica/function/routing.py
dennis-upbound: Done - the annotation is now `modelplane.ai/epp-config-checksum`.

negz (approved):

PR #343, opened 2026-07-13
Restructure ModelDeployment spec under a template

### LINKED ISSUE #257: Set your own labels on ModelReplicas and ModelEndpoints

A `ModelDeployment` fans out to one `ModelReplica` per scheduled replica, and one `ModelEndpoint` per replica. Modelplane stamps both with exactly three labels and nothing else: `modelplane.ai/deployment`, `modelplane.ai/cluster`, and `modelplane.ai/replica-index` (see [`compose-model-deployment/function/fn.py:373-377`](https://github.com/modelplaneai/modelplane/blob/c37732eaeec3e6f3860b604bba00274468cc057c/functions/compose-model-deployment/function/fn.py#L373-L377) for replicas, and the endpoint labels at [`fn.py:490-494`](https://github.com/modelplaneai/modelplane/blob/c37732eaeec3e6f3860b604bba00274468cc057c/functions/compose-model-deployment/function/fn.py#L490-L494)).

There's no way to put your own labels on a replica or endpoint. The `ModelDeployment`'s own labels aren't copied down, and the API has no replica- or endpoint-level metadata field. You can template labels and annotations onto the *engine pods* (`spec.engines[].members[].template.metadata.labels`), but that doesn't reach the `ModelReplica` or `ModelEndpoint` objects themselves.

This shows up in a couple of places:

- **Organizing and selecting your own resources.** You can't tag a deployment's replicas with something like `tier: prod` or a team or cost label and then list or select on it. `kubectl get modelreplica -l tier=prod` isn't possible.
- **Routing.** `ModelService` selects `ModelEndpoint`s by label, but the only labels that exist to select on are the three Modelplane sets. If you wanted to route to a subset of replicas by some attribute of your own, beyond which deployment and which cluster they're on, you can't, because there's no label to carry that attribute.

We could copy whatever labels are on a `ModelDeployment` down to its replicas and endpoints. This pattern has bitten us in the past in Crossplane though (...but I forget why). The alternative is updating `ModelDeployment` with e.g. `spec.childLabels` or similar so you can set the labels explicitly. 

(Or arguably if we wanted to make a breaking change, move everything but `spec.replicas` under `spec.template` - as in template for a `ModelReplica` - and allow `spec.template.metadata.labels`. That'd be closest to what a Kubernetes `Deployment` does.)

dennis-upbound:
Fixes #257.

Setting your own labels on a deployment's ModelReplicas and ModelEndpoints had no home on the API. Rather than a one-off field, this adopts the Deployment-style shape from the discussion on this PR: everything but `spec.replicas` moves under `spec.template`, with the replica shape in `spec.template.spec` and labels in `spec.template.metadata.labels`, which apply to the composed replicas and their endpoints.

Before:

```yaml
spec:
  replicas: 1
  engines: [...]
```

After:

```yaml
spec:
  replicas: 1
  template:
    metadata:
      labels: {tier: prod, team: search}
    spec:
      engines: [...]
```

The shape leaves room to template annotations and, later, the ModelEndpoint spec, with a path to split replica and model templates if we need it. It's a breaking change, taken now while pre-v0.1 to land a cleaner API before v1beta1.

Labels are validated on the ModelDeployment itself — Kubernetes label syntax, no reserved `modelplane.ai/` prefix, bounded count (`maxProperties: 32`) and length (`maxLength: 63`) — so a malformed label is rejected on the resource the user wrote rather than failing when the composed child is applied. Verified on a dev cluster: `"my label!"` and a 300-char value are rejected on the ModelDeployment.

The diff is large but mostly the XRD re-indent from nesting the existing fields under `spec.template.spec`; the composition change is just reading fields from their new path and stamping `template.metadata.labels` onto the children.

I have:

### DIFF
--- a/apis/modeldeployments/definition.yaml
+++ b/apis/modeldeployments/definition.yaml
@@ -36,20 +36,7 @@ spec:
         properties:
           spec:
             type: object
-            required: [replicas, engines]
-            # PrefillDecode serving runs one Prefill and one Decode engine, each
-            # marking its phase. Validate the pairing here, where both serving
-            # and the engine list are visible. serving is optional and defaults
-            # to Unified, under which no engine may carry a phase.
-            x-kubernetes-validations:
-            - rule: "!has(self.serving) || self.serving.mode != 'PrefillDecode' || size(self.engines) == 2"
-              message: "PrefillDecode serving requires exactly two engines (one Prefill, one Decode)"
-            - rule: "!has(self.serving) || self.serving.mode != 'PrefillDecode' || self.engines.exists_one(e, has(e.phase) && e.phase == 'Prefill')"
-              message: "PrefillDecode serving requires exactly one engine with phase Prefill"
-            - rule: "!has(self.serving) || self.serving.mode != 'PrefillDecode' || self.engines.exists_one(e, has(e.phase) && e.phase == 'Decode')"
-              message: "PrefillDecode serving requires exactly one engine with phase Decode"
-            - rule: "(has(self.serving) && self.serving.mode == 'PrefillDecode') || !self.engines.exists(e, has(e.phase))"
-              message: "engine phase may only be set when serving.mode is PrefillDecode"
+            required: [replicas, template]
             properties:
               replicas:
                 type: integer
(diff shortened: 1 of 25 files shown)


functions/compose-model-deployment/function/fn.py
```diff
@@ -350,6 +350,19 @@ def schedule(self) -> list[scheduling.Candidate]:
 
         return matched
 
+    def _child_labels(self, cluster_info: scheduling.Candidate) -> dict[str, str]:
+        """Labels for a composed ModelReplica or ModelEndpoint.
+
+        The user's childLabels first, then the labels Modelplane manages, so a
+        managed key always wins over a colliding childLabel (the XRD also rejects
```
haarchri: Today that is only enforced by CEL at admission, the function itself is where the invariant lives, and it has no test. Pass `childLabels={"modelplane.ai/cluster": "wrong"}` directly to the function and assert cluster-a wins. That keeps the guarantee if the XRD rule ever gets relaxed or the function is reused elsewhere ?!

apis/modeldeployments/definition.yaml
```diff
@@ -99,6 +101,17 @@ spec:
                       Unified serves prefill and decode on one engine.
                       PrefillDecode splits them across two engines, each marking
                       its phase as Prefill or Decode.
+              childLabels:
```
haarchri: are we fine that this in uncapped ? something like `maxProperties: 32` ?! , what are the costs for the CEL rules here ?!

apis/modeldeployments/definition.yaml
```diff
@@ -99,6 +101,17 @@ spec:
...
+                type: object
+                description: >-
+                  Extra labels to set on the ModelReplicas and ModelEndpoints
+                  this deployment composes, alongside the labels Modelplane
+                  manages. Use them to organize and select your own resources
+                  (kubectl get modelreplica -l tier=prod) or to route to a
+                  subset of endpoints from a ModelService. Keys under the
+                  modelplane.ai/ prefix are reserved.
+                additionalProperties:
```
haarchri: No label syntax validation. `additionalProperties: {type: string}` with only the prefix rule means the apiserver will happily admit:

```
childLabels:
  "my label!": "a-very-long-value…"   # 300 chars
```

The XR is accepted, then the composed ModelReplica gets rejected at apply time, and the user sees a composition error on a child object instead of a validation error on the thing they wrote.

functions/compose-model-deployment/function/fn.py
dennis-upbound: Added a function-level test: pass `childLabels={"modelplane.ai/cluster": "wrong"}` straight to the function and assert the managed value wins, so the guarantee holds even if the CEL rule is ever relaxed or the function is reused.

apis/modeldeployments/definition.yaml
dennis-upbound: Capped it at `maxProperties: 32`. On CEL cost: I also had to add `maxLength: 63` on the value — without a length bound the `all(... matches ...)` rule blew the CRD's CEL cost budget and the CRD wouldn't establish (hit this on a dev cluster). With both caps the estimate is bounded.

apis/modeldeployments/definition.yaml
dennis-upbound: Added CEL for key and value syntax (Kubernetes label rules: an optional DNS-subdomain prefix plus a name up to 63 chars, value up to 63 chars, both `[A-Za-z0-9._-]` edged by alphanumerics), plus `maxLength: 63` on the value. Verified on a dev cluster that `"my label!"` and a 300-char value are now rejected on the ModelDeployment itself instead of failing when the child ModelReplica is applied.

negz:
@dennis-upbound @haarchri What do you think about this API?

```yaml
kind: ModelDeployment
metadata:
  name: qwen-demo
  namespace: ml-team
spec:
  replicas: 1
  template:
   # These implicitly apply to the corresponding endpoints too.
    metadata:
      labels:
        tier: prod
        team: search
    spec:
      engines:
      - name: qwen
        members:
        - role: Standalone
          nodeSelector:
            devices:
            - name: gpu
              count: 1
              selectors:
              - cel: device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("20Gi")) >= 0
          template:
            spec:
              containers:
              - name: engine
                image: vllm/vllm-openai:v0.23.0
                args: ["--model=Qwen/Qwen2.5-0.5B-Instruct"]
```

Pros:
- It leaves room to template other metadata we may need in future, e.g. annotations
- It creates a fairly clean path to break out to `replicaTemplate` and `modelTemplate` if we ever need to

Cons
- `template` appears twice at two different levels

It's also a breaking API change, but I don't think we should worry about that at all right now. It's important we try hard to avoid breaking API changes, but we should make them if we need them while we're pre-v0.1 as it'll hopefully mean we land on a more elegant API when we get to v1beta1.

haarchri:
+1 sounds a bit more on the core Deployment then - like this more then childLabels - i added no comments as your Issue suggested childLabels ...

dennis-upbound:
Sold. It's the more Kubernetes-native shape, and pre-v0.1 is the right time to take the breaking change to land somewhere cleaner for v1beta1.

I'll do it here: everything but `replicas` moves under `spec.template.spec`, labels go on `spec.template.metadata.labels` (replacing `childLabels`, and applying to the endpoints too), and the label-syntax/reserved-prefix validation moves with them. I'll rebase #344 on top so `tolerations` lands at `spec.template.spec.tolerations` rather than the top level.

dennis-upbound:
Done — pushed. `spec.template.spec` holds the replica shape, `spec.template.metadata.labels` the labels (applied to replicas and endpoints), everything validated on the ModelDeployment itself. The diff's large but it's mostly the XRD re-indent from nesting the existing fields; the composition change is small. Retitled the PR. I'll rebase #344 so `tolerations` lands at `spec.template.spec.tolerations`.

docs/content/models/model-deployment.md
```diff
@@ -15,11 +15,11 @@ its flags, and Modelplane shapes a serving topology around it. The engine flags
 you write carry parallelism, quantization, and KV transfer, never injected by
 Modelplane.
 
-A deployment's `spec.engines` describes its topology through two choices:
+A deployment's replica shape lives under `spec.template` (mirroring a Kubernetes Deployment): the replica count is `spec.replicas`, labels for the composed ModelReplicas and ModelEndpoints go on `spec.template.metadata.labels`, and everything else lives in `spec.template.spec`. Its `spec.template.spec.engines` array describes the topology through two choices:
```
negz: Wrap at 80 chars per `CONTRIBUTING.md` please. Here and elsewhere.

negz (approved):
One nit and needs a rebase but otherwise LGTM.

PR #344, opened 2026-07-13
Schedule around taints on InferenceClusters

### LINKED ISSUE #16: Support taints on InferenceEnvironments for scheduling control and drain

Platform teams need lifecycle control over InferenceEnvironments. Today there's no way to take an IE out of service gracefully -- no way to stop new models from being scheduled to it, no way to drain existing models off it, and no way for a high-priority deployment to override those restrictions.

A platform team decommissioning a GPU cluster, performing maintenance, or shifting capacity between regions has to coordinate manually: find all ModelDeployments targeting the IE, delete or reschedule them, then delete the IE. There's no declarative way to express "this environment is going away" and let the system handle the rest.

InferenceEnvironments should support taints, following the Kubernetes node taint model. The deploy function would check taints during scheduling, and ModelDeployments could declare tolerations.

```yaml
kind: InferenceEnvironment
metadata:
  name: gpu-us-east
spec:
  taints:
  - key: modelplane.ai/maintenance
    value: decommissioning
    effect: NoSchedule
```

```yaml
kind: ModelDeployment
metadata:
  name: critical-model
  namespace: ml-team
spec:
  modelRef:
    kind: ClusterModel
    name: llama-70b
  environments: 1
  tolerations:
  - key: modelplane.ai/maintenance
    operator: Exists
    effect: NoSchedule
```

Two effects, matching Kubernetes semantics:

- **`NoSchedule`**: The deploy function skips this IE for new placements. Existing placements stay. Use case: maintenance window, no new models but current ones keep running.
- **`NoExecute`**: The deploy function removes existing placements targeting this IE and reschedules to other compatible environments (respecting `spec.environments` count). Use case: decommissioning, drain all workloads before teardown.

A `NoExecute` taint is the declarative equivalent of `kubectl drain`. The deploy function would remove placements from the tainted IE and create replacements on other compatible environments. If no replacement is available, the ModelDeployment's `spec.environments` count won't be met -- surfaced via status conditions.

Tolerations on ModelDeployments let the ML team express "this deployment can stay on a tainted environment." A critical production model could tolerate maintenance taints to avoid unnecessary disruption. The platform team and ML team negotiate via taints and tolerations without direct coordination.

dennis-upbound:
Fixes #16.

Taking an InferenceCluster out of service meant coordinating by hand: find every ModelDeployment placed on it, reschedule or delete those, then delete the cluster. There was no declarative way for a platform team to say "stop scheduling here" or "drain what's here," and no way for an ML team to keep a critical deployment in place through it.

This adds taints to `InferenceCluster` (`spec.taints`) and tolerations to `ModelDeployment` (`spec.template.spec.tolerations`), following the Kubernetes model. When placing new replicas the scheduler skips a cluster carrying an untolerated taint (NoSchedule and NoExecute alike). A `NoExecute` taint additionally drains the replicas already there: the retain phase drops them so the fill phase reschedules them onto tolerated clusters (delete-plus-create, like a node drain). `NoSchedule` leaves existing replicas in place. A drain with nowhere to reschedule leaves the deployment below `spec.replicas`, reported by its existing `ReplicasScheduled` condition.

```yaml
kind: InferenceCluster
metadata:
  name: gpu-us-east
spec:
  taints:
  - key: modelplane.ai/maintenance
    effect: NoSchedule    # NoExecute to also drain existing replicas
```

```yaml
kind: ModelDeployment
spec:
  replicas: 1
  template:
    spec:
      tolerations:
      - key: modelplane.ai/maintenance
        operator: Exists
      engines: [...]
```

The issue framed this on an "InferenceEnvironment" with `spec.environments`; that kind is `InferenceCluster` today and a ModelDeployment scales by `spec.replicas` across clusters it selects by label, so the fields here map the idea onto the current API. Worth a look that the mapping is what you had in mind.

This revision keeps to the taint/toleration scheduling only. An earlier version also published `status.scheduledReplicas` (with a REPLICAS printer column) and a `Drained` condition on InferenceCluster. Per review, that condition was really just `scheduledReplicas == 0`, and Kubernetes' own `Node` neither counts scheduled pods nor carries such a condition — you list the pods on it. So both the field and the condition are dropped: a platform team confirms a drain by listing the ModelReplicas placed on the cluster, matching the Node pattern.

Rebased onto `main` now that the `spec.template` restructure (#343) has merged, so tolerations live at `spec.template.spec.tolerations`.

Validated with `crossplane render`: a NoSchedule-tainted cluster is skipped for new placement, and a NoExecute taint drains an existing replica and reschedules it onto another cluster. On a dev control plane the apiserver's CEL rejects keyless `Equal` tolerations, `Exists` tolerations carrying a value, keyless taints, and unknown effects. `nix flake check` passes (the taint/toleration scheduling tests and the docs-manifest validation run green locally).

I have:

### DIFF
--- a/apis/inferenceclusters/definition.yaml
+++ b/apis/inferenceclusters/definition.yaml
@@ -172,6 +172,30 @@ spec:
                           mk8s cluster Kubernetes version. Defaults to a
                           version where Dynamic Resource Allocation (DRA)
                           is generally available.
+              taints:
+                type: array
+                description: >-
+                  Taints that repel ModelDeployments from this cluster unless
+                  they carry a matching toleration, following the Kubernetes
+                  taint model. NoSchedule keeps new replicas off the cluster
+                  while leaving existing ones in place; NoExecute additionally
+                  drains the replicas already here, which the scheduler
+                  reschedules onto other tolerated clusters (the declarative
+                  equivalent of draining a node before maintenance).
+                x-kubernetes-list-type: map
+                x-kubernetes-list-map-keys: [key, effect]
+                items:
+                  type: object
+                  required: [key, effect]
+                  properties:
+                    key:
+                      type: string
+                      minLength: 1
+                    value:
+                      type: string
+                    effect:
+                      type: string
+                      enum: [NoSchedule, NoExecute]
               nodePools:
                 type: array
                 description: >-
--- a/apis/modeldeployments/definition.yaml
+++ b/apis/modeldeployments/definition.yaml
@@ -143,6 +143,43 @@ spec:
                               Unified serves prefill and decode on one engine.
                               PrefillDecode splits them across two engines, each marking
                               its phase as Prefill or Decode.
+                      tolerations:
+                        type: array
+                        description: >-
+                          Tolerations that let this deployment's replicas
+                          schedule onto, and stay on, InferenceClusters carrying
+                          matching taints, following the Kubernetes toleration
+                          model. A replica tolerating a cluster's NoSchedule
+                          taint can be placed there; one tolerating a NoExecute
+                          taint is not drained when that taint is applied.
+                        x-kubernetes-list-type: atomic
+                        items:
+                          type: object
+                          x-kubernetes-validations:
+                          - rule: "self.operator != 'Equal' || (has(self.key) && self.key != '')"
+                            message: "an Equal toleration must set a key"
+                          - rule: "self.operator != 'Exists' || !has(self.value)"
+                            message: "an Exists toleration must not set a value"
+                          properties:
+                            key:
+                              type: string
+                              description: >-
+                                Taint key to match. Empty with operator Exists
+                                tolerates every taint.
+                            operator:
+                              type: string
+                              enum: [Exists, Equal]
+                              default: Equal
+                              description: >-
+                                Exists matches any taint with the key (ignoring
+                                value); Equal matches key and value.
+                            value:
+                              type: string
+                            effect:
+                              type: string
+                              enum: [NoSchedule, NoExecute]
+                              description: >-
+                                Taint effect to match. Empty matches all effects.
                       engines:
                         type: array
                         description: >-
--- a/functions/compose-model-deployment/function/scheduling.py
+++ b/functions/compose-model-deployment/function/scheduling.py
@@ -318,6 +318,49 @@ def _cluster_ready(cluster: icv1alpha1.InferenceCluster) -> bool:
     return any(c.type == "Ready" and c.status == "True" for c in cluster.status.conditions or [])
 
 
+_EFFECT_NO_SCHEDULE = "NoSchedule"
+_EFFECT_NO_EXECUTE = "NoExecute"
+
+
+def _tolerates(toleration: mdv1alpha1.Toleration, taint: icv1alpha1.Taint) -> bool:
+    """Whether a toleration matches a taint, following Kubernetes semantics.
+
+    An empty toleration effect matches any effect; otherwise the effects must be
+    equal. Operator Exists matches any taint with the key (an empty key matches
+    every taint); Equal (the default) matches on key and value.
+    """
+    if toleration.effect and toleration.effect != taint.effect:
+        return False
+    if (toleration.operator or "Equal") == "Exists":
+        return not toleration.key or toleration.key == taint.key
+    return toleration.key == taint.key and (toleration.value or "") == (taint.value or "")
+
+
+def _has_untolerated_taint(
+    cluster: icv1alpha1.InferenceCluster,
+    tolerations: list[mdv1alpha1.Toleration],
+    effects: tuple[str, ...],
+) -> bool:
+    """Whether the cluster carries a taint with one of `effects` that no toleration matches."""
+    return any(
+        taint.effect in effects and not any(_tolerates(t, taint) for t in tolerations)
+        for taint in cluster.spec.taints or []
+    )
+
+
+def _repels_new(cluster: icv1alpha1.InferenceCluster, tolerations: list[mdv1alpha1.Toleration]) -> bool:
+    """Whether an untolerated taint keeps NEW replicas off this cluster.
+
+    Both NoSchedule and NoExecute block new placements.
+    """
+    return _has_untolerated_taint(cluster, tolerations, (_EFFECT_NO_SCHEDULE, _EFFECT_NO_EXECUTE))
+
+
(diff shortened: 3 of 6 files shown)


apis/inferenceclusters/definition.yaml
```diff
@@ -153,6 +153,28 @@ spec:
...
+                type: array
+                description: >-
+                  Taints that repel ModelDeployments from this cluster unless
+                  they carry a matching toleration, following the Kubernetes
+                  taint model. NoSchedule keeps new replicas off the cluster
+                  while leaving existing ones in place; NoExecute additionally
+                  drains the replicas already here, which the scheduler
+                  reschedules onto other tolerated clusters (the declarative
+                  equivalent of draining a node before maintenance).
+                items:
```
haarchri, line 187: should be `x-kubernetes-list-type: map` with `listMapKeys: [key, effect]`, like node taints upstream, right now you can add the same taint twice, and SSA will merge these atomically instead of by key, which will bite when a controller and a human both touch the list

apis/modeldeployments/definition.yaml
```diff
@@ -99,6 +99,41 @@ spec:
                       Unified serves prefill and decode on one engine.
                       PrefillDecode splits them across two engines, each marking
                       its phase as Prefill or Decode.
+              tolerations:
```
haarchri: spec.tolerations should be` x-kubernetes-list-type: atomic`

apis/inferenceclusters/definition.yaml
haarchri, line 323: The workflow in #16 is "platform team takes a cluster out of service." After this PR they can express it, but they still can not tell when it is done, they would  have to walk every `ModelDeployment` to find out whether anything is still placed on the cluster. 

A `status.replicas`count or a Drained condition on InferenceCluster is what actually closes the loop on the original use case. ?!

haarchri (commented):
can we add a bit more testing around the following:

- A toleration with effect: NoSchedule against a NoExecute taint, must not tolerate. This is the case where an operator thinks they're covered and gets drained anyway.
- A cluster with two taints where only one is tolerated, must still repel. _has_untolerated_taint looks correct, but nothing pins it.
- operator: Equal with a matching value, and with a mismatched value. _MAINT carries value="on" but every toleration test uses Exists, so the Equal path is never exercised end-to-end.
- Empty-key Exists tolerating everything

apis/inferenceclusters/definition.yaml
dennis-upbound: Done — `x-kubernetes-list-type: map` with `x-kubernetes-list-map-keys: [key, effect]`, matching upstream node taints.

apis/modeldeployments/definition.yaml
dennis-upbound: Done — `x-kubernetes-list-type: atomic` on `spec.tolerations`.

apis/inferenceclusters/definition.yaml
dennis-upbound: Agreed — this is what actually closes the #16 loop; a taint you can't watch drain is half the feature. I'd rather do it as a fast follow than grow this PR: add `status.scheduledReplicas` on InferenceCluster plus a `Drained` condition when it reaches zero, counting the ModelReplicas already labelled for the cluster (the same signal the scheduler reads). I'll open an issue and link it here.

dennis-upbound:
Added all four cases to `TestScheduleTaints`:

- a NoSchedule toleration against a NoExecute taint — matches the key but not the effect, so it doesn't tolerate and the replica is still drained;
- two taints on a cluster with only one tolerated — still repels new placement;
- `Equal` with a matching value (places) and a mismatched value (repels), so the Equal path is exercised end to end;
- an empty-key `Exists` tolerating every taint on the cluster.

apis/inferenceclusters/definition.yaml
dennis-upbound: Update: rather than a separate follow-up, we're landing this in this PR so #344 fully closes #16. Adding `status.scheduledReplicas` and a `Drained` condition on InferenceCluster (plus a REPLICAS printer column), counted from the ModelReplicas the deletion guard already observes for the cluster. Pushing shortly.

dennis-upbound: Pushed — `status.scheduledReplicas` + a `Drained` condition (and a REPLICAS printer column) are in this PR now, as a second commit. Verified with `crossplane render`: the count tracks the observed replicas and Drained flips at zero. It's only reported once Crossplane has resolved the guard's requirement, so a reconcile that hasn't observed the replicas yet doesn't momentarily claim zero.

functions/compose-inference-cluster/function/fn.py
```diff
@@ -184,7 +191,28 @@ def compose_replica_guard(self) -> None:
...
+        resolved = "model-replicas" in self.req.required_resources
         replicas = request.get_required_resources(self.req, "model-replicas")
+        # Once the requirement resolves, report how many ModelReplicas are on the
+        # cluster (and a Drained condition when none are), so a platform team
+        # taking a cluster out of service with a NoExecute taint can tell the
+        # drain has finished without walking every ModelDeployment. Until then we
+        # leave the last reported count rather than claim zero.
+        if resolved:
+            self.scheduled_replicas = len(replicas)
+            drained = self.scheduled_replicas == 0
```
negz: To me "drained" implies someone has actively drained the cluster, but I think this'd be true for a cluster that just happens to have no replicas right? e.g. Simply because nothing's been scheduled there yet.

We could address this by renaming the condition to e.g. `ClusterInUse`. I lean toward dropping the condition though, given it's really just telling you whether `status.scheduledReplicas == 0`, and thus is easily

apis/inferenceclusters/definition.yaml
negz: Interestingly the `Node` type in Kubernetes doesn't have a condition for this, or even a status field counting scheduled replicas. AFAICT the pattern there is indeed to list all pods and see which ones are on the node.

functions/compose-inference-cluster/function/fn.py
dennis-upbound: Agreed - dropped the `Drained` condition entirely. As you said, it was really just `scheduledReplicas == 0`, so it wasn't earning its keep.

apis/inferenceclusters/definition.yaml
dennis-upbound: Followed the Node pattern the rest of the way: dropped `status.scheduledReplicas` (and its REPLICAS printer column) too. A platform team confirms a drain by listing the ModelReplicas placed on the cluster, the same way you'd list pods on a Node. The PR is now taint/toleration scheduling only.

negz (approved):

Issue #334, opened 2026-07-09
Reduce control-plane CRD and controller footprint (MRAP + optional providers)


vfarcic:
### What problem are you facing?

Installing Modelplane hard-depends on seven Crossplane cloud provider families (provider-aws-eks/ec2/iam/efs, provider-gcp-container/compute/cloudplatform) and activates each family's full CRD set, though the compositions instantiate only about twenty managed-resource kinds. That inflates CRD count and controller memory, the symptom reported in #323 (idle control plane at ~9GB). Two things make it worse: a BYOC-only operator who never provisions a cluster still installs the entire AWS and GCP CRD and controller surface, because the providers are hard dependencies of the Configuration; and nothing narrows the activated CRD set below the family level.

### How could Modelplane help solve your problem?

Adopt Crossplane v2 SafeStart with a ManagedResourceActivationPolicy to activate only the managed-resource kinds the compositions use, leaving the rest inactive until needed, which cuts CRD count and controller footprint without removing capability. Separately, make the cloud provisioning provider families optional, via separate dependencies or a bring-your-own install profile, so operators who only register existing clusters do not install provider-aws or provider-gcp at all.

Related: #323.

haarchri:
Thanks for raising this. 
I think the upstream work proposed in https://github.com/crossplane/crossplane/issues/7573 is a promising way to address this.

If that proposal is accepted and implemented in Crossplane, Modelplane could generate a ManagedResourceActivationPolicy (MRAP) for each InferenceCluster variant. Each MRAP would declare only the managed resource kinds required for that specific infrastructure (for example, EKS, GKE, or a BYOC deployment).

With Crossplane's SafeStart and MRAP support, this would mean:

- Crossplane would activate only the provider referenced by the MRAP (new: scale from 0 -> 1)
- The runtime footprint would scale with the infrastructure being used instead of the full set of providers declared by Modelplane.
- Provider controllers would start only for the managed resource kinds that are actually required. (already possible)
- BYOC-only deployments would not activate any cloud provider controllers.

This would largely address the runtime memory concerns described in #323 without requiring us to maintain separate installation profiles or manually curate provider activation. I wouldd be happy to implement per-InferenceCluster MRAP generation in Modelplane once the upstream functionality is available.

negz:
Adding to v0.2 as a stretch, to generate an MRAP (or MRAPs) for Modelplane.

[...]

Issue #323, opened 2026-07-06
Getting started guide requires ~9GB of memory


negz:
### What problem are you facing?

I followed the [getting started guide](https://github.com/modelplaneai/modelplane/blob/main/docs/content/getting-started/) up through "Set up the InferenceGateway". At idle, the single kind node container was using:

```
CONTAINER                  CPU %     MEM USAGE / LIMIT    MEM %
modelplane-control-plane   10-16%    9.3GiB / 19.4GiB      48%
```

That's a lot of memory for a getting started guide. I suspect it won't fit on a lot of potential users' development laptops - e.g. folks with 16GB MacBooks who allocate 8GB to the Docker VM.

Broken down by container (`crictl stats` inside the node):

Crossplane providers:

| Memory | Provider |
|---|---|
| 426 MB | `provider-aws-ec2` |
| 341 MB | `provider-gcp-compute` |
| 301 MB | `provider-aws-iam` |
| 286 MB | `provider-aws-efs` |
| 285 MB | `provider-aws-eks` |
| 255 MB | `provider-family-aws` |
| 239 MB | `provider-gcp-cloudplatform` |
| 222 MB | `provider-gcp-container` |
| 220 MB | `provider-family-gcp` |
| 158 MB | `provider-helm` |
| 149 MB | `provider-kubernetes` |
| **2,882 MB** | **Subtotal** |

Modelplane composition functions:

| Memory | Function |
|---|---|
| 61 MB | `inference-gateway` |
| 44 MB | `eks-cluster` |
| 44 MB | `model-deployment` |
| 38 MB | `gke-cluster` |
| 35 MB | `serving-stack` |
| 34 MB | `model-replica` |
| 34 MB | `inference-cluster` |
| 33 MB | `model-cache` |
| 33 MB | `model-endpoint` |
| 33 MB | `inference-class` |
| 33 MB | `model-service` |
| 32 MB | `usages` |
| **454 MB** | **Subtotal** |

Everything else:

| Memory | Component |
|---|---|
| 3,028 MB | `kube-apiserver` |
| 554 MB | Crossplane core (`crossplane`) |
| 283 MB | `etcd` |
| 164 MB | `kube-controller-manager` |
| 141 MB | MetalLB controller |
| 31 MB | `kube-scheduler` |
| 27 MB | Traefik |
| 26 MB | `kindnet-cni` |
| 23 MB | `kube-proxy` |
| 22 MB | `crossplane-rbac-manager` |
[...]

PR #351, opened 2026-07-16
Activate only the managed resources Modelplane composes

negz:
Towards #334.

Installing Modelplane hard-depends on seven cloud provider families whose CRDs total around 900 on the control plane, but the compositions only ever compose about 40 managed resource kinds. Crossplane's default activation policy turns them all on (`"*"`), so a fresh install pays the API-server and controller memory for ~900 CRDs it never uses, the idle footprint reported in #323.

Crossplane v2 converts each provider CRD into an inactive ManagedResourceDefinition and only creates the CRD once a ManagedResourceActivationPolicy names it. This ships such a policy in the Configuration package (`apis/mrap.yaml`, which `crossplane project build` collects like an XRD or Composition), naming exactly the kinds the compositions compose, and installs Crossplane with an empty default activation (`provider.defaultActivations=[]`) so that policy is the only thing activating managed resources.

Measured on a kind install of v0.1.0-dev.1784158975.g3e46dfb: 107 CRDs instead of 912, and 7.2 GiB total instead of 12.3 GiB, with kube-apiserver at 1.3 GiB instead of 4.4 GiB. All 15 providers stayed healthy.

The activation list is hand-maintained, so a managed resource kind added to a composition without a matching entry here silently fails to activate, because its CRD won't exist. I derived the current 43 entries from the functions' model imports and cross-checked each against live MRDs, but there's no CI guard against future drift yet.

I have:

- [ ] ~Added or updated tests covering any composition function changes.~ (no composition function changes)


docs/manifests/getting-started/prerequisites.yaml
```diff
@@ -5,6 +5,76 @@
 # routing resources. They cannot be self-composed because Crossplane
 # needs the permissions before it can compose anything.
 ---
+# Crossplane converts every provider CRD into an inactive
```
negz: You can actually package an MRAP in a Configuration, but I'm not sure the `crossplane` CLI will let us do that using a project...

bassam (approved):

docs/manifests/getting-started/prerequisites.yaml
negz: (It does, updating now.)

apis/mrap.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - launchtemplates.ec2.aws.m.upbound.io
+  - natgateways.ec2.aws.m.upbound.io
+  - routes.ec2.aws.m.upbound.io
+  - routetables.ec2.aws.m.upbound.io
+  - routetableassociations.ec2.aws.m.upbound.io
+  - securitygroups.ec2.aws.m.upbound.io
+  - securitygroupegressrules.ec2.aws.m.upbound.io
+  - securitygroupingressrules.ec2.aws.m.upbound.io
+  - subnets.ec2.aws.m.upbound.io
+  - vpcs.ec2.aws.m.upbound.io
```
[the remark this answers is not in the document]
negz: Indentation looks fine to me.

PR #361, opened 2026-07-21
ci: skip credential-dependent steps in fork runs

haarchri:

Fork PRs do not have access to repository secrets, causing the registry login, push, and Algolia crawl steps to fail or run unnecessarily.

Split Build and Push into separate Build and Push steps, move Build before auth so CI fails fast on build errors before touching credentials
Gate Login to registry and Push on `secrets.UP_ROBOT_ID != ''` so forks build but skip the push
Gate the docsearch crawl job on `secrets.ALGOLIA_ADMIN_KEY != ''` so scheduled runs in forks don't spin up a runner at all

Fixes #

I have: 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,12 @@ jobs:
     needs: [check]
     runs-on: ubuntu-24.04
 
+    # Secrets aren't available in if expressions, so expose the (non-secret)
+    # robot ID as an env var that the push steps can gate on. Forks don't have
+    # the secret, so their runs build the package but skip pushing it.
+    env:
+      UP_ROBOT_ID: ${{ secrets.UP_ROBOT_ID }}
+
     steps:
       # Building and pushing the package needs more disk than the runner has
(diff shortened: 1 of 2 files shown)


negz (approved):

negz:
Seems like the CI check isn't running on this PR. 🤔

negz:
@haarchri I think `check` isn't running because your update is invalid. iirc You can't use `secret` in if blocks. I've worked around this in the past by exposing the username (not password) as an env var, which we can gate on.

https://github.com/crossplane/crossplane/blob/5cee1b741af3c62e81e4284cc424e9790af7fead/.github/workflows/ci.yml#L14-L19
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#context-availability

negz (changes requested):
(Can't merge as-is per last comment.)

haarchri:
remembers my why i like github-ci workflows ...

negz (approved):

Issue #210, opened 2026-06-19
Decompose ModelCache into a per-cluster ModelCacheHydration child


negz:
### What problem are you facing?

`compose-model-cache` does almost everything per matched `InferenceCluster`, inlined and fanned out by hand in one function. A single reconcile composes the PVC, the hydration `Job`, and the propagated token `Secret` for every cluster; derives each cluster's phase (Pending/Hydrating/Ready/Failed) from observed PVC/Job status; latches per-cluster readiness; drops the `Job` (and the token `Secret`) once a cluster is `Ready`; and rolls all of that into `status.clusters[]`. Nearly every method threads a `cluster_name` or a `per_cluster_phase` list. The function is large, and the per-cluster lifecycle is tangled into the fan-out loop rather than living in one place.

It's also notable that `ModelCache` is the one fan-out in the system that *doesn't* follow the `ModelDeployment` → `ModelReplica` pattern. `ModelDeployment` stamps one child per cluster and rolls up child readiness; `ModelCache` open-codes the equivalent inline.

### How could Modelplane help solve your problem?

Decompose `ModelCache` the way `ModelDeployment` is decomposed: introduce a namespaced, per-cluster `ModelCacheHydration` (name TBD), pinned to one cluster via `spec.clusterName`.

- **`ModelCache` becomes a fan-out + roll-up.** It resolves the matched `InferenceCluster`s, stamps one `ModelCacheHydration` per cluster carrying the source (`huggingFace` repo/revision/sizeGiB), the `clusterName`, and the *reference* to the auth `Secret` — name, key, and the cache's namespace, never the value. It rolls child readiness up into `status.clusters[]` / `status.summary`. No phase machine, no remote-`Object` wrapping, no `Job`/`Secret` lifecycle, no token resolution.
- **`ModelCacheHydration` owns one cluster's hydration.** It requires its `InferenceCluster` (for the providerConfig and effective storage class) and resolves the referenced auth `Secret` itself, then composes the PVC, the propagated token `Secret`, and the `Job`, runs the phase machine, and handles the drop-`Job`/drop-`Secret`-after-`Ready` lifecycle.

The auth `Secret` reference is copied down, not its value — the child resolves it, so token values never live in a CR spec, the same trust boundary as today.

PR #362, opened 2026-07-21
Design: decompose ModelCache into a per-cluster ModelCacheHydration

dennis-upbound:

A design doc, not code yet — opening as a draft for direction before implementation.

`ModelCache` runs every matched cluster's hydration lifecycle inline in one fan-out function (PVC + Job + token Secret + a four-phase machine + drop-after-Ready + token resolution, threading `cluster_name` through every method), and it's the one fan-out that doesn't follow `ModelDeployment → ModelReplica`. Separately (#186), a cache's `clusterSelector` and a deployment's placement are decided independently, so a replica can land on a cluster the cache never staged to and fail to mount at runtime, with nothing visible at apply time.

This proposes handling both together: a per-cluster `ModelCacheHydration` child (the cache analogue of `ModelReplica`) that owns one cluster's hydration, and making the `ModelCache` footprint follow where its referencing replicas are placed rather than a hand-maintained selector. Covers #210 and #186. Includes an architecture diagram and a lifecycle flow diagram.

Where I'd most like judgment (all in the doc): the unified vs phased delivery, keeping `modelCacheRef` vs deriving the model from the deployment, the fleet-reconciler-vs-reuse-`ModelCache` ownership question, and the child kind name (`ModelCacheHydration` vs `ModelCacheReplica`).

`design/modelcache-hydration.md` only; no code.

I have:

- [ ] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~ (design doc only, no code paths)
- [ ] ~Added or updated tests covering any composition function changes.~ (design doc only)

### DIFF
--- a/design/modelcache-hydration.md
+++ b/design/modelcache-hydration.md
@@ -185,9 +185,29 @@ Pending without a reason.
 
 ### Modelplane injects env values, never engine flags
 
-The engine command stays the ML team's. Where a value depends on the cluster,
-Modelplane injects an env var the command references, the way it injects
-`MODELPLANE_LEADER_ADDRESS` today. It never writes an engine flag.
+The engine command stays the ML team's, and Modelplane never writes an engine flag. In
+the common case it injects nothing. The cache mounts at the same path on every footprint
+cluster, so the ML team writes that path directly.
+
+```yaml
+engines:
+- name: kimi-k2
+  members:
+  - role: Standalone
+    template:
+      spec:
+        containers:
+        - name: engine
+          image: vllm/vllm-openai:v0.11.0
+          args:
+          - --model=/mnt/models      # the cache mount, uniform across the footprint
+```
+
+Injection covers the case where a value varies from one cluster to the next. Modelplane
+sets it as an env var and the ML team references it with Kubernetes `$(VAR)` expansion,
+the way `MODELPLANE_LEADER_ADDRESS` supplies a multi-node leader address today. A
+loader-plugin cache uses the same mechanism when its client needs a cluster-specific
+server address.
 
 A cache presents its weights to the engine one of two ways.
 
@@ -202,8 +222,8 @@ plugs into vLLM and SGLang. Each needs the engine's own `--load-format` and a
 loader-capable image. The ML team writes that, because they chose the cache, and
 Modelplane wires the env and the cluster-side pieces.
 
-Either way the flag that names the model belongs to the engine, so Modelplane stays out
-of it and injects only env values. Placing only where the cache is pre-warmed
+Either way the flag that names the model belongs to the engine, and Modelplane stays out
+of it. Placing only where the cache is pre-warmed
 makes whatever the ML team wrote valid wherever the replica runs. How the bytes reach
 the cluster, whether a shared filesystem, peer-to-peer distribution, or GPU-to-GPU
 streaming, is the platform team's concern and orthogonal to the command. The catalog of
--- a/design/modelcache-hydration.md
+++ b/design/modelcache-hydration.md
@@ -98,8 +98,9 @@ flowchart TD
 ## The per-cluster child
 
 `ModelCacheHydration` is a namespaced composite pinned to one cluster, the cache
-analogue of `ModelReplica`. Its spec carries what one cluster's hydration needs
-and nothing about fan-out:
+analogue of `ModelReplica`. The name states the lifecycle the child owns, where
+`ModelCacheReplica` would imply a copy of the parent. Its spec carries what one
+cluster's hydration needs and nothing about fan-out:
 
 - **`clusterName`**, the cluster it stages onto. Pinned at creation; the parent
   re-places only if the cluster disappears.
@@ -158,7 +159,9 @@ absorbs it once. An author onto a warmed cluster then waits only for the load,
 roughly fifteen minutes.
 
 Because the footprint is the static selector, `ModelCache` does not watch replica
-placement and does not recompose when replicas move.
+placement and does not recompose when replicas move. A `ModelCache` with no
+`clusterSelector` stages nowhere, so Modelplane rejects it at apply time rather
+than admit a cache no deployment can place against.
 
 ### Two modes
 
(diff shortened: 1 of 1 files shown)


design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+## What the parent composes
+
+`compose-model-cache` becomes fan-out plus roll-up:
+
+- **Resolve** the referencing `ModelReplica`s (the placed set) and, for pre-warm,
+  the `clusterSelector`.
+- **Stamp** one child per cluster in `placed ∪ preWarm`, named
+  `child_name("modelcache", ns, cache-name, cluster)`, copying down `huggingFace`,
```
haarchri, line 280: do we have here a problem with 63 char limit ?

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+has capacity; if the cache didn't stage there, the PVC is missing and the pod
+fails to mount at runtime, with nothing visible at apply time. Staging everywhere
+is the only safe workaround, and it wastes storage and pushes the download token
+and private weights onto clusters running none of the team's deployments.
+
+Instead, derive the footprint from placement. Each reconcile, `ModelCache`:
+
+1. requires every `ModelReplica` in the fleet (the scheduler already does this for
+   capacity accounting);
+2. keeps those that reference it (`spec.modelCacheRef.name` and namespace);
```
negz: We could make this more efficient by:

1. Have ModelDeployment label replicas with the cache they use (`modelplane.io/cache: foo`) when it creates them
2. Have ModelCache require only replicas with that label.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+Two changes:
+
+1. **Split out `ModelCacheHydration`**, a per-cluster child that owns one
+   cluster's hydration. Pinned by `spec.clusterName`, it resolves its
+   `InferenceCluster` and auth `Secret`, composes the PVC, token `Secret`, and
+   `Job`, runs the phase machine, and drops the `Job`/`Secret` once Ready.
+2. **Derive the footprint from placement.** `ModelCache` stages onto the
+   clusters where its referencing replicas are placed, not onto a hand-maintained
+   `clusterSelector`. A selector, if set, is an opt-in pre-warm set, not the whole
+   footprint.
```
negz: I suggest we rename or move the selector to make it clear that it doesn't actually limit where the cache is placed. e.g. `spec.preWarm.clusterSelector` or similar.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+- When the last referencing replica leaves X (and X isn't pre-warmed), X drops
+  out, the parent stops composing that child, and the child and its PVC are
+  reclaimed.
+
+No hydration is retracted out from under a live replica, and an unreferenced cache
+reclaims itself per cluster with no manual cleanup.
+
+### Hydrating before ready
+
+A replica can be scheduled onto a cluster new to the model before its PVC exists.
```
haarchri: If hydration on that cluster fails permanently, expired HF token, storage exhausted, bad revision, the replica appears to be gated indefinitely.
This needs a backoff plus a Failed condition that propagates up to the `ModelDeployment`, otherwise a credential problem on one cluster surfaces as a deployment that simply never becomes ready?!

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Alternatives considered
+
+### Keep `clusterSelector` as the whole footprint
+
+The status quo, and the divergence #186 is about: the selector and placement
+drift, and the failure is a runtime mount error with nothing at apply time.
+Deriving the footprint from placement removes the class of bug by construction.
+The selector survives only for the case placement can't serve, pre-warming ahead
+of a deployment.
```
negz: This isn't exactly the status quo since https://github.com/modelplaneai/modelplane/pull/189.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+
+- a structured model source on the deployment (`spec.template.spec.model`), and
+- hydration keyed by model identity (a content hash) rather than cache name, so
+  two deployments of one model share a copy per cluster, which also changes the
+  serving mount contract.
+
+That is a larger, user-facing change with its own migration. This proposal keeps
+`modelCacheRef` and solves the divergence, waste, sprawl, and lifecycle problems
+without it. Identity-keyed sharing can be added on top later, since the child
+already carries a source independent of how it was requested.
```
negz: I'm not advocating for this alternative at all, but: this is a non-issue at this stage in our lifecycle.

If there is a better design, we should adopt it now. Even if it requires breaking API changes and "migrations". No one is running Modelplane v0.2 - there's no one to break, no one to migrate.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Keep `clusterSelector` as the whole footprint
+
+The status quo, and the divergence #186 is about: the selector and placement
+drift, and the failure is a runtime mount error with nothing at apply time.
+Deriving the footprint from placement removes the class of bug by construction.
+The selector survives only for the case placement can't serve, pre-warming ahead
+of a deployment.
+
+### Drop `modelCacheRef`; derive the model from the deployment
```
negz, line 332: What is the purpose of ModelCache? Is it:

1. Get weights on a cluster before any model replica runs there
2. Get weights on a cluster the first time a model replica runs there
3. A bit of both?

Apart from storage, 1. feels optimal to me. I'm basing this on my experience trying to run larger models like Kimi when testing. It was something like 45 minutes to hydrate the cache PVC, then another 15+ minutes to load from the cache PVC.

With 1., the 45 mins is paid in advance. So the user experience for an MD author is "my model is ready in ~15 mins".

With 2., the 45 mins is paid the first time the model is deployed on the cluster. So for that MD author the user experience is "my model is ready in an hour", because we need to do two copies. We hydrate the cache, then load the model to the replica(s). Subsequent deploys are faster.

I think this alternative would be 2., whereas the actual proposal is 3 (a bit of both).

If it were possible I'd prefer optimizing for 1.:

* Drop `modelCacheRef`
* Place the cache on all clusters that match its selector
* Schedule MDs to any cluster that matches the MD's selector - regardless of whether there's a cache there

In this world it's up to the platform team to make sure the cache is where they want it. A model can't implicitly carry it to a new cluster the platform team didn't pre-hydrate. No MD ever pays the "double download" tax. They either load from cache, or load directly from the source.

All that said, I can't think of a way to actually make 1. work given that the MD needs to know whether its loading from cache or not when forming its engine args.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+No hydration is retracted out from under a live replica, and an unreferenced cache
+reclaims itself per cluster with no manual cleanup.
+
+### Hydrating before ready
+
+A replica can be scheduled onto a cluster new to the model before its PVC exists.
+Rather than fail the mount, the replica gates readiness on its cluster's
+`ModelCacheHydration`: it reports `Hydrating` until the PVC is Bound, then
+proceeds. The first replica on a new cluster pays the download once, as a visible
+state rather than a mount error.
```
negz: Can you clarify: is the ModelReplica watching the ModelCacheHydration to know when it's ready? Or the PVC directly? Watching the MCH seems best (e.g. if one day we have caches that don't use PVCs).

design/modelcache-hydration.md
dennis-upbound: No 63-char problem here: `resource.child_name` is DNS-label-safe by construction. It joins the parts, truncates the prefix, and appends a 5-char hash, so the result is always a valid label at or under 63 chars, including the child's own name with `cluster` folded in. The PVC/Job/Secret keep the cluster *out* of their names (`child_name("modelcache", ns, cache-name[, "hydrate"|"auth"])`) so they match the serving-side mount, and those stay ≤63 too.

Added a line to the naming section spelling this out. Pushed.

design/modelcache-hydration.md
dennis-upbound: Good catch, this is the real gap in the gate. It had no failure exit: a bad token or bad revision would hold the replica in `Hydrating` forever.

The child already has a `Failed` phase, so the fix is to use it: a failed hydration surfaces as `ArtifactReady=False` with reason `HydrationFailed` on the `ModelCache`, and the gated replica fails with that reason instead of gating indefinitely. The hydration `Job`'s `backoffLimit` bounds retries, so a permanent failure stops and is reported rather than retried forever.

Added a failure path to *Hydrating before ready*. Pushed.

design/modelcache-hydration.md
```diff
@@ -0,0 +1,295 @@
...
+`child_name("modelcache", <ns>, <name>)`, Job `…,"hydrate"`, Secret `…,"auth"`.
+Two things depend on this:
+
+1. The serving side mounts the PVC by that name (`base.cache_pvc_name` in
+   `compose-model-replica` is the same function), so the mount contract is
+   untouched.
+2. On upgrade the child adopts the existing, already-Bound PVC instead of
+   provisioning a new one and re-downloading the weights.
+
+## Placement-driven footprint
```
negz: I think this'll need https://github.com/crossplane/crossplane/pull/7572 to function correctly. Without that functionality nothing will trigger the ModelCache function to run when new ModelReplicas are placed, so the ModelCache function won't know to create a new ModelCacheHydration.

Given https://github.com/crossplane/crossplane/pull/7572 is approved I expect it'll ship next month in Crossplane v2.4. We may want to hold this implementation until then.

negz (commented):
I like this direction overall. The two things that need the most thought before we proceed:

* Does this need to wait for https://github.com/crossplane/crossplane/pull/7572?
* Is hydrate-on-demand definitely desirable?

dennis-upbound:
Went with pre-warm on both counts.

- **crossplane#7572:** only the placement-derived footprint needs it. Pre-warm makes `clusterSelector` authoritative, so `ModelCache` never watches replica placement and needs no cross-resource trigger. No hold on v2.4.
- **hydrate-on-demand:** dropped as the default. It's now the rejected alternative (the double-download tax, the #7572 dependency, and a model carrying private weights onto clusters the platform team didn't choose). Pre-warm pays the big hydration once, ahead of deployment; a deployment loads from the cache where it's staged and from the source elsewhere.

Doc reworked around this. Inline replies below.

design/modelcache-hydration.md
dennis-upbound: Went with option 1. On the engine-args blocker: `compose-model-deployment` already resolves the cache footprint (`resolve_cache_footprint`, the #189 path), so it knows per-replica whether that replica's cluster is in the footprint. It passes that down, and `base.cache_mount` injects `--model=<mount>` on footprint clusters and leaves the engine loading from source elsewhere. The decision is known at compose time, so forming args needs no runtime lookup, so option 1 works, and it drops the #7572 dependency too. New "Cache or source, per replica" section spells this out.

design/modelcache-hydration.md
dennis-upbound: This was the strongest reason to move off placement-derived. Pre-warm removes the dependency: the footprint is the static `clusterSelector`, so `ModelCache` doesn't need a watch on `ModelReplica` placement to stay correct. No hold needed. #7572 is now noted as required only by the on-demand alternative.

design/modelcache-hydration.md
dennis-upbound: With pre-warm the parent doesn't walk replicas at all. The footprint is the `clusterSelector`, so `compose-model-cache` fans out on the selector and never requires the replica set. The label is a nice optimization for the on-demand path, but that path is now the rejected alternative.

design/modelcache-hydration.md
dennis-upbound: Under pre-warm the selector *is* the footprint, so it keeps `clusterSelector` and its plain meaning: it's exactly where the cache is placed. The "doesn't limit placement" ambiguity came from the placement-derived model, which is gone.

design/modelcache-hydration.md
dennis-upbound: Corrected. #189 is now called out explicitly as an alternative ("Constrain placement to the footprint"): it constrains MD placement to the cache footprint, and this proposal relaxes that and loads from source off the footprint instead.

design/modelcache-hydration.md
dennis-upbound: Agreed, and that's the spirit of the rework: went straight to pre-warm rather than keeping the hedged both-modes design. Kept `modelCacheRef` only because it's load-bearing for the cache-vs-source decision, not for migration reasons.

design/modelcache-hydration.md
dennis-upbound: The `ModelCacheHydration` object, not the PVC, for exactly your reason (a future non-PVC cache keeps the same readiness contract). Stated explicitly now in "Hydrating before ready."

design/modelcache-hydration.md
negz: > injects `--model=<mount>`

Doesn't this go against our overarching design principle of not injecting engine args though? What if a new engine shipped tomorrow that used a different flag for this?

I know we already inject a bit today, but I was hopeful that was something temporary we could walk back.

negz: Thinking out loud, some options we could consider:

1. Admit we need to inject flags, transparent passthrough of engines isn't realistic.
2. Loosen to allow injecting flags, but only when they're best-effort optimizations (like using a cache).
3. Allow the user to specify conditional args - e.g. change these args if there's a cache.

dennis-upbound: Settled this. Rather than inject `--model` (or vary it per replica), a cache-referencing deployment is now placement-constrained to the cache footprint, so the engine command is uniform and Modelplane injects no engine flags. Where a value depends on the cluster it's an env var the command references, the way `MODELPLANE_LEADER_ADDRESS` works today.

The load-from-source-off-the-footprint idea is dropped: it needed the per-replica model-arg injection, and it doesn't generalize — a loader-plugin cache like ModelExpress needs `--load-format` plus a loader-capable image that are invalid off the cache, so those deployments have to be placement-constrained anyway. Constraining placement for every cache is simpler and keeps the engine opaque.

Two modes now: reference a cache and run only where it's pre-warmed, or reference none and run anywhere loading from the source with your own token. Doc updated ("Two modes" + "Modelplane injects env values, never engine flags").

design/modelcache-hydration.md
```diff
@@ -0,0 +1,358 @@
...
+loader-capable image. The ML team writes that, because they chose the cache, and
+Modelplane wires the env and the cluster-side pieces.
+
+Either way the flag that names the model belongs to the engine, so Modelplane stays out
+of it and injects only env values. Placing only where the cache is pre-warmed
+makes whatever the ML team wrote valid wherever the replica runs. How the bytes reach
+the cluster, whether a shared filesystem, peer-to-peer distribution, or GPU-to-GPU
+streaming, is the platform team's concern and orthogonal to the command. The catalog of
+backends is a separate design, and each backend has to present one of these two
+contracts.
```
negz: What env value will it inject? Can you show how a ModelDeployment will use it?

design/modelcache-hydration.md
negz: For posterity: we landed at keeping the behavior we introduced in #189. See the "Load from the source off the footprint" alternative for rationale.

negz (approved):
I'd like to clarify what env var(s) you'll inject and how they'll be used but other than that LGTM.

design/modelcache-hydration.md
dennis-upbound: In the common case, nothing. The cache mounts at the same path on every footprint cluster, so the ML team just writes `--model=/mnt/models`. Added a snippet showing it.

Injection only covers a value that varies per cluster: `MODELPLANE_LEADER_ADDRESS` today, or a loader-plugin's server address. Always an env var via `$(VAR)`, never a flag we write.

dennis-upbound: Got it, and my earlier "nothing" over-rotated. Locked the name.

With a cache Modelplane owns the mount, so it injects the mount path as `MODELPLANE_MODEL` and the ML team writes `--model=$(MODELPLANE_MODEL)` (or `--model-path=$(MODELPLANE_MODEL)` on SGLang). They never hardcode our mount path, so we stay free to move it. Same `$(VAR)` mechanism as `MODELPLANE_LEADER_ADDRESS`. Without a cache they write `--model=<source>` directly, since the source is theirs.

Updated the section and the snippet. `MODELPLANE_MODEL` is my pick for the name, easy to change if you'd rather something else.

Issue #269, opened 2026-06-23
Give engine metrics collection a managed lifecycle


pluna:
### What problem are you facing?

Collecting an engine's metrics today is hand wired. vLLM exposes `/metrics` on its serving port, and Modelplane runs a Prometheus on each workload cluster with open `PodMonitor` discovery, so an operator can scrape an engine by writing a `PodMonitor` that selects the engine pods. The engine container port is unnamed, so the `PodMonitor` has to match it by number with `targetPort`. The wiring also shifts with the deployment's shape: a leader/worker gang is scraped on the leader, and prefill/decode splits into two role-labelled engines.

Nothing owns that wiring. The operator assembles the `PodMonitor` by hand, keeps it in sync with the deployment's shape, and removes it when the deployment goes away. The example in #264 shows the manual path.

### How could Modelplane help solve your problem?

An operator should be able to turn on metrics collection for a deployment and have it managed for them, instead of hand building a `PodMonitor` and keeping it in sync. Collection should come up when the deployment is created and be removed when it goes away, with nothing to assemble per deployment or clean up on teardown.

Issue #77, opened 2026-05-13
v0.2 ModelService observability.traces: OTLP gateway emission


dennis-upbound:
Customers running production AI workloads need per-request observability — prompts, completions, tokens, latency, eval scores, user sessions. That layer is filled today by Langfuse, Langsmith, Phoenix, Helicone, Datadog LLM Observability, and homegrown OTel pipelines. Modelplane sits at the gateway and sees every request; it's the natural emission point. But picking one app-observability backend creates lock-in for the others.

The right shape is: **the gateway emits OpenTelemetry traces with `gen_ai.*` semantic conventions to an operator-configured OTLP endpoint**. Customers point at Langfuse, Langsmith, Phoenix, a custom OTel collector, or anything else that speaks OTLP. Modelplane stays neutral.

This is the v0.2 companion to [#74 fleet signal bus](https://github.com/modelplaneai/modelplane/issues/74). Both share OTLP as the transport, but carry different signal types for different audiences.

## Two observability planes, one transport

|  | [#74 signal bus](https://github.com/modelplaneai/modelplane/issues/74) | This issue (OTel gateway traces) |
|---|---|---|
| **Signal** | Aggregated metrics, sketches, time series | Per-request spans (`gen_ai.*`) |
| **Volume** | KB/s per cluster | MB/s per active deployment (sampled) |
| **Retention** | 30d sketches in control plane | Hours hot / months cold in user's backend |
| **Audience** | Operators, scheduler, fleet matcher | Application developers, eval pipelines |
| **PII exposure** | Hashes / aggregates only | Full prompts/responses |
| **Backend** | Modelplane control plane | User's choice (Langfuse / Langsmith / Phoenix / custom) |

Same OTLP collector substrate; one fanout aggregates spans into #74 sketches, the other forwards raw spans to the user's configured backend.

## Shape

```yaml
apiVersion: modelplane.ai/v1alpha1
kind: ModelService
spec:
  endpoints:
    - selector: { matchLabels: { modelplane.ai/deployment: kimi-k2 } }
  observability:
    traces:
      enabled: true
      endpoint: https://cloud.langfuse.com/api/public/otel    # or langsmith / phoenix / custom otel-collector
      headers:
        Authorization:
          valueFrom: { secretKeyRef: { name: langfuse-creds, key: auth } }
      sampling:
        ratio: 1.0
```

[...]

PR #363, opened 2026-07-21
Design: managed metrics collection

dennis-upbound:

A design doc and a draft docs page. No code. Addresses #269.

Modelplane installs a Prometheus on every workload cluster and stops. The operator writes their own `PodMonitor`, keeps it in sync, and reaches the store by `port-forward`. Every component names its metrics differently, and every cluster answers only for itself.

The design proposes an OpenTelemetry collector on each inference cluster, normalizing onto one `modelplane_*` surface and exporting OTLP to a collector on the control plane. That one holds no store, so the control plane stays a reconciler, and the operator picks the backend. The same pipeline carries #77's traces and the gateway's usage logs later.

Two calls worth flagging before you read it:

- **Top-line latency comes from the gateway, not the engines.** Envoy AI Gateway measures every request it proxies under the GenAI conventions, for whatever engine is behind it. Engine histograms then never have to merge, SGLang's inter-token latency never gets renamed onto vLLM's time per output token (different measurements, so that rename is wrong rather than lossy), and an unknown engine reports its headline numbers with no configuration.
- **Rates, quantiles and cross-cluster ratios move to your backend**, shipped as queries rather than recording rules. An operator running no backend gets a normalized stream and no fleet SLO series. Alternatives carries the Prometheus design that keeps them, and what it costs.

Two kinds, both envelopes for collector configuration Modelplane doesn't read: `TelemetryDestination` for the exporters block, `MetricMapping` for OTTL statements that have to reach every cluster. One new component, a control-plane exporter for what nothing else can report — allocatable GPUs under DRA, GPU-to-replica attribution, replica lifecycle timings, cluster connectivity.

Removing the per-cluster kube-prometheus-stack is the one breaking change and lands on its own with a release note.

`docs/content/guides/telemetry.md` is the page a user would read. It's `draft: true` and takes `collecting-engine-metrics`' URL when the stack removal lands.

I have:

- [x] ~Run `nix flake check`~ (design doc and a draft docs page, no code paths)
- [x] ~Added or updated tests~ (no code)


design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+## Open questions
+
+- **Port name.** `http` (the serving port that also serves `/metrics`) versus
+  `metrics`. Leaning `http`, since it's the one serving port.
+- **Configurability.** `interval` and `path` are fixed (30s, `/metrics`) for now.
+  The nested `metrics` object leaves room to add them if a real need appears.
+
+## Interaction with #264
+
+The [#264](https://github.com/modelplaneai/modelplane/issues/264) example
```
haarchri: In Release notes we need to add a line telling existing users to delete their hand-written `podmonitor.yaml`, or they get a duplicate scrape job against the same pods after upgrade.

design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+
+A `ServiceMonitor` scrapes through a Service, so it needs a Service per scrape
+target. The engine and picker pods are the targets, and a `PodMonitor` scrapes
+them directly, matching the manual path.
+
+## Open questions
+
+- **Port name.** `http` (the serving port that also serves `/metrics`) versus
+  `metrics`. Leaning `http`, since it's the one serving port.
+- **Configurability.** `interval` and `path` are fixed (30s, `/metrics`) for now.
```
haarchri: wonder if 30s is too Long If it comes to KV Cache utilization e.g.

design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+## Summary
+
+Collecting a deployment's metrics is hand-wired today. vLLM exposes `/metrics` on
+its serving port, the serving stack runs a Prometheus on every workload cluster
+with open `PodMonitor` discovery, and an operator writes a `PodMonitor` by hand to
+scrape the engine pods (the [#264](https://github.com/modelplaneai/modelplane/issues/264)
+example). Nothing owns that wiring. The operator builds it by hand and keeps it in
+sync with the deployment's shape. They delete it on teardown.
+
+Instead, Modelplane composes the collection, per source:
```
haarchri: is Monitoring for Cache / PVC here out of scope?

design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+The pd-sidecar's port stays unnamed. The `PodMonitor` then scrapes that port by
+name, which matters for prefill/decode: the decode engine serves on 8001 because
+the pd-sidecar takes 8000, so a `PodMonitor` matching port 8000 by number scrapes
+the sidecar on decode pods, not the engine. Referencing the port by name scrapes
+the engine on every pod regardless of its number. Naming is additive; the Service
+and probes that reference the port by number are unaffected.
+
+### What gets scraped
+
+The `PodMonitor` ingests everything the engine exposes on `/metrics`. What the
```
haarchri: what are the expected series-per-pod ? is the Prometheus Stack configured with rolling / retention of the storage ?

design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+
+Both come up with the deployment, are reclaimed with it, and share one opt-out
+field. On by default:
+
+```yaml
+spec:
+  replicas: 1
+  template:
+    spec:
+      metrics:
```
haarchri: consider observability/monitoring as the object name if it's going to grow past scraping... later

design/metrics.md
```diff
@@ -0,0 +1,208 @@
...
+kind: PodMonitor
+metadata:
+  name: <replica>-metrics
+  namespace: default
+spec:
+  selector:
+    matchLabels:
+      modelplane.ai/serving: <replica-name>
+  podMetricsEndpoints:
+  - port: http
```
haarchri: is http the right name here?

design/metrics.md
dennis-upbound: Agree, and it lines up with #77 (`ModelService.observability.traces`). Renamed to `spec.template.spec.observability.metrics.enabled` so traces and logs can slot beside `metrics` later. (Metrics live on the deployment since we scrape its pods; #77's traces live on `ModelService`, different resource, same `observability` umbrella.) Pushed.

design/metrics.md
dennis-upbound: Yes, without deleting it they'd get a duplicate scrape of the same pods after upgrade. Called it out as a release note in the #264 interaction section. Pushed.

design/metrics.md
dennis-upbound: For routing this interval doesn't matter: the EPP scrapes engine `/metrics` on its own fast internal loop (~100ms) to make routing decisions, so routing isn't gated on it. This `PodMonitor` feeds Prometheus for dashboards and alerting, where 30s is a normal default. If finer KV series are wanted, `interval` is a knob on the `metrics` object (open question). Made that distinction explicit in *What gets scraped*.

design/metrics.md
dennis-upbound: Out of scope here, yes, this doc is serving metrics (engine + EPP). `ModelCache` PVC and hydration observability is a separate concern that belongs with the ModelCache work. Added a scope line to the summary.

design/metrics.md
dennis-upbound: Order of dozens of series per engine pod (request/latency histograms, KV-cache, throughput), plus the `llm_d_epp_*` set wherever an EPP runs. Retention and storage sizing are the serving stack's Prometheus config in `compose-serving-stack`, not this composition, worth a sane default there but out of scope for this doc. Noted it in *What gets scraped*.

design/metrics.md
dennis-upbound: It's an open question in the doc (`http` vs `metrics`). Leaning `http` since it's the one serving port that also serves `/metrics`, but happy to go `metrics` if that reads clearer.

negz:
@dennis-upbound I think we need to step back and think about what we want to help folks monitor.

There's a few things you could monitor in Modelplane:

1. Inference signal (data plane). vLLM /metrics, EPP llm_d_epp_*, Envoy. TTFT, tokens/sec, queue depth, KV-cache utilization, request/error rates per model. "Is my model serving well, and is it saturated?"
2. Substrate health (the stack Modelplane installs on each workload cluster). Gateway up? cert-manager, LWS controller, NVIDIA DRA driver healthy? GPUs allocatable? "Is the machinery on this cluster working?"
3. Control plane health (Modelplane itself). Crossplane reconcile rates/errors, function latency and panics, provider health, the fleet scheduler actually placing replicas, XR Ready/Synced. "Is the thing I operate working?"
4. Fleet rollup. Across all clusters/deployments: total capacity, GPU utilization, how many deployments degraded, cost.

I think MD authors will care about 1. The rest feel more like platform team concerns.

This proposal helps with 1, but it feels incomplete. If I understand correctly, it configures the Prometheus instance we already deploy to each cluster to scrape ModelReplica (i.e. vLLM, EPP) metrics, but what then? How does the MD author see and consume those metrics? If the platform team needs to setup plumbing to actually expose the metrics somewhere the MD author can use them (e.g. a dashboard), why ask the MD author to opt-in (or out)?

My _hunch_ is Modelplane should do something like:

* Deploy a Prom per IC
* Always setup PodMonitors for all MD bits (engines, EPPs, etc) - no opt-in or opt-out
* Always setup PodMonitors for all serving stack components
* Deploy a centralized Modelplane Prom instance
* Have the per-IC instances feed into the centralized one, which also scrapes Crossplane metrics for Modelplane control plane health
* Setup recording rules as appropriate

This'd give you one central Prom instance you can scrape to get your entire Modelplane deployment's metrics. Metrics you'd presumably feed onwards into your monitoring and alerting system of choice.

dennis-upbound:
Completely agree, reframed the doc around this. It now opens with the four things to monitor (data plane / substrate / control plane / fleet roll-up), makes `PodMonitor` collection always-on at every layer (engine, EPP, serving-stack components) with no opt-in or opt-out, and adds a central Modelplane Prometheus that the per-IC instances feed (remote-write or federation) and that also scrapes Crossplane for control-plane health, with recording rules for the fleet roll-up.

The per-deployment opt-out field is dropped (now the first rejected alternative). Open questions left: remote-write vs federation, central retention sizing, and phasing (data-plane + substrate first, control-plane + roll-up after).

design/metrics.md
dennis-upbound: Superseded by the reframe: after Nic's review the doc drops the per-deployment opt-out field entirely. Collection is now always-on at every layer (engine, EPP, serving-stack) with no opt-in or opt-out, feeding a central Modelplane Prometheus. So there's no longer a field to name under `observability`. Leaving this for context; the `observability` grouping question may come back if we add a deployment-level knob later.

negz:
> reframed the doc around this

@dennis-upbound did you forget to push? 🤔 Doc looks unchanged to me.

dennis-upbound:
Agree — pulled the scope to exactly this. The doc now composes collection always-on inside each IC (engines, EPPs, serving-stack), no MD-author opt-in/out, and exposes the cluster's Prometheus URL on the `InferenceCluster` status (`status.metrics.prometheusURL`) so the platform team can scrape or federate without reaching into internals. Cross-cluster aggregation and Crossplane monitoring stay with the platform team. I left the one-central-Prom-for-the-whole-deployment out of this proposal and flagged it as the main open question, so we can decide separately how far Modelplane takes it.

(Also: the doc genuinely hadn't updated when I said it had — I'd pushed to the wrong remote. Fixed now, the diff reflects the change.)

design/metrics.md
```diff
@@ -0,0 +1,238 @@
...
+4. **Fleet roll-up.** Across every cluster and deployment: total capacity, GPU usage,
+   how many deployments are degraded, and cost.
+
+This proposal covers (1) and (2) inside each cluster and makes the data discoverable.
+An MD author cares about (1), and the platform team owns the rest. A per-deployment
+opt-in would cover only (1), and even then the platform team has to plumb the metrics
+somewhere the author can read them. So collection is always on inside each cluster,
+with no toggle for an MD author to manage. Aggregating across clusters and monitoring
+the control plane stay with the platform team, and the end of this document returns to
+them.
```
negz: Nit: the mention of per-deployment opt-in  feels like it's a reaction to the earlier iteration of the design. You already cover it in alts considered. I think you can drop it here.

design/metrics.md
```diff
@@ -0,0 +1,238 @@
...
+than wiring up what each cluster already runs, so this proposal leaves aggregation to the
+platform team and stops at the discoverable URL.
+
+### One cluster-wide PodMonitor from the serving stack
+
+`compose-serving-stack` could install a single `PodMonitor` selecting all serving pods
+(`modelplane.ai/serving` Exists). It would have no per-deployment lifecycle: it wouldn't
+come and go with a deployment. Composing the engine monitor per replica ties that
+collection to the thing it observes, while the serving-stack monitors, which are
+per-cluster, are composed once with the stack.
```
negz: Would this work? i.e. Just matching stuff to monitor as it comes and goes?

If so, what's the benefit of the current proposal over this alternative?

design/metrics.md
```diff
@@ -0,0 +1,238 @@
...
+
+Each `InferenceCluster` already runs a Prometheus, installed by `compose-serving-stack`
+(kube-prometheus-stack). With the `PodMonitor`s above composed, that Prometheus holds
+all of the cluster's Modelplane sources, from the engines and EPPs to the serving-stack
+components. Modelplane exposes its in-cluster URL on the `InferenceCluster` status:
+
+```yaml
+status:
+  metrics:
+    prometheusURL: http://prometheus-operated.modelplane-system.svc:9090
```
negz: How will we expose Prometheus so it can be scraped from outside the cluster?

design/metrics.md
dennis-upbound: Dropped it — the opt-in only lives in Alternatives now.

design/metrics.md
dennis-upbound: Agreed, and switched to it. Now that collection is always-on with no opt-out, the per-replica composition earns nothing over one cluster-wide monitor — so the serving stack composes a single `PodMonitor` matching `modelplane.ai/serving` (Exists) for engines, plus one for EPPs and one for the substrate, composed once with the stack. The per-replica version is now the first rejected alternative.

design/metrics.md
dennis-upbound: Made it explicit: Modelplane only publishes the in-cluster URL on the InferenceCluster status — it doesn't expose Prometheus outside the cluster. Reaching it across clusters stays the platform team's job through whatever cross-cluster path they already run (remote-write, a federation scrape, a peered network), the same way they reach any in-cluster service. Exposing it publicly would be a security surface we don't want to own; publishing where it is gets them what they need.

dennis-upbound: Update — the direction changed after discussion: aggregation is now Modelplane's job, up to a control-plane-level view, rather than the platform team's. The central layer receives each cluster's metrics (remote-write) or pulls them, so there's no per-cluster Prometheus to expose outside the cluster. The doc's been revved to match.

dennis-upbound:
Revved the doc. It now aggregates every cluster's metrics up to one Modelplane-level view at the control plane, rebranded under `modelplane_*` with a normalized label set (`engine`, `cluster`, `deployment`, `model`), rather than stopping at per-cluster collection — that's now the first rejected alternative. It also weighs the per-cluster mechanism, the incumbent Prometheus stack versus an OpenTelemetry collector, and leans toward the collector (lighter, does the `modelplane_*` rename in-pipeline, and something we already run elsewhere). The mechanism choice is the main open question.

negz:
@dennis-upbound thanks! I like the direction. I think this needs some POC work before I can approve. Specifically I'd like to lock down:

1. Do we use Prom or Otel collector?
2. What metrics do we want to expose (i.e. what rewrite rules do we need)? What metrics are possible? Can we normalize metrics for different engines?

dennis-upbound:
Locked both.

**Collector: OpenTelemetry.** The GenAI semantic conventions are the naming target, the rename runs in the collector's transform processor in-pipeline, and one pipeline carries metrics, traces (#77), and logs. A Prometheus stack would push the rename into per-cluster recording rules and stay metrics-only, so it's now a rejected alternative.

**Metrics + normalization.** Added the `modelplane_*` set and a per-engine mapping table (vLLM / SGLang / TRT-LLM), with TTFT, ITL, TPOT, and the prefill/decode split. On "can we normalize across engines": vLLM and SGLang map cleanly (near-identical names, both align to the OTel set); Triton/TRT-LLM exposes batch-manager stats, so those rows are derived — called out, not hidden.

On staying correct for an engine we don't recognize: an engine-type label picks the mapping, the same mechanism the GAIE picker already uses for routing, and an unmapped engine degrades to raw names rather than a wrong guess. See the new "Capture from an opaque engine" and "Normalize to `modelplane_*`" sections.

Still needs the POC to confirm the derived TRT-LLM rows and the collector's rename config, agreed.

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+**Normalize to `modelplane_*`.** Each engine names its metrics its own way (`vllm:*`,
+`sglang:*`). The collector renames them to one Modelplane vocabulary, picked by an
+engine-type label, so a dashboard reads Modelplane's names and not each engine's. That
+label is Modelplane's to stamp, from a new `engines[].type` on the `ModelDeployment`.
+
+**Aggregate to one view.** Every cluster's series roll up to one view at the control
+plane, so one query covers the whole deployment instead of a per-cluster island.
+
+**Collect with OpenTelemetry.** The collector is an OpenTelemetry collector, added beside
+the per-cluster Prometheus rather than replacing it. The section below gives the reasons.
```
lsviben: Is this Prometheus mention a leftover from the previous solution. With OTEL collector we dont need it anymore.

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+**Fleet roll-up.** Across every cluster and deployment: total capacity, GPU usage,
+degraded deployments, and cost.
+
+All four are in the central view. The data plane and substrate are collected on each
+cluster and aggregated up, the control plane is scraped at the center, and the fleet
+roll-up is the collector's aggregation over the collected series.
+
+## Collect on every cluster
+
+On each cluster Modelplane collects from every source it owns, with no opt-in or opt-out.
```
lsviben: Seems ok with setting it default as on, but why no opt-out? If the OTEL collector is not pushing the data anywhere, for what do we need it running?

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+
+- **The serving label spans every shape.** `modelplane.ai/serving` is on standalone pods,
+  LeaderWorkerSet leaders, Grove leader cliques, and both prefill and decode engines,
+  since it's the label the InferencePool selects on. One selector on it follows the shape,
+  so leader/worker, prefill/decode, and a Dynamo cluster's PodCliqueSets need no special
+  casing. The Dynamo work added a fourth workload kind without touching this, which is the
+  test this approach had to pass.
+- **The stack already scrapes.** `compose-serving-stack` runs a metrics stack on every
+  workload cluster and already scrapes the gateway's Envoy proxies, so adding a target is
+  composition, not new infrastructure.
```
lsviben: sorry i am not that familiar with Modelplane yet. Would that Prometheus composed here be replaced with the OTEL collector?

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+    vllm:gpu_cache_usage_perc: modelplane_kv_cache_utilization
+  labels:
+    add: { engine: vllm }
+```
+
+`compose-serving-stack` reads every `MetricMapping` as a required resource, the same way
+`compose-model-deployment` reads `InferenceCluster` and `ModelCache`. It renders them into
+the collector's config, the ConfigMap the OTel collector loads on each cluster. The
+`rename` map becomes transform-processor rules, applied to metrics from the pods the
+`selector` matches. A new engine is a new `MetricMapping`, not a package change.
```
lsviben, line 210: it would be good to see an example of the OTEL config generated based on a MetricMapping

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+design composes against. So the collector is added beside it rather than in place of it,
+and what the collector removes is per-cluster recording rules and a second Prometheus at
+the center, not the one already on each cluster.
+
+That leaves one thing to settle: a plain OTel collector doesn't read `PodMonitor`s. Either
+it runs under the OpenTelemetry Operator, whose target allocator consumes `PodMonitor` and
+`ServiceMonitor` directly and keeps the cluster-wide-selector shape below, or it carries
+its own Kubernetes service-discovery scrape config and `PodMonitor` is the wrong word
+throughout this document. The target allocator is the smaller change and preserves the
+#264 upgrade path, but nobody has run it here, so it is a decision this design records
```
lsviben: Is this Prometheus/PodMonitors/ServiceMonitors also a leftover? The OTEL Collector can scrape the same as Prometheus using its receiver, and as we are sending the data out we dont need Prom as a data storage.

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+  `volcano_queue_allocated_scalar_resources` against `_deserved_` and `_capacity_`, with
+  `volcano_queue_overused` for fairness.
+- **Preemptions and evictions.** `scheduler_preemption_victims`,
+  `volcano_pod_preemption_victims`.
+
+A scheduler's mapping is a `MetricMapping` like an engine's, and the degradation rule
+carries over, punctuation caveat included. An unmapped scheduler still gets scraped
+under its own names, and Modelplane surfaces that rather than guessing.
+
+## Aggregate to one view
```
lsviben, line 367: would be good to have a section as well on cardinality and what labels we dont really need in the metrics. We can remove some in the OTEL collector processing before sending it out.

design/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+control plane, which also collects the control plane's own metrics (Crossplane, the
+functions, the fleet scheduler). One query then covers the whole deployment rather than a
+per-cluster island an operator stitches together by hand.
+
+The cluster sends by pushing outbound. Each collector remote-writes or OTLP-exports to a
+control-plane endpoint, so the workload cluster needs only egress, which is what makes it
+work across regions and through firewalls. Nothing inbound to the cluster is required, and
+nothing is exposed outside it. The center can pull instead where it already reaches the
+cluster, publishing the cluster's endpoint on the `InferenceCluster` status, but push is the
+default for the regional and firewalled case.
```
lsviben: Would be good to state which exporters we would support and how to configure them as a ModelPlane user. Auth is also something not touched yet in this doc, and the OTEL collector will need a way to auth to wherever it is sending the data.

lsviben (commented):
I like the direction this design is taking with using the power of OTEL to collect and send the data from Modelplane. Discussed this with Dennis in a 1on1 as well and just left here some notes on things the desing could clarify better.

design/metrics.md
dennis-upbound: Leftover, and one I added rather than inherited. I'd argued Prometheus had to stay because its operator defines the `PodMonitor` CRD this design composes against, which gets the dependency backwards: `PodMonitor` is a consequence of having chosen Prometheus, not a requirement of collection. The `prometheus` receiver does Kubernetes SD itself.

The stack is gone now, with each thing it did named: `k8s_cluster` for kube-state-metrics, `kubeletstats` for cAdvisor, `hostmetrics` for node-exporter. Two costs are written down rather than left to be found — no local store to query ad-hoc, and the `collecting-engine-metrics` guide loses both halves of its workflow.

design/metrics.md
dennis-upbound: Good question, and it changes the shape rather than just the wording. "Always on" was arguing against a *per-deployment* toggle, which I still think is right: a `ModelDeployment` author shouldn't opt out of telemetry the platform team consumes. But you're right that a collector with nowhere to send is cost with no reader.

So the switch moves up a level. No export destination configured, no collector composed. Once a destination exists, collection is on for every source Modelplane owns, with no per-deployment switch.

design/metrics.md
dennis-upbound: Yes, replaced. And the migration is smaller than it looks: `compose-serving-stack` scrapes the Envoy proxies today through the Prometheus chart's `additionalScrapeConfigs`, which is a `kubernetes_sd_configs` block. That's the same format the collector's `prometheus` receiver takes, so the Envoy target moves across as it stands rather than being rewritten.

What it does add is a second collector tier. `kubeletstats` and `hostmetrics` are node-scoped, so they need a DaemonSet, where `k8s_cluster` and the engine and EPP scrapes run as one Deployment. The doc now says so, and asks whether the DaemonSet tier belongs in the first cut at all.

design/metrics.md
dennis-upbound: Added — the vLLM mapping now sits beside the config it renders, so the OTTL is visible: one `transform` statement per rename, gated on the `engine` resource attribute that `k8sattributes` lifts off the pod label, plus the scrape config and its two relabel rules.

It also makes the degradation concrete. An unmapped engine matches the scrape config and no `transform` block, so it arrives under its own names without a rule having to say so.

design/metrics.md
dennis-upbound: Same leftover, same answer as above — the stack and the `PodMonitor` framing are both out. Discovery is now a scrape config in the collector's own ConfigMap, so no CRD is involved and there's nothing to reconcile between an operator and a receiver.

The alternatives entry keeps "a Prometheus stack" as a considered option, since the local store it gives an operator is a real thing to give up, but it's recorded as rejected rather than half-adopted.

design/metrics.md
dennis-upbound: Agreed, added. Dropping `pod`, `pod_uid` and `container_id` — each is new on every restart, so each turns a rolling update into a fresh set of series nothing writes to again. Keeping `engine`, `cluster`, `model`, `deployment` and `namespace`, the dimensions the roll-up and dashboards group by.

One thing worth flagging since the obvious processor is the wrong one: `attributes`/`delete_key` removes the label but leaves the series that collided on it as separate undefined points. Merging needs `metricstransform` with an aggregation action. Getting it wrong looks like it worked.

Histogram buckets stay as the GenAI conventions define them — `le` is what makes the fleet histogram and the SLO ratio possible.

design/metrics.md
dennis-upbound: The biggest gap in the doc, and it needed a section rather than a sentence. Both are in now.

Exporters: OTLP (`otlp` over gRPC, or `otlphttp`) as the contract, `prometheusremotewrite` for an operator who wants a Prometheus-compatible store. Vendor exporters go behind the gateway, so one configuration serves the fleet instead of one per cluster. A user doesn't write collector YAML: the destination is fleet-level config propagated to each `ServingStack`. Where exactly that field or kind lives I've left open rather than guessed.

Transport is the part I'd most like your eyes on. Modelplane has exactly one connection it can count on and it runs the wrong way: the control plane reaches a cluster's API server through the kubeconfig, and nothing guarantees a path back from a firewalled GPU cluster. So push to a gateway collector is the default, with its credential riding the same Secret propagation that already carries a HuggingFace token to every matched cluster. Where a cluster has no egress, the center scrapes the collector through the API server proxy over the connection it already has — no inbound exposure, no second credential, at the cost of putting every series through the API server. Direct pull per cluster is named only to rule it out.

design/metrics.md
lsviben: 👍  on if no export destination configured, no collector composed.

design/metrics.md
```diff
@@ -0,0 +1,628 @@
...
+plane to every matched cluster so hydration can read a HuggingFace token. A bearer token or
+client certificate for the OTLP endpoint travels the same way, through the same mechanism,
+so this adds a Secret to propagate rather than a way to propagate Secrets.
+
+**Pull through the API server proxy.** The fallback, for a cluster with no egress. The
+center scrapes it over the connection it already has. The Kubernetes API server proxies to
+in-cluster Services, so the collector is reachable at
+`/api/v1/namespaces/<ns>/services/<collector>:<port>/proxy/metrics` with the credential
+Modelplane already holds. No inbound exposure, no firewall change, no second credential;
+the only addition is a `ClusterRole` granting `services/proxy`.
```
lsviben: Do we want to support 2 modes. Maybe going with just pull from the clusters to the control plane makes more sense. 

Makes it also easier considering that we want to roll up all telemetry in one place fleet-wide, as it seems that InferenceGateways are becoming their own clusters as well. 

Then we have just one place where to set up exporters and auth for them.

lsviben: Scratch that above 😄 . I did some digging and agree with your sentance below that this would be too intensive.

Anyway, i still think we should at least start with 1 mode only to keep things simpler, and that should be the push one.

design/metrics.md
```diff
@@ -0,0 +1,628 @@
...
+resolved per cluster, as above.
+
+Where that configuration lives needs deciding before this is built, not during. A field on
+a fleet-level resource and a kind of its own both work, and the choice is the same one
+`MetricMapping` faced: a typed kind validates on apply and lists under `kubectl get`, at
+the cost of another kind.
+
+It blocks more than the implementation. Configuring a destination is the first thing a user
+does, since nothing is collected until one exists, so it is the first thing the docs
+describe, and a draft of that page had to invent a `MetricsDestination` kind to say
```
lsviben: NIT: suggest going with TelemetryDestination/ObservabilityDestination instead, to keep it open for more signals in the future

design/metrics.md
dennis-upbound: Agreed, and it's out. One transport, push, and no field selecting it.

You talked me into it with the retraction rather than the original suggestion: once pull-only is off the table, the second mode is a fallback whose own analysis in this section admitted it was degraded, and I'd still have been paying for it in full. It isn't only a second transport, either. A pulled cluster exposes a scrape endpoint where a pushing one exports and exposes nothing, so it's a second collector configuration, a second exporter shape, a `ClusterRole` granting `services/proxy`, and a field with a status mirror and a printer column.

What makes dropping it easy is that deferring is free. The field would be optional and default to push, so adding it the day a cluster that genuinely can't egress turns up is additive and breaks nobody. So there's nothing to migrate later and no reason to build it now.

I've kept the analysis in the section rather than deleting it: what would be built, who declares the mode, and why Modelplane can't detect it for itself, since a composition function does no network probing. Pull direct stays ruled out on its own terms. The diagram now has both clusters pushing, and the API-change count drops from four to three.

design/metrics.md
dennis-upbound: Taking it, and it's cheaper than a nit. `TelemetryDestination`.

Two things push it past preference. The name was never really proposed: it's a placeholder a draft of the docs page had to invent to have something to write, on a question the doc explicitly leaves open, so there's nothing to reverse. And an OTLP endpoint carries metrics, logs and traces on the same wire, so the destination genuinely is signal-agnostic already. `MetricsDestination` would describe it narrower than it is, which makes the broader name more accurate rather than more speculative.

Also the asymmetry: the rename is free while nothing is built, and costs a deprecation window plus a migration for every user once something is.

I went with telemetry over observability since that's OTel's own word for the data, where observability is the property. And the answer holds whichever shape it lands in, so the doc now says `TelemetryDestination` as a kind or `spec.telemetry.destination` as a field, with the kind-or-field question still open.

design/metrics.md
```diff
@@ -0,0 +1,757 @@
...
+
+Name it `http` and not `metrics`, because it is the one serving port rather than a
+dedicated metrics one. The reason to go by name at all is prefill/decode: the decode
+engine serves on `_DECODE_ENGINE_PORT` (8001) because the pd-sidecar takes 8000, so
+matching 8000 by number scrapes the sidecar. By name, the scrape follows the engine on
+every pod, on every backend.
+
+**GPU utilization needs a source the stack doesn't install today.** `k8s_cluster` reports
+allocatable and requested `nvidia.com/gpu`, which answers how much of the fleet is claimed.
+Whether a claimed GPU is busy comes from DCGM, so `compose-serving-stack` installs the DCGM
```
haarchri: for AICR generated serving stack we Setup DCGM sounds like we should mirror to the handwritten too

design/metrics.md
```diff
@@ -0,0 +1,752 @@
...
+  token, queue depth, KV-cache occupancy. It's the metrics analogue of the OpenAI API
+  contract Modelplane already assumes for serving.
+- **Selection by a stamped label.** `ModelDeployment` gains `engines[].type`, and
+  Modelplane stamps `modelplane.ai/engine` from it. That label picks the `MetricMapping`.
+  The ML team already chose the engine in the image, so naming its kind touches nothing
+  about serving. It is a free-form string validated as a label value, not an enum: the
+  registry below is open to a mapping for a forked or unreleased engine, and an enum would
+  close the selector against the values that mapping needs to match. A value with no
+  mapping degrades to passthrough, which is the behaviour below rather than an error.
+- **A registry of first-class resources.** Each mapping is a `MetricMapping`, a Modelplane
```
haarchri: who is applying these CR for the default MetricMapping ?

design/metrics.md
```diff
@@ -0,0 +1,752 @@
...
+
+**The destination is the operator's, which is the point.** It sits where their
+observability already is, inside their network as often as not, so a cluster that can reach
+their backend needs no path anywhere else. Exporting direct also removes a hop that can
+fail and leaves a cluster's telemetry working while the control plane is upgrading.
+
+**Control-plane health comes from whoever runs the control plane, and Modelplane documents
+the path.** Crossplane's reconcile rates, function latency and the fleet scheduler's
+decisions are exactly what an operator wants when Modelplane itself misbehaves, and the
+constraint above means Modelplane cannot collect them. The path depends on how the control
```
haarchri: in crossplane ecosystem we have for this part https://github.com/crossplane-contrib/resource-state-metrics/tree/main/examples - so we could document RSM CRs for the healthy about the custom resources we have ...

docs/content/guides/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+
+The API line below is plain text rather than a `ref`, because the reference page is
+generated from a CRD that doesn't exist yet and Hugo fails a `ref` it can't resolve.
+
+[design]: https://github.com/modelplaneai/modelplane/pull/363
+{{< /hint >}}
+
+**API:** `modelplane.ai/v1alpha1` · TelemetryDestination, MetricMapping
+
+Modelplane collects metrics from everything it runs and sends them to a single
```
haarchri: are we planning to testing this with the e2e test flow ?

design/metrics.md
dennis-upbound: Good catch, and the failure mode turned out worse than a duplicate scrape. Since the Prometheus operator goes away entirely, a leftover `PodMonitor` is inert rather than double-scraping: it stops working quietly, which is harder to notice.

"Removing the Prometheus stack" now says that, and that the release note has to say the local store is going and where the series go instead.

design/metrics.md
dennis-upbound: Agreed, and the doc commits to a default now instead of calling the interval a knob: 15s for the engine and picker jobs, 30s for the substrate.

Your reason is the one written down. Queue depth and KV-cache occupancy move between scrapes, and a 30s sample smooths away the burst that caused the incident. A controller's health doesn't move that fast. The cost of 15s is in "Cost and cardinality" and it's affordable.

design/metrics.md
dennis-upbound: In scope, and it costs nothing extra. `k8s_cluster` reports the hydration Job (`k8s.job.failed_pods` and friends) and the claim's phase and requested size, so a cache that fails to stage shows up with no new component. Those PVC metrics are optional on that receiver, so the collector turns them on explicitly.

One limit is stated rather than hidden: how full the volume actually is comes from `kubeletstats`, which reads it off the kubelet and so belongs to the DaemonSet tier, which ships second. That matters less than it sounds, since a cache volume is written once and read after, so an undersized one fails at hydration rather than climbing into trouble later.

design/metrics.md
dennis-upbound: Both answered now, in "Cost and cardinality".

Series per pod: 359 metric lines from one vLLM 0.23.0 pod, measured on GKE. Per replica a cluster pays that plus its share of the EPP and DCGM's per-device series, at the interval above. A fifty-pod fleet is on the order of 40,000 series, around $640 a month, which is about one percent of the GPUs it watches.

Retention answered itself when the design changed: there's no per-cluster Prometheus any more. Every cluster exports straight to the fleet destination, so retention and storage are the operator's backend, configured once rather than per cluster.

design/metrics.md
dennis-upbound: You called this early and you were right. The field you commented on is gone, but the same argument decided the fleet resource: it's `TelemetryDestination`, not `MetricsDestination`, because an OTLP endpoint carries metrics, logs and traces on one wire.

It's already earned it. The AI gateway design's usage records are logs, and they land at that same destination.

design/metrics.md
dennis-upbound: Yes, deliberately, and the doc now gives the reason.

`http` rather than `metrics` because it's the one serving port that happens to also carry `/metrics`, not a dedicated metrics port, so naming it `metrics` would misdescribe it. The reason to go by name at all is prefill/decode: the decode engine serves on 8001 because the pd-sidecar takes 8000, so matching by number scrapes the sidecar instead of the engine.

design/metrics.md
dennis-upbound: Leftover, yes. The collector is an OpenTelemetry collector now and it replaces kube-prometheus-stack outright. "What replaces the stack" maps each thing that stack did onto a receiver, and discovery is a scrape config inside the collector rather than a PodMonitor per target.

design/metrics.md
dennis-upbound: This was the best question in the review and the doc had no answer to it. It does now, in "Cost and cardinality".

A fifty-pod fleet is on the order of 40,000 series, which is around $640 a month, against $40,000 to $125,000 of A100s. Telemetry is about one percent of the GPUs it watches. That's what makes always-on an easy trade rather than something I was asserting, and it's why there's no per-deployment opt-out: a `ModelDeployment` author shouldn't hold a toggle over telemetry the platform team consumes, and at one percent nobody needs one.

Your other half is unchanged and in the summary: no destination configured, no collector composed.

design/metrics.md
dennis-upbound: Yes. The Prometheus `compose-serving-stack` installs today goes, and the composed `OpenTelemetryCollector` takes its place. There's a "Removing the Prometheus stack" section now, because that's the one breaking change in this design and it should land last and on its own rather than alongside the additions.

design/metrics.md
dennis-upbound: Added. The `MetricMapping` YAML is now followed by exactly what it renders: receivers through exporters, with the `transform` block gated on the engine label and the `k8sattributes` processor that lifts that label onto the series.

design/metrics.md
dennis-upbound: Also a leftover, and gone for the reason you give. The collector's `prometheus` receiver does Kubernetes service discovery itself, so there's no CRD in the path, and with every cluster exporting to the destination there's no local store to keep either.

design/metrics.md
dennis-upbound: Added, and it ended up carrying the cost argument above.

Dropped before export: `pod`, `pod_uid`, `container_id`. The reason is sharper than tidiness. A billing backend counts a series as active for fifteen to thirty minutes after it stops receiving data, so every rolling update would otherwise mint a fresh set of per-pod series that stays billable long after the pod is gone. Never added: `caller`, which is unbounded by construction and belongs in the gateway's usage records rather than on a series.

One trap is written down too: `attributes`/`delete_key` removes the label but leaves the colliding series as separate undefined points. `metricstransform` with an aggregation action is the one that merges them.

design/metrics.md
dennis-upbound: Both covered now in "Exporters and destination".

Exporters: `otlp` over gRPC or `otlphttp`, and `prometheusremotewrite` for a Prometheus-compatible store. Vendor-specific exporters are out of scope; an operator who wants one puts their own collector in front of it, which is one configuration for the fleet rather than one per GPU cluster.

Auth was genuinely missing when you asked. It's now `auth.type` (`None`, `Basic`, `Bearer`) with a `secretRef` resolved in `modelplane-system`, the way `InferenceCluster` resolves a kubeconfig. The Secret propagates to each cluster through the same mechanism `ModelCache` already uses for a HuggingFace token.

design/metrics.md
dennis-upbound: Agreed, and that's what it is: one mode, push. It landed in the first rev after this comment.

Pull is recorded as ruled out rather than deleted, since it needs inbound exposure on every GPU cluster. And the roll-up question you raise is handled by every cluster exporting to the same destination under the same vocabulary, so the fleet view is one query there rather than a second hop through the middle.

design/metrics.md
dennis-upbound: Taken, and it's `TelemetryDestination`. Your reasoning is the reasoning in the doc: an OTLP endpoint carries metrics, logs and traces on the same wire, so a name scoped to metrics would describe it narrower than it is.

It paid off sooner than expected. The AI gateway design (upbound/inference#3) meters every request as a structured access log line, and those usage records now land at this same destination. They couldn't have if it were called `MetricsDestination`.

design/metrics.md
dennis-upbound: Taking this. You're right, and my reason for dropping DCGM was too narrow.

I'd removed it about four minutes before this comment landed, on the grounds that utilization is already covered by the engine's own queue depth and KV-cache occupancy. What I missed is that DCGM is the only source in this design for thermal throttling, ECC errors and XID faults, and those present as slow inference with entirely healthy-looking engine metrics. Nothing else catches them.

One thing I added while putting it back: installing it is conditional. `dcgm-exporter` runs `nv-hostengine` embedded, and a second host engine alongside an existing one conflicts rather than stacks, so one side stops reporting or crash-loops. Left unset, `compose-serving-stack` installs on a cluster Modelplane provisions, where nothing it installs brings an exporter, and skips an `Existing` one, where a GPU fleet usually already runs the GPU Operator's. `InferenceCluster.spec.gpuTelemetry` overrides both. The scrape is unconditional and collects whichever exporter is there.

design/metrics.md
dennis-upbound: Nobody, and the doc was claiming otherwise. Thank you, this was a real hole.

A Crossplane configuration package ships XRDs and compositions, not instances, so the package can't carry them. A composition can't apply one either, since that means writing to the control plane's own API server, which is the same reach problem this design already rules out for the collector. And the pattern I cited as precedent does the opposite: there are no built-in `InferenceClass`es, they're user-applied manifests that ship as docs examples.

So the built-in mappings live in `compose-serving-stack` as code, and they version with the package. `MetricMapping` is the extension point for an engine we don't ship, and a mapping selecting an engine already built in replaces it. The bullet is now "An extension point, not a registry", which is the smaller and more honest claim.

Related: two ship built in rather than three. vLLM and SGLang map cleanly; Triton/TRT-LLM resolves to "derived" or "waiting on newer metrics" in our own table, and shipping it would hand the first Triton user a mapping that doesn't map. It's documented as a `MetricMapping` to write, with the gaps named, which makes it the extension point's first real use.

design/metrics.md
dennis-upbound: Good pointer, and it landed on something bigger than the control-plane question.

`resource-state-metrics` is now cited in the doc as the way to turn XR conditions into series, with the same constraint as everything else there: it's a Deployment, so it runs where Crossplane does.

What it exposed: the fleet roll-up promised "degraded deployments" and nothing in the design produced it. A `ModelDeployment`'s degraded-ness is a condition on an XR the control plane holds, and this design collects from workload clusters. So the roll-up now counts replicas ready against desired, from `k8s_cluster`, with the gap named — a deployment the fleet scheduler never placed shows on the API, not in the roll-up. "Cost" came out for the same reason and is now GPU-hours, which matches what the AI gateway design says Modelplane publishes.

docs/content/guides/telemetry.md
dennis-upbound: Yes, and the harness already exists, which I hadn't said anywhere. There's a "Testing" section now.

`nix run .#e2e` brings up a control-plane cluster and a workload cluster registered with `source: Existing`, running a mock engine that answers the serving APIs the way a vLLM server does. Teaching that mock to serve a vLLM-shaped `/metrics` on a port named `http` turns it into the fixture this needs, with no cloud and no GPU.

That covers the whole claim: the collector composes on the workload cluster, the scrape finds the engine by `modelplane.ai/serving` and by port name, a `MetricMapping` renders into transform rules that rewrite `vllm:*` to `modelplane_*`, and a `TelemetryDestination` with its Secret propagates from the control plane and exports there. Pointing the destination at a collector inside the test makes the assertion a query. Two clusters is also what makes the fleet view testable rather than asserted.

What it can't cover is stated too: DCGM needs GPUs (and an `Existing` cluster skips it by default anyway), the `kubeletstats` tier, and any real engine's metric fidelity, which is what the GKE run in #412 is for.

dennis-upbound:
@negz your question about how an MD author sees and consumes these went unanswered until now, and answering it changed the design rather than restating it. It's in `## What to monitor`.

Both audiences read the same series. Every one carries `deployment`, `model`, `cluster` and `engine`, so an author's view is a filter on the shipped dashboard rather than a second pipeline, and Modelplane runs no author-facing store.

That's also the honest reason there's no toggle, which is the other half of what you asked. An author owns neither the destination, nor its cost, nor its retention, so a switch on their resource would govern none of the things that make collection a decision. What they own without asking anyone is on the API, where a `ModelDeployment`'s conditions say whether its replicas placed and are ready.

Your two blockers are settled in the doc. Collector is OpenTelemetry, with what replaces each piece of the stack, what we give up, and the Prometheus stack kept as a rejected alternative. Normalization is a `MetricMapping` per engine with the table of what maps and what doesn't; Triton and TensorRT-LLM don't map cleanly and the doc names that rather than papering it.

dennis-upbound:
Fair on the writing, and the fix is structural rather than a pass over the prose. I did several cleanup passes and they were all sentence-level, which left the real problem: 900 lines argued at one density, where nothing marks what needs your judgement against what's settled detail. If the skill is worth writing I'd read it.

On the mappings, you were right to be uneasy and there was a better idea I'd missed. Every engine already puts its own name on its metrics: `vllm:`, `sglang:`, `nv_trt_llm_`. A `MetricMapping` matches that prefix, so nothing is declared and no enum exists. A fork that kept vLLM's names keeps the mapping for free; one that renamed them writes a mapping against its own prefix, which it needed either way.

It only runs out on an engine whose names don't say what it is, like an OpenAI-compatible server publishing `http_requests_total`. `engines[].type` survives for that, out of the headline API changes and into the footnotes. Derived is the path, declared is the escape, which I think is what you were asking for.

Two things fell out. The rename is now a plain metric-name match, so `k8sattributes` leaves the rename path and the simpler metrics-transform processor is enough. And the doc had cited the GAIE model-server-protocol as precedent for reading a label; that protocol defines metric semantics and maps each engine's names but leaves engine detection unspecified, so the citation claimed more than the source said. Gone with the mechanism it justified.

On recording rules, you're right: OTTL runs per data point with no history, so there's no `rate(x[5m])` and no ratio across two metrics. I don't think it binds, because the design precomputes nothing. The roll-up is queries we ship, not a collector we run, and they evaluate at the destination where the full query language is there. Capacity sums, GPU-hours and the SLO ratio are all query-time.

What the collector does beyond rename is aggregation across dropped labels, which the cardinality story already depends on: dropping `pod` needs metrics-transform to merge the colliding series, or you get undefined points instead of a sum.

Where it would bite is pre-aggregating per cluster to cut egress, shipping a rate instead of raw samples. At roughly one percent of GPU spend we don't need to, and if that changed the `interval` processor is the partial answer rather than a redesign. Recording rules on every cluster were one of the things the Prometheus stack made us own, so losing them reads as a feature.

design/telemetry.md
```diff
@@ -0,0 +1,503 @@
...
+power and energy per device, and the fault taxonomy behind a drain: temperature, throttling,
+ECC and interconnect errors. The last two need collectors that DCGM leaves off by default,
+and Modelplane enables them, because on a multi-node gang one bad NVLink degrades an engine
+that otherwise looks healthy.
+
+**The substrate** reports whether the machinery works. The gang controller, LeaderWorkerSet
+or Grove, carries completeness in its status. That is the only place a half-placed gang
+shows up: the leader is running, so every per-pod view looks fine while the workers never
+scheduled. The DRA driver counts allocation failures, which is why a replica stays Pending.
+`kube-state-metrics` reads both controller statuses through its custom-resource-state
```
lsviben, line 144: can we do this with [resource-state-metrics](https://github.com/kubernetes-sigs/resource-state-metrics)?

design/telemetry.md
```diff
@@ -0,0 +1,503 @@
...
+**On each inference cluster**, a collector scrapes everything above that runs there: the
+gateway and its sidecar, the engines, the picker, DCGM, the controllers through
+`kube-state-metrics`, and ModelExpress. It renames what it scraped, merges each deployment's
+replicas into one series, stamps `cluster`, and exports OTLP to the control plane. It needs
+egress and nothing inbound.
+
+Targets come from the OpenTelemetry Operator's target allocator with
+`prometheusCR.enabled`, reading `PodMonitor` objects Modelplane composes. Its selectors
+match labels Modelplane stamps, because an unrestricted selector lets anyone who can create
+a monitor point the collector at a target of their choosing.
```
lsviben, line 207: do we really need the PodMonitors? We can point the `prometheusreceiver` to scrape the pods with the labels we want already.

design/telemetry.md
```diff
@@ -0,0 +1,503 @@
...
+
+The Prometheus receiver puts each pod's identity in resource attributes, which a metric
+processor cannot see. So `groupbyattrs` removes those attributes and merges the resources,
+and `aggregate_on_attributes` then combines the data points. Order matters, and so does the
+function: counters and counts are summed, while a ratio is averaged, since four replicas
+each at 0.5 are not a cache two hundred percent full.
+
+**On the control plane**, a collector receives from every cluster, scrapes Modelplane's own
+exporter and Crossplane's runtime metrics, and exports onward. It sees one merged stream, so
```
lsviben, line 216: what does "scrapes Modelplane's own exporter" here mean?

design/telemetry.md
```diff
@@ -0,0 +1,503 @@
...
+    prometheusremotewrite:
+      endpoint: https://prom.acme.example/api/v1/write
+```
+
+`spec.exporters` is the collector's exporters block, passed through unread. Modelplane
+validates that it parses and reports whether the destination is accepting writes; it does
+not model what an exporter is. So anything the collector supports works, with its auth, TLS,
+retry and queue settings intact. A field-by-field schema would have had to restate all of
+that or cap it. This one keeps working when the collector gains an exporter Modelplane has
+never heard of.
```
lsviben, line 243: some `exporters` use can use auth `extensions` . Maybe not in the first pass but we should consider also allowing setting `extensions` as well.

But in terms of keeping this simpler, we could always keep the fleet exporter simple and ask folks to create their own OtelCollector in the control plane cluster, with all the auth/extra processing etc and have our fleet one just send to it locally.

design/telemetry.md
```diff
@@ -0,0 +1,503 @@
...
+host and nothing about the workload, so nothing joins a GPU to a replica. Nothing times a
+replica from created to serving, or says whether its weights were staged before it started.
+And only the control plane knows whether it can still reach a cluster.
+
+Modelplane either made those decisions or holds the object that answers them. So an exporter
+beside the composition functions reads the `ResourceSlices` and the replicas' claims, and
+publishes every capacity and cost series in the appendix. It is the only component here
+Modelplane writes; both collectors are upstream.
+
+### Normalizing
```
lsviben, line 163: what happens to the metrics read but not normalized to `modelplane_*` ?

Do we want to be opinionated and just send the `modelplane` metrics forward and filter out the rest? Maybe that should be the default, and some advanced toggle for folks who want more.

design/telemetry.md
dennis-upbound: Yes — taking it. That removes the only component this design was adding.

Everything I'd written the exporter for is already a field on an object we reconcile, so a `ResourceMetricsMonitor` covers it.

One catch that changed the metric surface: RSM reports current state and doesn't accumulate, so there's no GPU-seconds counter. Modelplane records the allocation timestamp and the backend multiplies. The appendix was also claiming a histogram and two counters from that source — gauges now.

design/telemetry.md
dennis-upbound: Agreed, and I'd argued your side earlier in this thread before contradicting myself: further up I said the `prometheus` receiver does Kubernetes SD itself and the CRD was a consequence of choosing Prometheus. Then I reintroduced the target allocator while fixing a different gap.

Gone. It's a scrape config matching `modelplane.ai/serving`, port by name. That also drops the caveat I'd had to write about restricting monitor selectors.

Migration note changes with it — a leftover `PodMonitor` now stops being read by anything, quietly.

design/telemetry.md
dennis-upbound: Vague on my part, and moot now. It meant a small exporter I was proposing to write. Per your other comment that's `resource-state-metrics`, so the line names RSM and Crossplane's runtime metrics instead.

design/telemetry.md
dennis-upbound: Took the first half — `spec.extensions` beside `spec.exporters`, both passed through unread. It was worse than missing: the example referenced `authenticator: bearertokenauth` without defining it, so it wouldn't have started. Credentials go in a Secret via `secretRef`, mounted as env vars.

On the second half, passthrough gets both. Anyone wanting heavy processing still points our exporter at their own collector; passthrough just means a bearer token doesn't require standing one up first.

You'd be right the moment we start adding fields that *model* exporters rather than carry them. Worth watching for.

design/telemetry.md
dennis-upbound: Real gap. The doc was silent, and inconsistent since the Prometheus version, which said engine series "stay in the cluster's Prometheus" — there's no cluster store now.

Default is in: only `modelplane_*` leaves. An unrenamed series is one whose meaning we can't vouch for across engines, and the cardinality argument applies to it in full.

Toggle is `passthrough: true` on the `MetricMapping`, per-engine, since whoever wants raw `vllm:*` is debugging that engine rather than watching a fleet.

design/telemetry.md
```diff
@@ -0,0 +1,663 @@
...
+capacity has no series at all. DCGM labels a GPU with its UUID and its host and nothing
+about the workload, so nothing joins a GPU to a replica. Nothing times a replica from
+created to serving. And only the control plane knows whether it can still reach a cluster.
+
+Every one of those answers is already a field on an object Modelplane reconciles, so
+nothing has to be written to observe them. `resource-state-metrics` turns custom resource
+fields into series from a `ResourceMetricsMonitor`, and the control-plane collector
+scrapes it. A `ModelReplica` carries the GPUs it holds, the devices its claim resolved to,
+and when it was allocated and when it became ready; an `InferenceCluster` carries
+reachability and the allocatable count read off the slices.
```
haarchri, line 175: What does `allocatable` mean for partitionable devices (MIG, DRA partitions)? Is it whole GPUs, partitions, or something else?

design/telemetry.md
```diff
@@ -0,0 +1,663 @@
...
+
+OTTL is the stable, well-documented part of the collector, and renaming, scaling a unit and
+setting a label are what it is for. Each engine prefixes its metrics with its own name, so
+nothing declares which engine a deployment runs. A fork that kept vLLM's names is handled by
+vLLM's statements without knowing it is a fork.
+
+Only `modelplane_*` leaves a cluster. A series the statements did not rename is a series
+whose meaning Modelplane cannot vouch for across engines, and the cardinality argument below
+applies to it in full, so the cluster collector drops it after the `transform` stage. An
+operator who wants an engine's raw names too sets `passthrough: true` on the
```
haarchri: vLLM built-in statements are not a `MetricMapping`, so how does someone enable passthrough for vLLM? Do passthrough series skip the replica merge and carry pod identity with them?

design/telemetry.md
```diff
@@ -0,0 +1,663 @@
...
+has moved to 8001, so the scrape would find the sidecar and report it as the engine. No
+`PodMonitor` and no Prometheus operator: the CRD was a consequence of having chosen
+Prometheus, and choosing a collector instead removes it from the path. Discovery is a scrape
+config in the collector's configuration, which Modelplane composes with everything else in
+it.
+
+The Prometheus receiver puts each pod's identity in resource attributes, which a metric
+processor cannot see. So `groupbyattrs` removes those attributes and merges the resources,
+and `aggregate_on_attributes` then combines the data points. Order matters, and so does the
+function: counters and counts are summed, while a ratio is averaged, since four replicas
```
haarchri, line 270: KV utilization hide the one hot replica that is preempting? Should saturation gauges export max, or keep a bounded replica identity ?

Issue #179, opened 2026-06-17
Make the EndpointPicker (EPP) config user-configurable for PrefillDecode serving


haarchri:
### What problem are you facing?
`serving.mode: PrefillDecode` composes an EndpointPicker (EPP) whose routing policy is a hardcoded `EndpointPickerConfig` baked into the composition function [routing.py](functions/compose-model-replica/function/routing.py#L37-L60). Today every disaggregated replica gets the exact same plugin set, scorer weights, and decider parameters, with no way for a user to influence them.

This is increasingly limiting, and PR #175 makes the case concrete:

- The config is fragile and version-coupled. PR #175 shows the EPP config is finely tuned to a specific image (llm-d-inference-scheduler v0.8.0 / GIE v1.5.0), approx-prefix-cache-producer with `autoTune: false`, `nonCachedTokens: 16`, specific scorer weights. 
A user on a different scheduler version, or who wants different behavior, has no escape hatch short of forking the function.
- It is coupled to engine settings the user owns. `blockSizeTokens: 16` only works because it matches vLLM's default KV block size. A user who sets `--block-size 32` on their engine (engine flags are the user's, per #137) silently degrades prefix-cache routing with no error and no knob to correct it.
- Legitimate workloads want different policies. The routing trade-offs are workload-dependent and not one-size-fits-all:
- tuning scorer weights (prefix-cache vs queue/load) for chat (long shared prefixes) vs bursty short-prompt traffic
- swapping max-score-picker for a weighted-random picker to spread load
- choosing a different decider (always-disagg-pd-decider vs the selective prefix-based-pd-decider PR #175 explicitly keeps the former "in mind as a fallback")
- adopting new EPP plugins from a newer scheduler release without a Modelplane code change

Right now any of these requires patching and rebuilding the composition function.

### How could Modelplane help solve your problem?
Make the EPP EndpointPickerConfig user-overridable, while keeping the current hardcoded value as the default so existing replicas are unaffected.

I would suggest a tiered approach (and I think a combination is the right end state):

A. specTemplate-style raw override. An optional field on the serving block that lets a user supply their own `EndpointPickerConfig` (inline or via ConfigMap ref); fall back to the current default when unset.
```
spec:
  serving:
    mode: PrefillDecode
    endpointPicker:
      configTemplate: |          # raw EndpointPickerConfig, used verbatim
        apiVersion: inference.networking.x-k8s.io/v1alpha1
        kind: EndpointPickerConfig
[...]

Issue #71, opened 2026-05-11
Add routing affinity to ModelService for fleet-level cache locality


dennis-upbound:
`ModelService` routes weighted across `ModelEndpoint`s today. With replicas of one deployment landing on multiple clusters, the same multi-turn chat conversation can end up on a different cluster's replica each turn — cold KV cache every time. Turns 2-N pay full prefill cost (~800ms TTFT) when they could be hitting a warm cache (~150ms).

The fleet has the right information to fix this — Modelplane sees every `ModelEndpoint` across every cluster. It just doesn't bias requests toward cache locality today. This issue proposes adding routing affinity at the cluster granularity: same session or same prompt → same cluster. Per-replica KV-cache-aware routing inside each cluster is already llm-d's and Dynamo's job (per #65) and stays there.

## What we want to get done

Two affinity modes on `ModelService`:

**Sticky session.** A header value routes the request to the same cluster every time. Useful when clients carry conversation IDs.

**Prefix-hash.** No header required from the client. A body-hasher computes a stable hash over the first N bytes of the prompt; the gateway consistent-hashes on that. Useful for OpenAI-API-compatible clients that don't pass session IDs.

```yaml
apiVersion: modelplane.ai/v1alpha1
kind: ModelService
metadata:
  name: kimi-k2
spec:
  endpoints:
  - selector:
      matchLabels:
        modelplane.ai/deployment: kimi-k2
  affinity:
    type: PrefixHash         # PrefixHash | Session | None
    prefixBytes: 1024        # bytes hashed for PrefixHash
    sessionHeader: X-Session-ID   # used for Session
    fallback: NextEndpoint
```

## Cluster granularity, not replica granularity

The fleet gateway routes to clusters for locality. Within a cluster, the per-engine router (llm-d's inference scheduler for llm-d backends, Dynamo's frontend for Dynamo, per #65) handles the KV-cache-aware replica pick. We don't duplicate that work cross-cluster, we don't need cross-cluster cache-state tracking, and we don't tokenize at the fleet level.

The fleet gateway hashes a stable input and forwards. Same prompt body → same hash → same cluster → that cluster's router serves with a warm cache. If the cluster's cache is cold on the first request, it builds up naturally as subsequent matching requests reinforce.

## How it composes

The composition function for `ModelService` emits Envoy resources behind whatever InferenceGateway is configured:

[...]

PR #364, opened 2026-07-21
Design: inference-aware fleet routing

dennis-upbound:

A design doc, not code yet — opening as a draft for direction before implementation.

The control plane's fleet gateway routes by path and static weight only, so it's cache-blind: a multi-turn chat can land on a different cluster each turn and pay full prefill every time (~800ms TTFT vs ~150ms warm), and the fleet can't express request priority. The per-cluster EPP is already inference-aware; the fleet layer isn't.

This proposes cache-locality affinity (Session / ClientIP) and criticality on the fleet gateway, **composed against the Gateway API standard** — `SessionPersistence` (GEP-1619) and GAIE (`InferencePool`/`InferenceModel`) — rather than any one gateway's own resources, so a conformant gateway works with one codepath instead of a per-vendor fork. The fleet routes to a cluster and delegates the replica pick to that cluster's EPP (move requests, not KV state). Covers #71 and #8.

Two things I'd flag for judgment:
- **The one off-standard dependency** is per-backendRef `URLRewrite` (the reason we're on Traefik, #85/#89 — it's beyond the Gateway API spec, which only allows `URLRewrite` at the rule level). The doc proposes relocating it (deployment-scoped replica paths, GAIE canary weighting, a per-provider adapter) so the fleet gateway needs only the standard. Honest cost: Traefik handles provider path/TLS for free today.
- **The two-layer shape** (fleet gateway + per-cluster gateway/EPP) may not be the end state; the doc flags collapsing a layer as an open question rather than locking it in.

The standard is treated as a baseline with clean extension seams (the `backend` discriminator, the configurable EPP in #179, the ext_proc chain) so richer routing can attach without forking core.

`design/routing.md` only; no code.

I have:

- [ ] ~Run `nix flake check` (or `./nix.sh flake check`) and made sure it passes.~ (design doc only, no code paths)
- [ ] ~Added or updated tests covering any composition function changes.~ (design doc only)

### DIFF
--- a/design/routing.md
+++ b/design/routing.md
@@ -36,6 +36,17 @@ composes these against the Gateway API standard and its inference extension,
 rather than any one gateway's own resources, so a conformant gateway works without
 Modelplane carrying a codepath per vendor.
 
+Concretely, this proposes:
+
+1. an `affinity` block on `ModelService` (Session or ClientIP)
+2. criticality and weighting expressed through GAIE, replacing the static HTTPRoute
+   weights `compose-model-service` emits today
+3. moving the fleet gateway from Traefik to the standard profile (Envoy Gateway with
+   GAIE and Envoy AI Gateway), with heterogeneous path rewrites handled by an ext_proc
+
+Approving this means agreeing to that direction and that gateway profile. The YAML
+below illustrates the shapes. It isn't the committed field spec.
+
 ## The user model: ModelService and ModelEndpoint
 
 A user works with two resources.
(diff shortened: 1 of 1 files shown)


design/routing.md
```diff
@@ -0,0 +1,246 @@
...
+rather than sit in a CR. **This document does not propose building it.** The
+adapter belongs with a provider-auth design, not with fleet routing. Until it
+exists, heterogeneous-path provider aggregation stays a Traefik-backend concern,
+and the standard, inference-aware backend serves self-hosted fleets.
+
+No single gateway meets the whole profile today. Envoy Gateway is closest (GAIE
+and `SessionPersistence`) once the self-hosted normalization above is in place.
+Traefik keeps provider aggregation, but not the profile. As the standard matures,
+more gateways qualify with no new Modelplane code. That is the point of composing
+to the standard: the supported set grows without the design diverging.
```
negz: Why does supported set matter? Do we want to let folks bring their own preferred GAIE implementation? If so why?

design/routing.md
```diff
@@ -0,0 +1,246 @@
...
+The remaining case is folding an external provider into the same weighted split,
+where the provider's `/openai/v1/` differs from the self-hosted `/v1/`.
+Normalizing that off the gateway would take a **per-provider adapter**: a small
+proxy that presents the canonical path and injects the provider's API key. That is
+more than a path rewrite, because Modelplane doesn't inject provider auth at all
+today (an endpoint just points at the FQDN), so the key has to come from a Secret
+rather than sit in a CR. **This document does not propose building it.** The
+adapter belongs with a provider-auth design, not with fleet routing. Until it
+exists, heterogeneous-path provider aggregation stays a Traefik-backend concern,
+and the standard, inference-aware backend serves self-hosted fleets.
```
negz: I think if this doc proposes moving the InferenceGateway implementation back to a GAIE implementation, it needs to propose a solution for heterogenous rewrites. I'm open to a component that does the rewrite for us if you think it's the best path.

dennis-upbound: Agreed it needed a real answer — added one, in a new "Heterogeneous rewrites without the beyond-spec feature" section. Two cases:

- **Self-hosted per-endpoint paths** move to an ext_proc rewrite (`EnvoyExtensionPolicy`/`extProc`), which rides GAIE's EPP since that's already an ext_proc in the request path — or normalize to one deployment-scoped path so a single rule-level rewrite covers it.
- **External providers** are handled by Envoy AI Gateway's `AIServiceBackend` + `BackendSecurityPolicy` (path/schema translation + credential injection). That's a built-in feature of the gateway we're adopting, not a Modelplane-built adapter, so it also resolves the provider-auth piece the doc used to punt.

Both put the rewrite in an ext_proc rather than a beyond-spec gateway feature, which is what lets the standard profile replace Traefik. (Also reworked the doc around `ModelService` — that reframe hadn't actually been pushed before.)

design/routing.md
dennis-upbound: Took a stance: we target one gateway (Envoy Gateway + GAIE + Envoy AI Gateway), not arbitrary implementations. Composing to the standard is about getting off Traefik's beyond-spec behavior and onto the inference-aware features — not BYO. The `InferenceGateway` `backend` discriminator just stays as the seam if a platform team ever needs a different conformant gateway; breadth of support is a consequence, not the goal. Reworded the "supported set grows" line into a new "Which gateway" section.

negz:
@dennis-upbound I'm starting to think we probably need an AI gateway, not GAIE, at the InferenceGateway level. Something like Envoy AI Gateway, or Agent Gateway. I'm doing a bit of research into this myself.

PR #360, opened 2026-07-18
feat(docs): Serve the Anthropic Messages API (Claude Code) end to end

haarchri:
### Description of your changes
Enables the Anthropic Messages API (`/v1/messages`) to work through Modelplane's serving path, and adds a validated, copy-paste example showing how to deploy a model on it and drive it from Claude Code via `ANTHROPIC_BASE_URL`.

Why bump the endpoint-picker image:
Even for Unified single-pod serving, a replica is fronted by a GAIE `InferencePool` + endpoint picker (EPP), which sits in the request path as an Envoy ext_proc and parses the request body to score endpoints. At the pinned `llm-d-inference-scheduler:v0.8.0` (gateway-api-inference-extension v1.5.0) the EPP's body parser is OpenAI-only: an Anthropic Messages request (`messages` + `max_tokens`, no `prompt`) is dispatched to the completions parser and rejected **before it reaches vLLM**:

```
inference error: BadRequest - invalid completions request: must have prompt field at gateway-api-inference-extension@v1.5.0/pkg/epp/handlers/server.go:311
```

vLLM itself serves `/v1/messages` fine, the picker was the blocker...

`llm-d-inference-scheduler` **v0.9.0** adds a native Anthropic parser (`/v1/messages`, tools, `messages/count_tokens`) and enables all HTTP parsers by default, so Anthropic-shaped requests route through the picker. v0.9.0 also renamed the images (the old `llm-d-inference-scheduler` / `llm-d-routing-sidecar` packages stop at v0.8.0); the scheduler/sidecar now ship as `llm-d-router-endpoint-picker` / `llm-d-router-disagg-sidecar`. ref: https://github.com/llm-d/llm-d-router/pull/1508 https://github.com/llm-d/llm-d-router/pull/1554  

updated the APIGroup from inference.networking.x-k8s.io APIGroup to llm-d.ai APIGroup for `EndpointPickerConfig` https://github.com/llm-d/llm-d-router/pull/972

```
{"level":"info","ts":1784399821.526125,"caller":"loader/configloader.go:89","msg":"DEPRECATION: apiVersion inference.networking.x-k8s.io/v1alpha1/EndpointPickerConfig is deprecated","replacement":"llm-d.ai/v1alpha1/EndpointPickerConfig"}
```

Fixes #283

I have: 

- [x] Read and followed Modelplane's [contribution process](https://github.com/modelplaneai/modelplane/blob/main/CONTRIBUTING.md).
[...]

PR #365, opened 2026-07-21
Add a local two-cluster end-to-end test (no cloud, no GPU)

ytsarev:

Modelplane has no end-to-end test in the repo. Unit tests cover what each composition function renders, but nothing checks that the composed resources actually reconcile and serve on a real cluster. The one path that exercises the full flow (publish capacity, register a cluster, deploy a model, serve a request) is getting started, which provisions EKS/GKE/Nebius and needs GPUs, so it can't run on a laptop or gate CI without a cloud account and accelerators. The failures that only surface at runtime (missing RBAC, cross-cluster routing, provider readiness, teardown ordering) have had nothing to catch them.

This adds `nix run .#e2e`: the whole path on two local `kind` clusters.

- **No cloud, no GPU, no credentials.** The `InferenceCluster` registers the workload cluster from a kubeconfig Secret (`source: Existing`) instead of provisioning EKS/GKE/Nebius; the `InferenceClass` advertises a fake `gpu.example.com` device (`claim: DRA`, vendored dra-example-driver) that binds on a GPU-less node; and the `ModelDeployment` runs a mock server in place of a real engine, answering both the OpenAI and Anthropic message APIs, so the pod goes Ready without a real model.
- **Dev/CI parity.** `nix run .#e2e -- --verify` brings it up, waits for the ModelService, and asserts a live 200. You run it locally; CI runs the identical command. The workflow is a thin wrapper (disk, cache, diagnostics, teardown) with no test logic of its own.
- **Label-gated.** Behind a `test-e2e` PR label, not every push (a run is 15-20 minutes). No secrets, so it can gate a PR on its own.
- **Real orchestration.** Two clusters, because the control-plane and workload gateways both install the Gateway API CRDs and collide on one; the full ModelDeployment, ModelReplica, ModelEndpoint, ModelService chain; a DRA ResourceClaim actually bound on the workload cluster.

`nix run .#e2e -- --verify` brings both clusters up, asserts the endpoint serves (it curls from a pod on the control plane and checks the status), and leaves the whole platform queryable, no cloud in sight:

```console
$ nix run .#e2e -- --verify
...
==> ModelService address: http://172.19.255.200/ml-team/mock
==> verify attempt 1 (OpenAI): HTTP 200
==> verify (Anthropic /v1/messages): HTTP 200
==> End to end OK: http://172.19.255.200/ml-team/mock serves OpenAI (/v1/chat/completions) and Anthropic (/v1/messages)

$ kubectl get inferencegateway,inferencecluster,inferenceclass
NAME                                     GATEWAY   ADDRESS          SYNCED   READY
inferencegateway.modelplane.ai/default   Traefik   172.19.255.200   True     True
NAME                                   SOURCE     GATEWAY          SYNCED   READY
inferencecluster.modelplane.ai/local   Existing   172.19.255.100   True     True
NAME                                         SYNCED   READY
inferenceclass.modelplane.ai/synthetic-gpu   True     True

$ kubectl -n ml-team get modeldeployment,modelservice
NAME                                      REPLICAS   SYNCED   READY
modeldeployment.modelplane.ai/mock-demo   1          True     True
NAME                              ADDRESS                              SYNCED   READY
modelservice.modelplane.ai/mock   http://172.19.255.200/ml-team/mock   True     True
```

**Scope.** Stops short of real token generation (mock engine), GPU drivers/CUDA (fake DRA), cloud provisioning, and multi-node serving. `e2e/README.md` has the design and a tested/not-tested table; CONTRIBUTING gains a note placing the two layers (unit via `nix flake check`, integration via `nix run .#e2e`).

**Alternatives ruled out.**

- A declarative test runner like Chainsaw for the assertions: the hard part is the two-cluster bootstrap, which is irreducible shell a runner would only wrap, so it wasn't worth a third testing idiom for one scenario.
- A small real model on CPU: exercises the inference engine, not Modelplane's orchestration, at a large cost in image size, weights, and runtime.

`nix run .#e2e -- --verify` passes both locally and in CI (green run: https://github.com/ytsarev/modelplane-public/actions/runs/30901652231).

I have: 

- [ ] ~~Added or updated tests covering any composition function changes.~~ (no `functions/` changes)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,99 @@
+name: E2E
+
+# The local two-cluster e2e in e2e/ is heavy — it stands up two kind clusters
+# and the full serving stack (cert-manager, Envoy Gateway/AI Gateway, GAIE,
+# kube-prometheus-stack, NFD, DRA) and waits for it to reconcile, ~15-20 min. The
+# time is that install and reconcile, not model weights (the engine is a mock).
+# So it doesn't run on every push: add the `test-e2e` label to a PR to run it (it
+# re-runs on each push while the label is set), or trigger it manually. Unlike
+# the package push, it needs no cloud or registry credentials.
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+  workflow_dispatch: {}
+
+# Cancel in-progress runs for the same PR. Runs on main are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    # Run on manual dispatch, when the test-e2e label is added, or on a push
+    # (synchronize/opened/reopened) while the PR carries it. Scoping the
+    # `labeled` event to test-e2e keeps adding any *other* label from firing a
+    # fresh ~20 min run.
+    if: >-
+      github.event_name == 'workflow_dispatch'
+      || (github.event.action == 'labeled' && github.event.label.name == 'test-e2e')
+      || (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'test-e2e'))
+    runs-on: ubuntu-24.04
+    # Generous: a cold Nix cache (the Actions cache throttles) rebuilds the
+    # function images, on top of the two-cluster bring-up and serving-stack
+    # install. 45m was too tight on a cold run.
+    timeout-minutes: 60
+
+    steps:
+      # Two kind clusters and the serving stack need more disk than the runner
+      # has free by default, the same reason the push-package job reclaims space.
+      - name: Cleanup Disk
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+        with:
+          android: true
+          dotnet: true
+          haskell: true
+          tool-cache: true
+          swap-storage: false
+          large-packages: false
+          docker-images: false
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: log-lines = 500
+
+      # See ci.yml for why use-flakehub is off.
+      - name: Cache Nix store
+        uses: DeterminateSystems/magic-nix-cache-action@908b263ff629f4cc17666315b7fd3ec127c6244d # v14
+        with:
+          use-flakehub: false
+
+      # The exact command a developer runs locally (--print-build-logs is only CI
+      # verbosity), so a green run here and a green run on a laptop mean the same
+      # thing. It brings up both clusters, deploys the mock model, waits for the
+      # ModelService, and asserts a live 200 — exiting non-zero on any failure.
+      - name: Run e2e
+        run: nix run .#e2e --print-build-logs -- --verify
+
+      # On failure, capture what the clusters looked like before teardown.
+      # `kind export logs` dumps every pod's logs (crossplane, providers, the
+      # serving stack); the status snapshots make triage quick.
+      - name: Collect diagnostics
+        if: failure()
+        run: |
+          mkdir -p e2e-logs
+          nix develop --command bash -c '
+            kind export logs e2e-logs/control-plane --name modelplane-e2e-local || true
+            kind export logs e2e-logs/workload --name modelplane-e2e-workload || true
+            kubectl --context kind-modelplane-e2e-local get composite -o wide > e2e-logs/composites.txt 2>&1 || true
+            kubectl --context kind-modelplane-e2e-local -n ml-team get modeldeployment,modelreplica,modelendpoint,modelservice -o yaml > e2e-logs/model.yaml 2>&1 || true
+          '
+
+      - name: Upload diagnostics
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-diagnostics
+          path: e2e-logs
+          if-no-files-found: ignore
+
+      # Always tear both clusters down, even on failure or cancellation.
+      - name: Teardown
+        if: always()
+        run: nix run .#e2e -- --clean
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,16 @@ nix flake check            # or: ./nix.sh flake check
 `nix run .#fix` auto-fixes most lint and formatting issues. Run it before
 opening a PR.
 
+`nix flake check` is the unit layer — fast and sandboxed, proving each
+composition function renders the right resources. The integration layer is
+`nix run .#e2e`, which brings up two local `kind` clusters and runs the
+whole path — scheduling, the serving-stack install on a registered cluster,
+gateway routing, a live request — with no cloud credentials. Add `-- --verify`
+and it waits for readiness, asserts a 200, and exits non-zero on failure. That
+verify command is what the label-gated `E2E` workflow runs on CI (add the
+`test-e2e` label to a PR), so a green local `--verify` and a green CI run mean
+the same thing. See `e2e/README.md`.
+
 ## Submitting changes
 
 Before opening a PR, run `nix flake check` and make sure it passes. If you
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,186 @@
+# Local end-to-end test (no cloud, no GPU)
+
+Exercise the full Modelplane path — publish capacity, register a cluster, deploy
+a model, route a request through the control-plane gateway — on local `kind`
+clusters, with **no cloud provider and no GPU**.
+
+This is the integration layer. The composition functions run in-cluster against
+real providers, so it catches failures unit tests can't: missing RBAC,
+cross-cluster routing, CLI parsing, provider readiness gating, and teardown
+ordering. It needs no cloud or registry credentials, only public images — so it
+can gate a merge where the cloud e2e can't.
+
+It uses **two clusters**, mirroring a real deployment:
+
+- a **control-plane** cluster (crossplane + the Configuration + the
+  `InferenceGateway`), managed by `crossplane project run`;
+- a **workload** cluster registered via `source: Existing`, where the serving
+  stack and the model run.
+
+Two clusters rather than one because the control-plane `InferenceGateway`
+(Traefik) and the workload `ServingStack` (Envoy) both install the Gateway API
+CRDs — co-located on one cluster they race for the same cluster-scoped CRDs and
+the gateway wedges (`encountered composed resource without required
+composition-resource-name annotation`). Separate clusters, as in production,
+avoid it.
+
+Two Modelplane primitives make it cloud-free:
+
+- **`source: Existing`** — the `InferenceCluster` registers a bring-your-own
+  cluster (the workload kind cluster) via a kubeconfig Secret instead of
+  provisioning EKS/GKE/Nebius.
+- **`claim: DRA` + a fake GPU driver** — the `InferenceClass` advertises a
+  `gpu.example.com` device backed by the **dra-example-driver**
+  (kubernetes-sigs), which publishes fake GPUs with **no real hardware**. The
+  engine's `ResourceClaim` binds a fake device on a GPU-less node, so the *real*
+  DRA allocation path runs. (A claimable device is required: the fleet scheduler
+  rejects an engine whose only device is `Synthetic`.)
+
+The engine is a **mock server** (a few lines of Python) that answers both
+`/v1/chat/completions` (OpenAI) and `/v1/messages` (Anthropic), the way a vLLM
+server exposes both, so the pod goes Ready without a real model or GPU.
+
+## What this does and does not test
+
+| Tested | Not tested |
+|---|---|
+| Fleet scheduling / placement (CEL vs declared capacity) | Real token generation (mock engine) |
+| `ModelDeployment` → `ModelReplica` → `ModelEndpoint` → `ModelService` wiring | Real GPU drivers / CUDA (fake DRA devices only) |
+| DRA `ResourceClaim` → fake device binding (the real allocation path) | Multi-node / disaggregated (`PrefillDecode`) serving |
+| Serving-stack install on a real (BYO) workload cluster | Cloud provisioning (EKS/GKE/Nebius) |
+| Control-plane `InferenceGateway` + cross-cluster routing to the replica | |
+| Status propagation and foreground-deletion ordering | |
+
+## Prerequisites
+
+- **Docker** with real headroom — **≥ 16 GB memory** and **plenty of disk**
+  (raise Docker Desktop's disk-image size). Two kind clusters, 13 provider
+  packages, the built function images, and the serving stack
+  (kube-prometheus-stack et al.) add up fast; a full Docker disk surfaces as
+  `no space left on device`. Reclaim between runs with `docker builder prune -af`
+  and `docker image prune -af`.
+- Everything else — `kind`, `kubectl`, `curl`, `git`, and the `crossplane` CLI —
+  is provided by the flake via `nix run .#e2e`.
+
+The workload cluster is pinned to **k8s v1.34** (in `run.sh`) for the
+`resource.k8s.io` (DRA) APIs, on-by-default in 1.34: both the serving stack's
+NVIDIA DRA driver and the dra-example-driver register DeviceClasses, and the
+example driver publishes the `ResourceSlice`s the engine's `ResourceClaim` binds
+against. The control-plane cluster needs no DRA.
+
+## Run
+
+```bash
+nix run .#e2e              # bring up both clusters + deploy the mock model
+nix run .#e2e -- --verify  # same, then wait for readiness and assert a live 200
+nix run .#e2e -- --clean   # tear both clusters down
+```
+
+`crossplane project run` installs the config and applies the resources, then
+returns; the serving-stack install and model rollout reconcile in the background.
+So wait for the `ModelService` to publish an address before curling. That address
+is on the kind Docker subnet, which the host can't route to on macOS, so curl
+from a pod on the control plane:
+
+```bash
+kubectl -n ml-team get ms mock -w          # wait for ADDRESS to appear
+addr=$(kubectl -n ml-team get ms mock -o jsonpath='{.status.address}')
+
+# OpenAI Chat Completions
+kubectl run curl -n ml-team --rm -it --image=curlimages/curl@sha256:7c12af72ceb38b7432ab85e1a265cff6ae58e06f95539d539b654f2cfa64bb13 -- \
+  curl -s "$addr/v1/chat/completions" -H 'content-type: application/json' \
+  -d '{"model":"mock","messages":[{"role":"user","content":"hi"}]}'
+
+# Anthropic Messages API (same address; vLLM and the mock serve both)
+kubectl run curl -n ml-team --rm -it --image=curlimages/curl@sha256:7c12af72ceb38b7432ab85e1a265cff6ae58e06f95539d539b654f2cfa64bb13 -- \
+  curl -s "$addr/v1/messages" -H 'content-type: application/json' -H 'anthropic-version: 2023-06-01' \
+  -d '{"model":"mock","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
+```
+
+`--verify` runs both of those (OpenAI then Anthropic) and exits non-zero on
+failure. It's the exact command the `E2E` CI workflow runs, so a green `--verify`
+locally and a green CI run mean the same thing; use the manual curls above to
+poke the endpoints interactively.
+
+## How it's structured
+
+`nix run .#e2e` materialises the Nix-built function images and hands off to
+`run.sh`, which does the cross-cluster orchestration that `crossplane project
+run` flags can't express:
+
+1. Create the **workload** kind cluster (pinned v1.34).
+2. Install MetalLB on it (the serving stack doesn't) with a pool inside the
+   detected kind subnet and disjoint from the InferenceGateway's, install the
+   **dra-example-driver** (fake GPUs), and label its node for the `gpu-synthetic`
+   pool.
+3. `crossplane project run` for the **control plane**, with
+   `lean-control-plane.yaml` as `--init-resources` so the provider trims land
+   before the providers install.
+4. Finish the setup the getting-started flow does by hand (as the nix run app
+   does since #375): `kubectl apply` the RBAC prerequisites, point provider-helm
+   at its DeploymentRuntimeConfig, add the workload kubeconfig Secret (`kind get
+   kubeconfig --internal`, reachable from control-plane pods over the shared kind
+   network), then apply the subnet-templated Modelplane manifests.
+
+Everything the control plane needs is a declarative manifest; the shell in
+`run.sh` is only the irreducible cross-cluster setup (a second cluster, its
+MetalLB and DRA driver, the cross-cluster kubeconfig).
+
+```
+e2e/
+  run.sh                     # two-cluster orchestration
+  dra-example-driver.yaml    # vendored fake DRA GPU driver (applied to workload)
+  manifests/                 # applied to the control plane after setup
+    00-namespaces.yaml
+    10-inference-gateway.yaml
+    20-inference-class.yaml
+    30-inference-cluster.yaml # source: Existing -> the workload cluster
+    40-model-deployment.yaml
+    50-model-service.yaml
+```
+
+## Why the extra moving parts
+
+- **MetalLB on both clusters.** Both gateways — control-plane Traefik and the
+  workload Envoy Gateway (whose readiness the serving stack gates on,
+  `_GATEWAY_READY_CEL`) — need `LoadBalancer` addresses kind can't provide. The
+  `InferenceGateway` installs MetalLB on the control plane itself
+  (`compose_metallb`, pool `.200-.250`); the serving stack does *not*, so `run.sh`
+  installs MetalLB on the workload cluster with a **disjoint** pool (`.100-.149`).
+  Both pools sit inside the detected kind Docker subnet (see caveat) so the
+  control plane can route across it to the workload gateway's IP.
+- **Fake DRA driver.** A `claim: DRA` engine emits a `ResourceClaim`; with no DRA
+  driver it stays Pending and the pod never schedules. `run.sh` applies the
+  vendored **dra-example-driver**, which publishes fake `gpu.example.com` devices
+  so the claim binds on a GPU-less node.
+- **Cross-cluster kubeconfig.** `source: Existing` needs a kubeconfig the
+  control-plane provider pods can use to reach the workload API server.
+  `kind get kubeconfig --internal` gives an address routable across the shared
+  kind network; a host kubeconfig (`127.0.0.1:<port>`) wouldn't be.
+- **Node label.** On a BYO cluster Modelplane doesn't provision/label pools, so
+  `run.sh` labels the workload node `modelplane.ai/pool=gpu-synthetic` (matching
+  `nodePools[].name`); without it worker pods stay Pending.
+
+## Caveats / open questions
+
+- **Cross-cluster networking uses the detected kind subnet.** `run.sh` reads the
+  `kind` Docker network's subnet (usually 172.18.0.0/16, but kind bumps to
+  172.19/... when earlier networks already hold 172.18) and derives both disjoint
+  MetalLB pools from it — the workload pool directly, the InferenceGateway's by
+  rewriting its manifest. A hardcoded 172.18 would leave the LB IP off-subnet and
+  the cross-cluster curl would time out.
+- **Reconcile runs after the command returns.** `crossplane project run` waits
+  for the config to install, then applies the resources and exits — it doesn't
+  block on XR readiness. The serving-stack install (the long pole) and the model
+  rollout happen after, so watch the `ModelService` address rather than the
+  command's exit. `--timeout` in `run.sh` bounds the build and config install.
+- **Two DRA drivers on a GPU-less node.** The serving stack's **NVIDIA** DRA
+  driver targets NFD-GPU-labelled nodes, so it sits at 0/0 (inert) yet its Helm
+  release still reports Ready. The **dra-example-driver** `run.sh` installs is the
+  active one — it publishes the fake `gpu.example.com` devices the engine binds.
+- **Serving-stack weight.** cert-manager, Envoy Gateway, Envoy AI Gateway, GAIE
+  CRDs, kube-prometheus-stack, LeaderWorkerSet, NFD, DRA driver — all on the
+  workload node. Give Docker headroom.
+- **Package version.** Installs the **current branch** build (via `crossplane
+  project run`), not the published `v0.1.0`, because the `source: Existing` /
+  DRA-on-BYO path may postdate that tag.
--- a/e2e/manifests/40-model-deployment.yaml
+++ b/e2e/manifests/40-model-deployment.yaml
@@ -1,0 +1,79 @@
+# A single Standalone engine (Unified serving, the default). Modelplane composes
+# the full serving path for it (an InferencePool, an endpoint-picker, and an
+# HTTPRoute), so this exercises real routing. The engine is a mock server: it
+# needs no GPU and no model, so the pod goes Ready and answers both
+# /v1/chat/completions (OpenAI) and /v1/messages (Anthropic), the way a vLLM
+# server exposes both, without real inference.
+apiVersion: modelplane.ai/v1alpha1
+kind: ModelDeployment
+metadata:
+  name: mock-demo
+  namespace: ml-team
+spec:
+  replicas: 1
+  template:
+    spec:
+      engines:
+      - name: mock
+        members:
+        - role: Standalone
+          nodeSelector:
+            devices:
+            - name: gpu
+              count: 1
+              selectors:
+              # Matches the fake GPU's declared capacity (80Gi >= 20Gi). Domain is
+              # gpu.example.com — the dra-example-driver's driver name.
+              - cel: |
+                  device.capacity["gpu.example.com"].memory.compareTo(quantity("20Gi")) >= 0
+          template:
+            spec:
+              containers:
+              - name: engine
+                # python:3.12-alpine, pinned by digest (multi-arch) so a moving tag can't flake CI.
+                image: python@sha256:6d43704baacd1bfbe7c295d7f13079d5d8104ed33568873133f8fc69980419df
+                command: ["python", "-u", "-c"]
+                args:
+                - |
+                  import json, http.server
+                  class H(http.server.BaseHTTPRequestHandler):
+                      def _s(self, o, c=200):
+                          b = json.dumps(o).encode()
+                          self.send_response(c)
+                          self.send_header("content-type", "application/json")
+                          self.send_header("content-length", str(len(b)))
+                          self.end_headers()
+                          self.wfile.write(b)
+                      def do_GET(self):
+                          if self.path == "/health":
+                              self._s({"status": "ok"})
+                          elif self.path.startswith("/v1/models"):
+                              self._s({"object": "list", "data": [{"id": "mock", "object": "model"}]})
+                          else:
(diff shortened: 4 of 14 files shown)


hack/local-e2e/manifests/10-inference-gateway.yaml
```diff
@@ -0,0 +1,20 @@
+# Control-plane front door. On kind there's no cloud LoadBalancer, so Traefik's
+# Service gets its external IP from MetalLB. Setting loadBalancer: MetalLB makes
+# the InferenceGateway composition install MetalLB and configure this pool
+# itself (see compose_metallb) — no separate MetalLB install needed, and it
+# serves the workload cluster's gateway too, since it's all one cluster.
```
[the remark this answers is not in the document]
ytsarev: good catch, amended

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,202 @@
...
+CP=modelplane-local
+WL=modelplane-workload
+# Pinned so the workload cluster has the DRA APIs the serving stack's NVIDIA DRA
+# driver needs (resource.k8s.io, GA in k8s 1.34). The control-plane cluster that
+# project run creates needs no DRA, so its image doesn't matter here.
+WL_NODE_IMAGE=kindest/node:v1.34.0@sha256:7416a61b42b1662ca6ca89f02028ac133a309a2a30ba309614e8ec94d976dc5a
+METALLB_URL=https://raw.githubusercontent.com/metallb/metallb/v0.14.8/config/manifests/metallb-native.yaml
+ROOT="$(git rev-parse --show-toplevel)"
+
+log() { printf '\n\033[1;34m==> %s\033[0m\n' "$*"; }
```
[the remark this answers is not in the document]
ytsarev: Added an `EXIT` trap that removes the temp init file, the isolated `DOCKER_CONFIG`, and the rendered manifests dir. Kept `ROOT` on `git rev-parse`: the flake app always runs `run.sh` from the repo tree (like the `run`/`build` apps, and `crossplane project run` itself needs the repo root), so there's no non-git invocation path here.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,202 @@
...
+# writes the status once, then exits) rather than `kubectl wait`, so a failed
+# attempt costs seconds instead of a full wait timeout.
+payload='{"model":"'"$svc"'","messages":[{"role":"user","content":"ping"}]}'
+code=""
+for attempt in $(seq 1 10); do
+	kubectl --context "$cpctx" -n "$ns" delete pod e2e-verify --ignore-not-found --now >/dev/null 2>&1 || true
+	kubectl --context "$cpctx" -n "$ns" run e2e-verify --restart=Never --image=curlimages/curl \
+		--command -- curl -sS --max-time 15 -o /dev/null -w '%{http_code}' \
+		"$addr/v1/chat/completions" \
+		-H 'content-type: application/json' -d "$payload" >/dev/null 2>&1 || true
```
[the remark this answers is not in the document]
ytsarev: Pinned the curl image to a multi-arch digest via a `CURL_IMAGE` constant in run.sh.

e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+So wait for the `ModelService` to publish an address before curling. That address
+is on the kind Docker subnet, which the host can't route to on macOS, so curl
+from a pod on the control plane:
+
+```bash
+kubectl -n ml-team get ms mock -w          # wait for ADDRESS to appear
+addr=$(kubectl -n ml-team get ms mock -o jsonpath='{.status.address}')
+kubectl run curl -n ml-team --rm -it --image=curlimages/curl -- \
+  curl -s "$addr/v1/chat/completions" -H 'content-type: application/json' \
+  -d '{"model":"mock","messages":[{"role":"user","content":"hi"}]}'
```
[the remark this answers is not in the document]
ytsarev: Pinned to the same digest, consistent with run.sh.

nix/apps.nix
```diff
@@ -236,6 +236,52 @@
...
+            pkgs.git
+            pkgs.bash
+          ];
+          inheritPath = false;
+          text = ''
+            mkdir -p _output
+            rm -f _output/functions
+            ln -s ${functionsPkg} _output/functions
+
+            exec bash hack/local-e2e/run.sh "$@"
```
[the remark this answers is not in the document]
ytsarev: Kept as-is for consistency: the `run`/`build`/`push` apps use the same relative `_output` handling and repo-root assumption, and `crossplane project run` requires the repo root regardless. Making the apps CWD-independent (and `rm -rf` the symlink target) is worth doing across all of them in a separate change rather than diverging one app here.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+# registry credentials.
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+  workflow_dispatch: {}
+
+# Cancel in-progress runs for the same PR. Runs on main are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
ytsarev: This is the same concurrency key as `ci.yml`. Kept them consistent; switching to the PR number would be a change to both workflows, so better done separately if we want it.

hack/local-e2e/manifests/40-model-deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+              selectors:
+              # Matches the fake GPU's declared capacity (80Gi >= 20Gi). Domain is
+              # gpu.example.com — the dra-example-driver's driver name.
+              - cel: |
+                  device.capacity["gpu.example.com"].memory.compareTo(quantity("20Gi")) >= 0
+          template:
+            spec:
+              containers:
+              - name: engine
+                image: python:3.12-alpine
```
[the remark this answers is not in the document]
ytsarev: Done — pinned `python:3.12-alpine` to its multi-arch digest, matching the curl pin.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,210 @@
...
+rendered="$(mktemp -d)"
+cp "$ROOT/hack/local-e2e/manifests/"*.yaml "$rendered/"
+sed -i.bak "s/172\.18\.255/${PREFIX}.255/g" "$rendered/"*.yaml && rm -f "$rendered/"*.bak
+extra=(--extra-resources "$rendered")
+verify=0
+case "${1:-}" in
+--no-apply)
+	extra=()
+	log "--no-apply: config only; apply manifests from $rendered"
+	;;
```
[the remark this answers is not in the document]
ytsarev: Good catch, that was a regression from the cleanup trap. Fixed: `cleanup` now keeps `$rendered` in `--no-apply` mode (a `keep_rendered` guard), so the manifests it points you at survive the exit. The init file and DOCKER_CONFIG are still removed since `project run` has already consumed them.

e2e/run.sh
```diff
@@ -0,0 +1,210 @@
...
+#     + the config + the InferenceGateway.
+set -euo pipefail
+
+CP=modelplane-local
+WL=modelplane-workload
+# Pinned so the workload cluster has the DRA APIs the serving stack's NVIDIA DRA
+# driver needs (resource.k8s.io, GA in k8s 1.34). The control-plane cluster that
+# project run creates needs no DRA, so its image doesn't matter here.
+WL_NODE_IMAGE=kindest/node:v1.34.0@sha256:7416a61b42b1662ca6ca89f02028ac133a309a2a30ba309614e8ec94d976dc5a
+METALLB_URL=https://raw.githubusercontent.com/metallb/metallb/v0.14.8/config/manifests/metallb-native.yaml
```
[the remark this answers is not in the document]
ytsarev: Kept as the versioned URL. dra-example-driver had to be vendored because it's an unpublished local chart (`0.0.0-dev`) with no released artifact; MetalLB ships a stable released manifest at this tagged URL, so fetching it is fine and standard for a test harness. I'd rather not vendor a ~500-line manifest for marginal benefit, but happy to pin the URL to the tag's commit SHA if we want the extra immutability.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,216 @@
...
+fi
+
+# Remove the temp files/dirs created below (init resources, the isolated Docker
+# config, the rendered manifests) on any exit, so a run doesn't leak them.
+cleanup() {
+	rm -rf "${init:-}" "${DOCKER_CONFIG:-}" >/dev/null 2>&1 || true
+	# --no-apply hands $rendered to the user to apply after the script exits, so
+	# keep it in that mode; otherwise project run has already consumed it.
+	[ "${keep_rendered:-0}" = 1 ] || rm -rf "${rendered:-}" >/dev/null 2>&1 || true
+}
```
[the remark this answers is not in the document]
ytsarev: Fixed, and a genuinely dangerous one, thanks. The temp Docker config now lives in a script-local `docker_config` var; cleanup removes that (an mktemp dir we created), never the exported `DOCKER_CONFIG` env var. If the script exits before creating it, `docker_config` is unset so nothing is removed, so a user's real `$DOCKER_CONFIG` is never touched.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,216 @@
...
+		"$ROOT/docs/manifests/getting-started/prerequisites.yaml"
+	printf '\n---\napiVersion: v1\nkind: Secret\nmetadata: {name: local-cluster-kubeconfig, namespace: modelplane-system}\nstringData:\n  kubeconfig: |\n'
+	kind get kubeconfig --internal --name "$WL" | sed 's/^/    /'
+} >"$init"
+
+# No system PATH in the nix app, so a Docker config using credsStore "desktop"
+# would break package resolution. The provider packages are public.
+DOCKER_CONFIG="$(mktemp -d)"
+printf '{}' >"$DOCKER_CONFIG/config.json"
+export DOCKER_CONFIG
```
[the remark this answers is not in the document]
ytsarev: Done as suggested: the temp path is a separate `docker_config` var, and we `export DOCKER_CONFIG="$docker_config"` from it, so cleanup only ever targets the path we created.

CONTRIBUTING.md
```diff
@@ -115,6 +115,16 @@ nix flake check            # or: ./nix.sh flake check
 `nix run .#fix` auto-fixes most lint and formatting issues. Run it before
 opening a PR.
 
+`nix flake check` is the unit layer — fast and sandboxed, proving each
+composition function renders the right resources. The integration layer is `nix
+run .#e2e-local`, which brings up two local `kind` clusters and runs the whole
+path — scheduling, the serving-stack install on a registered cluster, gateway
```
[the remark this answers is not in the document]
ytsarev: Reflowed so the `nix run .#e2e-local` span stays on one line.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,219 @@
...
+	for _ in $(seq 1 30); do
+		code="$(kubectl --context "$cpctx" -n "$ns" logs e2e-verify 2>/dev/null | tr -dc '0-9' || true)"
+		[ -n "$code" ] && break
+		sleep 2
+	done
+	log "verify attempt $attempt: HTTP ${code:-none}"
+	[ "$code" = "200" ] && break
+	sleep 10
+done
+kubectl --context "$cpctx" -n "$ns" delete pod e2e-verify --ignore-not-found --now >/dev/null 2>&1 || true
```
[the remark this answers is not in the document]
ytsarev: Fixed: each attempt now uses a unique pod (`e2e-verify-$attempt`) and reads that pod's logs, with a label-based cleanup at the end, so retries are fully isolated and can't read a previous pod's output.

e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+| Fleet scheduling / placement (CEL vs declared capacity) | Real token generation (mock engine) |
+| `ModelDeployment` → `ModelReplica` → `ModelEndpoint` → `ModelService` wiring | Real GPU drivers / CUDA (fake DRA devices only) |
+| DRA `ResourceClaim` → fake device binding (the real allocation path) | Multi-node / disaggregated (`PrefillDecode`) serving |
+| Serving-stack install on a real (BYO) workload cluster | Cloud provisioning (EKS/GKE/Nebius) |
+| Control-plane `InferenceGateway` + cross-cluster routing to the replica | |
+| Status propagation and foreground-deletion ordering | |
+
+## Prerequisites
+
+- **Docker** with real headroom — **≥ 16 GB memory** and **plenty of disk**
```
haarchri, line 56: can we start crossplane without the default activation policy and only create an MRAP for the provider-kubernetes, provider-helm resources - as the PR is using an "existing" cluster aka kind ?!, for the function an imageConfig can reference an deploymentRuntimeConfig with scale to 0 ?!

e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+`crossplane project run` installs the config and applies the resources, then
+returns; the serving-stack install and model rollout reconcile in the background.
+So wait for the `ModelService` to publish an address before curling. That address
+is on the kind Docker subnet, which the host can't route to on macOS, so curl
+from a pod on the control plane:
+
+```bash
+kubectl -n ml-team get ms mock -w          # wait for ADDRESS to appear
+addr=$(kubectl -n ml-team get ms mock -o jsonpath='{.status.address}')
+kubectl run curl -n ml-team --rm -it --image=curlimages/curl@sha256:7c12af72ceb38b7432ab85e1a265cff6ae58e06f95539d539b654f2cfa64bb13 -- \
```
haarchri, line 90: after we merged https://github.com/modelplaneai/modelplane/pull/360 would be good to run OpenAI and Antropic Message test here ?

something like this as addition:

```
curl -si $addr/v1/messages \
  -H "Content-Type: application/json" -H "anthropic-version: 2023-06-01" \
  -d '{"model":"qwen","max_tokens":1024,"messages":[{"role":"user","content":"Hi"}]}' | head -30

```

e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+  rejects an engine whose only device is `Synthetic`.)
+
+The engine is a **mock OpenAI server** (a few lines of Python), so the pod goes
+Ready and answers `/v1/chat/completions` without a real model or GPU.
+
+## What this does and does not test
+
+| Tested | Not tested |
+|---|---|
+| Fleet scheduling / placement (CEL vs declared capacity) | Real token generation (mock engine) |
```
haarchri, line 47: wonder if we can add this easy with some KWOK Nodes - as it will start specific Pod on it and it reached Ready - so we can test placement with the scheduler ...

hack/local-e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+  cluster (the workload kind cluster) via a kubeconfig Secret instead of
+  provisioning EKS/GKE/Nebius.
+- **`claim: DRA` + a fake GPU driver** — the `InferenceClass` advertises a
+  `gpu.example.com` device backed by the **dra-example-driver**
+  (kubernetes-sigs), which publishes fake GPUs with **no real hardware**. The
+  engine's `ResourceClaim` binds a fake device on a GPU-less node, so the *real*
+  DRA allocation path runs. (A claimable device is required: the fleet scheduler
+  rejects an engine whose only device is `Synthetic`.)
+
+The engine is a **mock OpenAI server** (a few lines of Python), so the pod goes
```
haarchri: wonder if we can use here an CPU based model instead ?!

hack/local-e2e/dra-example-driver.yaml
```diff
@@ -0,0 +1,271 @@
...
+# run.sh applies this to the workload cluster. The kubeletplugin publishes 8 fake
+# gpu.example.com devices via a ResourceSlice, so a `claim: DRA` engine's
+# ResourceClaim binds on a GPU-less node — exercising the real DRA allocation
+# path the fleet scheduler and composition emit, without a GPU.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: dra-example-driver
+---
+---
```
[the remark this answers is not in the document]
ytsarev: Fixed — dropped the redundant separator so there's no empty document, and updated the regen note so re-vendoring won't reintroduce it (the `helm template` output already starts with `---`).

hack/local-e2e/README.md
```diff
@@ -0,0 +1,176 @@
...
+  driver targets NFD-GPU-labelled nodes, so it sits at 0/0 (inert) yet its Helm
+  release still reports Ready. The **dra-example-driver** `run.sh` installs is the
+  active one — it publishes the fake `gpu.example.com` devices the engine binds.
+- **Serving-stack weight.** cert-manager, Envoy Gateway, Envoy AI Gateway, GAIE
+  CRDs, kube-prometheus-stack, LeaderWorkerSet, NFD, DRA driver — all on the
+  workload node. Give Docker headroom.
+- **Package version.** Installs the **current branch** build (via `crossplane
+  project run`), not the published `v0.1.0`, because the `source: Existing` /
+  DRA-on-BYO path may postdate that tag.
+```
```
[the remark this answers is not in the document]
ytsarev: Fixed — removed the stray closing fence; the code fences balance now.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,221 @@
...
+	# keep it then; otherwise project run has already consumed it.
+	if [ -n "${rendered:-}" ] && [ "${keep_rendered:-0}" != 1 ]; then rm -rf "$rendered" || true; fi
+}
+trap cleanup EXIT
+
+WLCTX="kind-$WL"
+if ! kind get clusters 2>/dev/null | grep -qx "$WL"; then
+	log "Creating workload cluster $WL (k8s v1.34, for DRA)"
+	kind create cluster --name "$WL" --image "$WL_NODE_IMAGE"
+fi
```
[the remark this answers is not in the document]
ytsarev: Good call. Added a guard: if a `modelplane-workload` cluster already exists, the run now checks its kubelet version and fails fast with a `--clean` remediation unless it's v1.34, instead of silently reusing a stale cluster and hitting missing DRA APIs later.

hack/local-e2e/README.md
ytsarev: I thought about this, but from the modelplane e2e testing perspective, will it provide more coverage than a lightweight mock?

e2e/README.md
ytsarev: Thanks a lot, I think it is a great idea for this e2e test suite evolution. Will probably address in a separate PR

e2e/README.md
ytsarev: Thanks, great suggesttion -  implemented in `hack/local-e2e/lean-control-plane.yaml` - currently scaled down providers -only 
https://github.com/ytsarev/modelplane-public/actions/runs/29962201358 - reduced CI run by 10 minutes 👍

e2e/README.md
ytsarev: image

Implemented, see test CI run https://github.com/ytsarev/modelplane-public/actions/runs/29970107639/job/89090097159

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    if: >-
+      github.event_name == 'workflow_dispatch'
+      || contains(github.event.pull_request.labels.*.name, 'test-e2e')
+    runs-on: ubuntu-24.04
```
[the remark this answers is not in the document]
ytsarev: Good catch. Fixed in `15958e3`: the `labeled` event is now scoped to the `test-e2e` label, so adding any *other* label to a PR that already carries it no longer fires a fresh ~20 min run.

```yaml
|| (github.event.action == 'labeled' && github.event.label.name == 'test-e2e')
|| (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'test-e2e'))
```

Pushes (`synchronize`) still re-run while the label is set, and manual `workflow_dispatch` is unaffected.

hack/local-e2e/dra-example-driver.yaml
```diff
@@ -0,0 +1,271 @@
...
+      - name: plugin
+        securityContext:
+          privileged: true
+        image: registry.k8s.io/dra-example-driver/dra-example-driver:v0.4.0
+        imagePullPolicy: IfNotPresent
+        command: ["dra-example-kubeletplugin"]
+        resources:
+          {}
+        
+        livenessProbe:
```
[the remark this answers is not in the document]
ytsarev: Fixed in `aa268c1` — stripped the trailing whitespace on that line.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,232 @@
...
+# node (vendored dra-example-driver — see dra-example-driver.yaml). Without a DRA
+# driver the ResourceClaim stays Pending and the engine pod never schedules; the
+# fleet scheduler also rejects an engine whose only device is Synthetic.
+log "Installing dra-example-driver (fake GPUs) on the workload cluster"
+kubectl --context "$WLCTX" apply -f "$ROOT/hack/local-e2e/dra-example-driver.yaml"
+kubectl --context "$WLCTX" -n dra-example-driver rollout status ds/dra-example-driver-kubeletplugin --timeout=120s
+
+# BYO clusters aren't labelled by Modelplane; the gpu-synthetic pool selects on this.
+log "Labelling the workload node for pool gpu-synthetic"
+kubectl --context "$WLCTX" label node "${WL}-control-plane" modelplane.ai/pool=gpu-synthetic --overwrite
```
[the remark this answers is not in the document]
ytsarev: Leaving this one as-is. The script creates the workload cluster itself as a single-node `kind` cluster named `modelplane-workload`, so the node name is deterministically `modelplane-workload-control-plane`. It only reuses a pre-existing cluster when a version guard confirms it's the expected v1.34 single-node layout, and fails fast (pointing at `--clean`) otherwise. A hand-built multi-node cluster reusing that name is out of scope for the harness, so targeting the known node name is intentional here.

e2e/README.md
ytsarev: Opened #367 to track this: https://github.com/modelplaneai/modelplane/issues/367 — it captures the placement-testing gap and the KWOK approach, plus the DRA-on-fake-nodes question to work through. Thanks again for the idea 🙏

hack/local-e2e/dra-example-driver.yaml
```diff
@@ -0,0 +1,271 @@
...
+    app.kubernetes.io/instance: dra-example-driver
+    app.kubernetes.io/version: "v0.4.0"
+    app.kubernetes.io/managed-by: Helm
+---
+# Source: dra-example-driver/templates/clusterrole.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: dra-example-driver-role
+  namespace: dra-example-driver
```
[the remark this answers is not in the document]
ytsarev: This file is vendored `helm template` output (see the header) and applied as-is. It's generated with `--namespace dra-example-driver`, and Helm stamps that namespace onto the cluster-scoped resources too. Kubernetes ignores `namespace` on cluster-scoped kinds, and both `kubectl` and `crossplane project run` apply the file cleanly (the e2e passes), so I'd rather keep it as faithful generator output than hand-edit it — a manual removal would just reappear on the next regen.

hack/local-e2e/run.sh
```diff
@@ -0,0 +1,269 @@
...
+fi
+
+# Both kind clusters share one Docker network; MetalLB hands out LoadBalancer IPs
+# from it, and the control plane must ROUTE to the workload gateway's IP across
+# it. So the pools must sit inside the *actual* kind subnet — normally
+# 172.18.0.0/16, but kind bumps to 172.19/172.20/... when earlier Docker networks
+# already hold 172.18. Detect it and derive both pools (this one and the
+# InferenceGateway's) from the same prefix; a hardcoded 172.18 leaves the LB IP
+# off-subnet and silently breaks cross-cluster routing (curl times out).
+SUBNET="$(docker network inspect kind -f '{{range .IPAM.Config}}{{println .Subnet}}{{end}}' | grep -E '^[0-9]+\.' | head -1)"
```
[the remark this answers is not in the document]
ytsarev: Fixed in e804450 — added `|| true` to the subnet-detection command substitution so a detection miss no longer trips `set -e` before the `[ -n "$PREFIX" ]` guard. It now reports "could not detect the kind Docker subnet" as intended instead of aborting opaquely. Confirmed the behavior with a standalone `set -euo pipefail` repro. Good catch.

hack/local-e2e/lean-control-plane.yaml
```diff
@@ -0,0 +1,50 @@
...
+#    compositions use, so the cloud providers' MRs stay dormant (fewer CRDs).
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: ManagedResourceActivationPolicy
+metadata:
+  name: default
+spec:
+  activate:
+  - "*.kubernetes.crossplane.io"
+  - "*.kubernetes.m.crossplane.io"
+  - "*.helm.crossplane.io"
```
haarchri: we we need the cluster scoped ? - in the `apis/mrap.yaml` we have 


```
  # provider-helm and provider-kubernetes (serving stack, gateway, caches).
  - releases.helm.m.crossplane.io
  - objects.kubernetes.m.crossplane.io
```

e2e/run.sh
```diff
@@ -0,0 +1,271 @@
...
+	--control-plane-name "$CP" --cluster-admin --timeout 25m \
+	--init-resources "$init" \
+	"${extra[@]}"
+
+if [ "$verify" = 0 ]; then
+	log "Done. Curl the ModelService per the README; clean up with: nix run .#e2e-local -- --clean"
+	exit 0
+fi
+
+# --verify: project run returns once the config is healthy and the resources are
```
haarchri, line 180: i wonder if we can run some python test-suites after the project is running in kind - perhaps good for an follow up ?!

e2e/manifests/40-model-deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+              count: 1
+              selectors:
+              # Matches the fake GPU's declared capacity (80Gi >= 20Gi). Domain is
+              # gpu.example.com — the dra-example-driver's driver name.
+              - cel: |
+                  device.capacity["gpu.example.com"].memory.compareTo(quantity("20Gi")) >= 0
+          template:
+            spec:
+              containers:
+              - name: engine
```
haarchri, line 32: are we sure that an override of this will also test something we bumped recently ? or is this followup then ?! disaggregation with EPP and then we have multiple container involved eg. https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-replica/function/routing.py#L60-L61 https://github.com/modelplaneai/modelplane/blob/main/functions/compose-model-replica/function/routing.py#L523 recently bumped because it was not working with Anthropic message style .. eg. https://github.com/modelplaneai/modelplane/pull/360/

e2e/run.sh
ytsarev: Yes, I would love to use the consolidated test framework, but upbound-style E2E API is not yet available upstream.

hack/local-e2e/lean-control-plane.yaml
ytsarev: Good catch. Dropped the cluster-scoped forms — the BYO compositions run provider-helm/kubernetes v2, whose MRs are namespaced (`.m.crossplane.io`), exactly what `apis/mrap.yaml` activates. The lean MRAP now keeps only `*.kubernetes.m.crossplane.io` and `*.helm.m.crossplane.io`.

Verified with a fresh run, locally and in CI — both `/v1/chat/completions` and `/v1/messages` still return 200. (2063156)

e2e/run.sh
ytsarev: Great idea. Opened #368 to track it as a follow-up. I've kept this PR's `--verify` at a smoke-level 200; behavioral suites against the running control plane are the natural next step (pairs with the CPU-model idea, for something meaningful to assert against). Thanks!

e2e/manifests/40-model-deployment.yaml
ytsarev: Good question — split answer:

- The recent Unified-through-EPP routing and the Anthropic `/v1/messages` surface (#360) **are** exercised: the mock is a Unified engine, so `a7d28c3` routes it through the composed InferencePool + endpoint-picker + HTTPRoute, and `--verify` asserts 200 on both `/v1/chat/completions` and `/v1/messages` through that path.
- Disaggregation (PrefillDecode — the pd-sidecar, two-phase engines, disagg EPP config, multiple containers) is **not** covered here — a single Unified engine doesn't touch it. That's a genuine follow-up, already listed under "Not tested" in the harness README.

hack/local-e2e/lean-control-plane.yaml
ytsarev: https://github.com/ytsarev/modelplane-public/actions/runs/30088208329/job/89465250927 related CI run

flake.nix
```diff
@@ -174,6 +174,7 @@
             dockerCredentialUp = pkgs.upbound;
           };
           stop = apps.stop { inherit crossplane; };
+          e2e-local = apps.e2eLocal { inherit crossplane functionsPkg; };
```
negz: Why `e2e-local`, vs just `e2e`?

nix/apps.nix
```diff
@@ -236,6 +236,52 @@
       );
     };
 
+  # Run the two-cluster local end-to-end test (no cloud, no GPU): a workload
```
negz: Nit: "no cloud, no GPU" reads like LLM selling/bragging here. I'd drop it.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,97 @@
+name: E2E
+
+# The local two-cluster e2e (hack/local-e2e) is heavy — two kind clusters plus
+# the full serving stack, ~15-20 min — so it doesn't run on every push. Add the
+# `test-e2e` label to a PR to run it (it re-runs on each push while the label is
+# set), or trigger it manually. Unlike the package push, it needs no cloud or
+# registry credentials.
```
negz: What makes it take so long? 15-20 mins is surprising given it's not pulling any weights or large engine images.

e2e/run.sh
```diff
@@ -0,0 +1,271 @@
+#!/usr/bin/env bash
```
negz, line 1: WDYT about homing all of this under `./e2e`, not `./hack/local-e2e`? Or is the expectation there'd be other, meaningfully different E2E tests?

e2e/run.sh
```diff
@@ -0,0 +1,271 @@
...
+# workload ServingStack (Envoy) both install the Gateway API CRDs — on one
+# cluster they race for the same cluster-scoped CRDs and the gateway wedges. So:
+#   - a workload kind cluster (this script creates it), registered via
+#     source: Existing, where the serving stack + model run;
+#   - a control-plane cluster (crossplane project run manages it) with crossplane
+#     + the config + the InferenceGateway.
+set -euo pipefail
+
+CP=modelplane-local
+WL=modelplane-workload
```
negz: Maybe `modelplane-e2e-local` and `modelplane-e2e-workload` to reduce chance of collision with other kind clusters?

e2e/run.sh
```diff
@@ -0,0 +1,271 @@
...
+
+# rm_tmp deletes an mktemp path this script created, but only if it genuinely
+# lives under a temp root. The one recursive removal we can't avoid — the Docker
+# config dir, which Docker populates and may nest subdirs in — stays confined
+# behind this check, so no bug here can rm -rf something real.
+rm_tmp() {
+	case "${1:-}" in
+	"$tmproot"/* | /tmp/* | /var/folders/* | /private/var/folders/*) rm -rf "$1" || true ;;
+	esac
+}
```
negz: Do I understand correctly that this is maintaining a list of dirs it's acceptable to delete the temp subdirectories from? Feels a little over-engineered / over-defensive.

e2e/run.sh
```diff
@@ -0,0 +1,271 @@
...
+	# resources (Gateway API, MetalLB, Service/EndpointSlice) on the control plane;
+	# without them the InferenceGateway wedges on "failed waiting for informer to
+	# sync" — crossplane can't watch the Namespace the IG composes.
+	#
+	# Strip prerequisites.yaml's leading comment-only document: `crossplane project
+	# run --init-resources` rejects it ("invalid or empty raw resource") where
+	# kubectl would skip it. awk drops leading comment/blank/`---` lines and prints
+	# from the first real resource line on (inline comments in resources are fine).
+	# Workaround for crossplane/cli#207 (fix: crossplane/cli#208); once that ships,
+	# drop this awk and pass the file through unmodified.
```
negz: Does https://github.com/modelplaneai/modelplane/pull/375 help with this? Might be cleaner than awking the file.

negz (commented):
Thanks Yury! A few small suggestions but I think this is a good start for an E2E framework.

flake.nix
ytsarev: my original thinking was - local e2e only, then extend with 'real' e2e. But i think they will build up on each other, so it make sesnse to simplify naming, thanks for the catch!

e2e/run.sh
ytsarev: Yes, I tried to guard against `rm -rf` (unintentional) hijacking.

.github/workflows/e2e.yml
ytsarev: Profiled a CI run ([run 30901652231](https://github.com/ytsarev/modelplane-public/actions/runs/30901652231)), ~17.5 min total, timed from the Actions log. It is install and reconcile, not model weights:

- **Install config + ~13 providers: 6:13** (crossplane installs the Configuration and its provider deps, waits for every revision healthy)
- **Serving-stack reconcile: 4:18** (cert-manager, Envoy Gateway + Envoy AI Gateway, GAIE, kube-prometheus-stack, LeaderWorkerSet, NFD, NVIDIA DRA driver)
- **nix build of function images: 3:51** (cache-dependent)
- crossplane setup + package load: 2:11
- workload cluster + MetalLB + dra-example-driver: 0:59
- verify curls (both surfaces): 0:04

The engine is a mock, so no model weights or large engine images are pulled.

e2e/run.sh
ytsarev: Good call, done. `--init-resources` now carries only `lean-control-plane.yaml`; `prerequisites.yaml` is applied with `kubectl` after the config is healthy, with provider-helm repointed at its DeploymentRuntimeConfig (the finishing steps #375 added). The awk is gone, and it is validated end to end: both `/v1/chat/completions` and `/v1/messages` return 200, locally and in CI.

e2e/run.sh
ytsarev: Renamed to `modelplane-e2e-local` and `modelplane-e2e-workload`.

nix/apps.nix
ytsarev: Dropped "no cloud, no GPU" from the app description and its comment.

e2e/run.sh
ytsarev: Done, moved everything to `./e2e/` and renamed the app to `.#e2e`.

e2e/run.sh
ytsarev: Good call, dropped the allowlist for the idiomatic `work="$(mktemp -d)"` + `trap 'rm -rf "$work"' EXIT` (Docker config and rendered manifests are subpaths now); validated locally and in CI.

ytsarev:
@negz, thanks a lot for the review! I addressed all comments. Please check them inline.

The latest E2E CI run reflecting the current state https://github.com/ytsarev/modelplane-public/actions/runs/30912005510 

UPD: very latest run after minor fixes https://github.com/ytsarev/modelplane-public/actions/runs/30915499396

e2e/run.sh
```diff
@@ -0,0 +1,250 @@
...
+log "Verifying the model serves end to end"
+ns=ml-team
+svc=mock
+
+addr=""
+for _ in $(seq 1 80); do
+	addr="$(kubectl --context "$cpctx" -n "$ns" get modelservice "$svc" -o jsonpath='{.status.address}' 2>/dev/null || true)"
+	[ -n "$addr" ] && break
+	sleep 15
+done
```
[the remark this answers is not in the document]
ytsarev: ShellCheck does not flag this. `_` is its designated name for an intentionally-unused variable, so it does not raise SC2034 for it (that is why the loop uses it). `shellcheck e2e/run.sh` is clean, and the CI `check` job that runs shellcheck passes on these exact loops (for example run `affd2fb`). No change needed.

negz (approved):
Thanks @ytsarev!

PR #366, opened 2026-07-22
feat: configurable ProviderConfig per cluster type; derive GKE project from ProviderConfig

### LINKED ISSUE #177: compose-gke-cluster MRs hard-pinned to default ClusterProviderConfig, no way to target multiple projects
### What happened?

None of the cloud managed resources composed by `compose-gke-cluster` (Network, Subnetwork, Cluster, NodePools, ServiceAccount, IAM, etc.) set a `spec.providerConfigRef`. With the namespaced provider model (`gcp.m.upbound.io`), an omitted ref resolves to the `ClusterProviderConfig` named `default`. There is also no field on the InferenceCluster / GKECluster API to carry a provider-config name, so it cannot be overridden.

Net effect: we can only ever target the single cloud project/account wired into the `default` ClusterProviderConfig. No per-tenant, multi-project, or multi-credential setups are possible.

Pretty sure this applies to the AWS composition as well, not just GKE.

Expected v1 scoping decision, or a gap to close (thread a `providerConfigRef` through the API down to the MRs)?

### How can we reproduce it?

1. Apply an InferenceCluster pointing at a GKE backend.
2. Inspect any composed MR, e.g.: kubectl get cluster.container.gcp.m.upbound.io <name> -o yaml
3. Observe `spec.providerConfigRef.name: default` (the implicit fallback) — there is no way to set it to a different ClusterProviderConfig via the InferenceCluster API.

Workaround: none today. The only project/account reachable is the one bound to the `default` ClusterProviderConfig.

Modelplane version: v0.1.0-rc.0

### LINKED ISSUE #178: InferenceCluster requires spec.cluster.gke.project despite ClusterProviderConfig already setting projectID
### What happened?

ProjectID is specified in two places. The `ClusterProviderConfig.gcp.m.upbound.io` already carries `spec.projectID`, but the InferenceCluster also requires `spec.cluster.gke.project`. Since the provider config already establishes the project, requiring it again on the InferenceCluster is redundant.

Could we derive it from the resolved provider config and make `spec.cluster.gke.project` optional?

Minor today, but combined with the single-`default`-provider-config limitation https://github.com/modelplaneai/modelplane/issues/177 it will bite anyone trying to run more than one project.

### How can we reproduce it?

1. Author an InferenceCluster manifest.
2. Note that `spec.cluster.gke.project` is required even though the bound `ClusterProviderConfig` already sets `spec.projectID`.
3. Set the two to mismatched values to see there is no reconciliation/validation between them.

Workaround: keep both fields manually in sync.

Modelplane version: v0.1.0-rc.0

haarchri:

Add an optional credentials field (`type`, `name`) to all five cluster types (`GKECluster`, `EKSCluster`, `AKSCluster`, `NebiusCluster`, InferenceCluster sub-specs). Defaults to {type: ClusterProviderConfig, name: default} for full backward compatibility. Allows deploying multiple clusters against different cloud accounts, subscriptions, or projects...

Remove `spec.project `from `GKECluster` and the InferenceCluster GKE sub-spec. The GCP project is now derived from `spec.projectID` on the referenced ProviderConfig or ClusterProviderConfig, the same pattern Nebius uses. All GCP managed resource `forProvider.project` fields are optional (the provider uses the config's projectID when unset) the only place the value is needed is the Workload Identity pool string (`{project}.svc.id.goog`). An `_observed_project()` fallback reads the project from the observed cluster's workloadPool if the ProviderConfig is temporarily unavailable, preventing desired-state gaps from tearing down a running cluster.

Fixes #178
Fixes #177

I have: 

### Tested

```
kubectl get cluster.container.gcp.m.upbound.io/starter-795585bfb267  -o yaml -n modelplane-system
apiVersion: container.gcp.m.upbound.io/v1beta1
kind: Cluster
metadata:
  annotations:
    crossplane.io/composition-resource-name: cluster
    crossplane.io/external-name: starter-795585bfb267
  creationTimestamp: "2026-07-22T13:47:39Z"
  generateName: starter-
  generation: 2
  labels:
    crossplane.io/composite: starter
  name: starter-795585bfb267
  namespace: modelplane-system
  ownerReferences:
  - apiVersion: infrastructure.modelplane.ai/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: GKECluster
    name: starter
    uid: 9d617041-d4e1-4a26-8ea0-8af3ce69dd88
  resourceVersion: "5846"
  uid: ec198538-b8ee-49b8-9b80-cae94024b614
spec:
  forProvider:
    addonsConfig:
      gcpFilestoreCsiDriverConfig:
        enabled: true
    deletionProtection: false
    initialNodeCount: 1
    ipAllocationPolicy:
      clusterSecondaryRangeName: pods
      servicesSecondaryRangeName: services
    location: us-central1
    minMasterVersion: "1.35"
    networkSelector:
      matchControllerRef: true
    releaseChannel:
      channel: REGULAR
    removeDefaultNodePool: true
    subnetworkSelector:
      matchControllerRef: true
    workloadIdentityConfig:
      workloadPool: chris-gcp-project.svc.id.goog
  initProvider: {}
  managementPolicies:
  - '*'
  providerConfigRef:
    kind: ClusterProviderConfig
    name: chris-test
...
```


apis/gkeclusters/definition.yaml
```diff
@@ -17,9 +17,6 @@ spec:
   - name: v1alpha1
     referenceable: true
     additionalPrinterColumns:
```
haarchri, line 19: wonder if we want to add the new Credentails Field informations as printerColumns ?!

docs/manifests/examples/collecting-engine-metrics/inference-cluster.yaml
```diff
@@ -16,7 +16,6 @@ spec:
   cluster:
     source: GKE
     gke:
-      project: my-gcp-project  # Replace with your GCP project ID.
```
haarchri, line 19: its safe to remove as we have the following on the field docs:

`The ID of the project in which the resource belongs. If it is not provided, the provider project is used.
`

functions/compose-gke-cluster/function/fn.py
```diff
@@ -160,8 +162,74 @@ def __init__(self, req: fnv1.RunFunctionRequest, rsp: fnv1.RunFunctionResponse)
         self.req = req
         self.rsp = rsp
         self.xr = v1alpha1.GKECluster(**resource.struct_to_dict(req.observed.composite.resource))
+        self.project: str | None = None
+
+    def _cred_kind(self) -> str:
+        creds = self.xr.spec.credentials
+        return creds.type if creds and creds.type else "ClusterProviderConfig"
```
haarchri, line 169: tbd. if we need the else here, for backward yes otherwise we getting it via CRD default always set ...

functions/compose-gke-cluster/function/fn.py
```diff
@@ -160,8 +162,74 @@ def __init__(self, req: fnv1.RunFunctionRequest, rsp: fnv1.RunFunctionResponse)
...
+
+    def _cred_name(self) -> str:
+        creds = self.xr.spec.credentials
+        return creds.name if creds and creds.name else "default"
```
haarchri, line 173: same here: tbd. if we need the else here, for backward yes otherwise we getting it via CRD default always set ...

functions/compose-gke-cluster/function/fn.py
```diff
@@ -160,8 +162,74 @@ def __init__(self, req: fnv1.RunFunctionRequest, rsp: fnv1.RunFunctionResponse)
...
+        as the default for all managed resources, so we only need the value
+        explicitly for the workloadPool string in the GKE cluster spec.
+
+        When the ProviderConfig is transiently gone but the cluster already
+        exists, falls back to the project embedded in the observed cluster's
+        workloadPool so a mistakenly deleted config doesn't tear down the
+        cluster. A recreated config takes over on the next reconcile."""
+        cred_kind = self._cred_kind()
+        cred_name = self._cred_name()
+        response.require_resources(
```
dennis-upbound, line 188: should this be namespace scoped?

functions/compose-nebius-cluster/function/fn.py
```diff
@@ -255,44 +259,60 @@ def resolve_credentials(self) -> Credentials | None:
...
+        cred_kind = self._cred_kind()
+        cred_name = self._cred_name()
         response.require_resources(
             self.rsp,
             name="nebius-provider-config",
             api_version="nebius.m.upbound.io/v1beta1",
-            kind="ClusterProviderConfig",
-            match_name=_NEBIUS_PROVIDER_CONFIG_NAME,
+            kind=cred_kind,
+            match_name=cred_name,
```
dennis-upbound, line 278: should this be namespace scoped?

dennis-upbound (approved):
lgtm - 2 small nits/questions

functions/compose-gke-cluster/function/fn.py
haarchri: thanks good catch

functions/compose-nebius-cluster/function/fn.py
haarchri: thanks good catch

Issue #331, opened 2026-07-08
Keep cluster provisioning implementations consistent as we add providers


negz:
### What problem are you facing?

`InferenceCluster` provisions clusters on two clouds today: GKE via [`compose-gke-cluster`](https://github.com/modelplaneai/modelplane/blob/45fc1998b5f4c9231c119255e209a2cc88e3f17e/functions/compose-gke-cluster/function/fn.py) and EKS via [`compose-eks-cluster`](https://github.com/modelplaneai/modelplane/blob/45fc1998b5f4c9231c119255e209a2cc88e3f17e/functions/compose-eks-cluster/function/fn.py), each a Python composition function behind an internal `GKECluster` / `EKSCluster` XR. The [Ecosystem backlog](https://github.com/modelplaneai/modelplane/issues?q=is%3Aissue+is%3Aopen+label%3AEcosystem) has around twenty more clouds queued (#290–#305, #93, #94), with Nebius (#299) next.

Each new cloud is another cluster composition function. I want to make sure our implementations don't drift as we add more. I imagine we'll lean heavily on agents to do this work.

I'm worried about two kinds of drift:

* Behavioral: an implementation tracks readiness, gates a dependent resource, or hands off its kubeconfig a little differently, passes review because it looks plausible, and now one cloud carries a bug the others don't.
* Stylistic: I don't want to implement the same thing N different ways. Our patterns, naming, etc should be clear and consistent so that if you've worked on one implementation you don't need to learn new patterns to work on another.

### How could Modelplane help solve your problem?

I'm not sure. Some options:

- A written contract and reference: what every implementation must expose (the API shape) and deliver (kubeconfig Secret, cache StorageClass), plus the conventions, with GKE and EKS as the worked examples.
- A skill carrying that same content as an agent workflow, given how agentic this work will be.
- Conformance tests each implementation must pass: every composed resource appears in the readiness list, status publishes a Kubeconfig secret and a cache StorageClass, and so on. These catch the behavioral drift a doc can't.
- A single canonical implementation to copy from, rather than two to choose between.

The trade-offs cut across these. Docs and skills inform but don't enforce; tests enforce behavior but not style; a canonical reference anchors style but its copies drift once edited. With only two implementations we also don't yet know what's genuinely invariant, so proving the pattern on the next provider or two (#299) before locking in enforcement likely beats designing it up front.

Issue #204, opened 2026-06-18
Model load from the EFS-backed ModelCache is slow


negz:
### What problem are you facing?

Serving Qwen3-Coder-480B (894 GiB on disk, BF16) as a two-node Leader+Worker gang on EKS, vLLM spent **~43 minutes** just reading the safetensors weights from the `ModelCache` RWX volume at startup — before CUDA graph capture and KV cache init even began. The cache itself *hydrated* fine (HuggingFace → EFS, ~960 GB in ~31 min, ~31 GiB/min); the bottleneck is the per-pod **read** of weights when the engine loads.

The per-shard load rate sawtooths: bursts of 1–2.5 it/s when a shard is already in the OS page cache (vLLM prefetches ahead with 8 threads), collapsing to 15–30 s/it when the loader catches the prefetch frontier and reads cold from EFS. Net sustained ~2–4 %/min across 241 shards. This read cost is paid on *every* pod start, so it dominates startup and can make the cache a net latency regression versus a direct pull for a single deployment.

Infra (us-east-2):
- 2× `p5en.48xlarge` (16× H200), AZ us-east-2b, EKS 1.36, AL2023 NVIDIA AMI.
- ModelCache RWX backing: Amazon EFS, **throughputMode `elastic`**, performanceMode `generalPurpose`, encrypted, ~960 GB. EFS CSI StorageClass `modelplane-rwx-efs` (`provisioningMode=efs-ap`). Same-AZ mount target (no cross-AZ).
- 16 vLLM workers (8 TP ranks/node × 2 PP nodes) read shards concurrently from the one filesystem.

EFS elastic throughput doesn't sustain the multi-GB/s a 16-GPU cold load wants; the page cache (8 prefetch threads into ~1.9 TiB RAM) hides latency but not the cold reads. EFS *write* during hydration was fine — it's the concurrent read fan-out that bogs down.

### How could Modelplane help solve your problem?

The `ModelCache`'s EFS-backed storage tier looks like the practical floor for cold-loading a ~1 TB model. Worth exploring faster options, but no specific proposal yet.

Related to #186 / #115 (cache footprint / hydration image).

negz:
I looked into this and I think we're blaming the wrong layer. The EFS-backed cache wasn't the bottleneck on the #193 run — the vLLM weight loader was. Before we go build a faster storage tier, we should try a faster loader, because it's free and the dependency is already in the image.

[...]

PR #370, opened 2026-07-28
Add VULTR vke  as an inference cluster provider

### LINKED ISSUE #305: Support Vultr (VKE) for cluster provisioning

`InferenceCluster` can provision clusters on [GKE and EKS](https://github.com/modelplaneai/modelplane/blob/main/apis/inferenceclusters/definition.yaml) today. Teams running on Vultr can't have Modelplane provision a cluster for them, so they fall back to a bring-your-own cluster via `source: Existing` (#3).

Add `VKE` as a value for `spec.cluster.source` on `InferenceCluster`, so Modelplane provisions the cluster and its GPU node pools the same way it does for GKE and EKS. Provisioning runs through a Crossplane provider for Vultr; where one doesn't exist yet, building that provider is part of this work. The rest of the stack is provider-agnostic and only needs a working kubeconfig.

haarchri:

Modelplane can provision inference clusters on GKE, EKS, AKS, and Nebius; Vultr was listed as Planned on the providers page. This adds Vultr VKE as a fifth provisioned cluster source, built on the `xpkg.upbound.io/upbound/provider-vultr` provider.

An `InferenceCluster` with `source: Vultr` composes a new `VultrCluster` XR, served by the new `compose-vultr-cluster` function. The VKE cluster carries a fixed inline system pool for control-plane components; each user GPU pool becomes a separate `VkeNodePool` managed resource gated on cluster readiness, so pools can be added, resized, or removed without touching the cluster. `InferenceClass` gains a `vultr` provisioning block. Vultr sizes nodes by plan (e.g. `vcg-l40s-16c-180g-48vram` for 1x L40S), so the block is just plan plus accelerator.

Two things VKE ships built in shape the composition. Vultr pre-installs the NVIDIA GPU Operator on GPU clusters, so rather than composing a GPU stack, `VultrCluster` gates its readiness on an Observe-only `Object` with a CEL query over the `nvidia-operator-validator` DaemonSet, the cluster only reports Ready once the GPU stack is validated and `nvidia.com/gpu` is advertised, and the serving stack never starts before that. The Vultr CSI driver likewise installs the `vultr-vfs-storage` RWX StorageClass (Vultr File System) on every cluster, and node autoscaling is served by VKE itself, so no CSI driver, StorageClass, or in-cluster autoscaler is composed.

Known limitation: 

- Vultr uses `vultr-vfs-storage` with ReadWriteMany (RWX), backed by Vultr File System. However, this storage is currently not usable on GPU nodes. Because of that, we decided not to ship ModelCache with Longhorn for Vultr for now. The main concern is performance: Longhorn may become a bottleneck when multiple engines load a model at the same time. Instead, Vultr will ship without ModelCache support for now. This is better than adding a cache that could actually make model startup slower.
- fabric networking is not available for GPU nodes either, so there is no high-bandwidth interconnect for multi-node serving. Serving is therefore limited to single-node GPU instances#

Note: 
- VKE does not support VPC-only clusters (the `vpc_only` flag fails cluster creation with Error 500), so nodes keep public network interfaces; a VPC can be attached but not made the only network; thats how Vultr API works today

Fixes #305

I have: 

### Tested
Validated end to end: a single-node L40S GPU pool provisions, passes the GPU readiness gate, and serves a model. 

- Create ProviderConfig + Secret

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: vultr-credentials
  namespace: crossplane-system
type: Opaque
stringData:
  creds: |
    {
      "api_key": "$VULTR_API_KEY"
    }
---
apiVersion: vultr.m.upbound.io/v1beta1
kind: ClusterProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: vultr-credentials
      key: creds

```

- Create Modelplane Manifests:

```yaml
---
kind: InferenceGateway
metadata:
  name: default
  namespace: crossplane-system
spec:
  backend: Traefik
  traefik:
    version: "40.2.0"
    loadBalancer: MetalLB
    metallb:
      addressPool: "172.18.255.200-172.18.255.250"
---
kind: InferenceClass
metadata:
  name: a16-1x
  namespace: crossplane-system
spec:
  description: "Vultr vcg-a16-6c-64g-16vram, NVIDIA A16 16GB"
  provisioning:
    provider: Vultr
    vultr:
      plan: vcg-a16-6c-64g-16vram
      accelerator:
        type: nvidia-a16
        count: 1
  devices:
  - name: gpu
    claim: DRA
    driver: gpu.nvidia.com
    deviceClassName: gpu.nvidia.com
    count: 1
    attributes:
      architecture: { string: Ampere }
      cudaComputeCapability: { version: "8.6.0" }
    capacity:
      memory: { value: "16384Mi" }
---
kind: InferenceCluster
metadata:
  name: vultr-demo
  namespace: crossplane-system
  labels:
    modelplane.ai/region: ewr
spec:
  cluster:
    source: Vultr
    vultr:
      region: ewr
  nodePools:
  - name: gpu-a16
    className: a16-1x
    nodeCount: 1
    minNodeCount: 1
    maxNodeCount: 1
---
apiVersion: v1
kind: Namespace
metadata:
  name: ml-team
---
kind: ModelDeployment
metadata:
  name: qwen-7b
  namespace: ml-team
spec:
  replicas: 1
  template:
    spec:
      engines:
      - name: qwen-7b
        members:
        - role: Standalone
          nodeSelector:
            devices:
            - name: gpu
              count: 1
              selectors:
              - cel: |
                  device.capacity["gpu.nvidia.com"].memory.compareTo(quantity("14Gi")) >= 0
          template:
            spec:
              containers:
              - name: engine
                image: vllm/vllm-openai:v0.9.2
                args:
                - --model=Qwen/Qwen2.5-7B-Instruct-AWQ
                - --served-model-name=qwen-7b
                - --max-model-len=8192
                - --gpu-memory-utilization=0.85
                - --enforce-eager
---
kind: ModelService
metadata:
  name: qwen-7b
  namespace: ml-team
spec:
  endpoints:
  - selector:
      matchLabels:
        modelplane.ai/deployment: qwen-7b
```

- Tested Model + ModelEndpoint:

```json
{"id":"chatcmpl-5f3c5820-b281-4872-8ed0-9137b2890359","object":"chat.completion","created":1785133519,"model":"qwen-7b","choices":[{"index":0,"message":{"role":"assistant","reasoning_content":null,"content":"Hello from Vultr! It's nice to meet you. Vultr is a cloud computing platform that offers flexible and affordable virtual server instances. How can I assist you today with regards to Vultr or cloud computing in general?","tool_calls":[]},"logprobs":null,"finish_reason":"stop","stop_reason":null}],"usage":{"prompt_tokens":35,"total_tokens":81,"completion_tokens":46,"prompt_tokens_details":null},"prompt_logprobs":null,"kv_transfer_params":null}
```


docs/manifests/getting-started/vultr/platform.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: modelplane.ai/v1alpha1
```
negz, line 1: Are these getting started manifests used anywhere?

docs/content/platform/providers.md
```diff
@@ -53,7 +52,7 @@ native provisioning.
 | Scaleway (Kapsule) | {{< accel nvidia >}} | Planned | ✓ | {{< repolink "https://github.com/scaleway/crossplane-provider-scaleway" "crossplane-provider-scaleway" "official" >}} |
 | Tencent Cloud (TKE) | {{< accel nvidia >}} | Planned | ✓ | {{< repolink "https://github.com/crossplane-contrib/provider-tencentcloud" "provider-tencentcloud" "community" >}} |
 | Voltage Park | {{< accel nvidia >}} | Planned | ✓ | none yet |
-| Vultr (VKE) | {{< accel nvidia >}} {{< accel amd >}} | Planned | ✓ | {{< repolink "https://github.com/vultr/crossplane-provider-vultr" "crossplane-provider-vultr" "official" >}} |
+| Vultr (VKE) | {{< accel nvidia >}} {{< accel amd >}} | ✓ | ✓ | {{< repolink "https://github.com/upbound/provider-vultr" "provider-vultr" "community" >}} |
```
negz, line 55: Should this provider be in crossplane-contrib if it's not official? 🤔

docs/manifests/concepts/inference-cluster-vultr.yaml
```diff
@@ -0,0 +1,58 @@
+# An InferenceCluster backed by a Vultr Kubernetes Engine (VKE) cluster.
+#
+# Modelplane provisions the full VKE cluster (VPC, control plane, system +
+# GPU node pools) and installs the inference stack (cert-manager, Traefik,
+# Prometheus, KEDA, LeaderWorkerSet).
```
negz: Is Traefik part of the inference stack on the cluster? It isn't for other clouds.

functions/compose-inference-cluster/function/fn.py
```diff
@@ -1144,19 +1306,21 @@ def observed_cache_storage_class(self) -> str | None:
...
-            return self._observed_cluster_cache_class("gke-cluster", gkev1alpha1.GKECluster)
-        if cluster.source == CLUSTER_SOURCE_EKS:
-            return self._observed_cluster_cache_class("eks-cluster", eksv1alpha1.EKSCluster)
-        if cluster.source == CLUSTER_SOURCE_AKS:
-            return self._observed_cluster_cache_class("aks-cluster", aksv1alpha1.AKSCluster)
-        if cluster.source == CLUSTER_SOURCE_NEBIUS:
-            return self._observed_cluster_cache_class("nebius-cluster", nebiusv1alpha1.NebiusCluster)
+        if cluster.source in provisioned:
+            key, model = provisioned[cluster.source]
+            return self._observed_cluster_cache_class(key, model)
```
negz, line 1331: Nice 👍

functions/compose-vultr-cluster/function/fn.py
```diff
@@ -0,0 +1,568 @@
...
+        creds = self.xr.spec.credentials
+        return creds.type if creds and creds.type else "ClusterProviderConfig"
+
+    def _cred_name(self) -> str:
+        creds = self.xr.spec.credentials
+        return creds.name if creds and creds.name else "default"
+
+    def compose(self) -> None:
+        self.compose_cluster()
+        if self._cluster_ready():
```
negz: This means we'll stop composing all these things if the cluster becomes unready (e.g. temporarily) right? I believe in the other implementations we also check whether we've already composed the dependent things and if so keep doing so.

functions/compose-vultr-cluster/function/fn.py
```diff
@@ -0,0 +1,568 @@
...
+                    readiness=k8sobjv1alpha1.Readiness(
+                        policy="DeriveFromCelQuery",
+                        celQuery=_DS_READY_CEL,
+                    ),
+                    forProvider=k8sobjv1alpha1.ForProvider(manifest=manifest),
+                ),
+            ),
+        )
+
+    def compose_longhorn(self) -> None:
```
negz: I'm not familiar with Longhorn. I assume it could work for other providers that don't have native support for RWX PVs? Does it warrant a small design before we choose and commit to it? Are there alternatives we should consider?

I'm asking because I want to make sure we extend our tech stack intentionally. If Longhorn is the right choice I'm all for it, but I want to make sure we thought it through.

functions/compose-vultr-cluster/function/fn.py
```diff
@@ -0,0 +1,568 @@
...
+        self.rsp = rsp
+        self.xr = v1alpha1.VultrCluster(**resource.struct_to_dict(req.observed.composite.resource))
+
+    def _cred_kind(self) -> str:
+        creds = self.xr.spec.credentials
+        return creds.type if creds and creds.type else "ClusterProviderConfig"
+
+    def _cred_name(self) -> str:
+        creds = self.xr.spec.credentials
+        return creds.name if creds and creds.name else "default"
```
negz, line 146: Does InferenceCluster ever set `spec.credentials`? Should it?

docs/content/platform/inference-cluster.md
```diff
@@ -82,6 +85,10 @@ depends on the source:
 - **`GKE`** (Filestore Enterprise), **`EKS`** (EFS), **`AKS`** (Azure Files),
   and **`Nebius`** (shared filesystem): auto-provisioned. Those classes are
   fixed; nothing for the admin to do.
+- **`Vultr`** (Vultr File System): built into VKE - the Vultr CSI driver
+  installs the `vultr-vfs-storage` RWX class on every cluster, so nothing is
+  provisioned and nothing for the admin to do. Note Vultr File System isn't
+  available in every region.
```
negz: We're using Longhorn now instead right?

functions/compose-vultr-cluster/function/fn.py
```diff
@@ -0,0 +1,568 @@
...
+    def _node_pool(self, pool: v1alpha1.NodePool) -> vkenodepoolv1alpha1.VkeNodePool:
+        """Map an XR node pool to a VkeNodePool managed resource."""
+        labels = {_LABEL_POOL: pool.name}
+        if pool.role == "GPU" and pool.gpu:
+            labels[_LABEL_GPU] = pool.gpu.acceleratorType
+
+        fp = vkenodepoolv1alpha1.ForProvider(
+            label=pool.name,
+            plan=pool.plan,
+            nodeQuantity=pool.nodeCount,  # ty: ignore[invalid-argument-type]  # the XRD defaults nodeCount
```
negz: How does this interact with the autoscaler? Does it work fine when both are set? (Should this be in initProvider?)

functions/compose-vultr-cluster/function/fn.py
negz: I did a little research. Any idea what its performance is like?

Seems like there could be some quirks, namely:

* I think it allocates up to 70% of node disk for storage, so the model weights need to fit in that.
* The NFS `share-manager` funnels all traffic through a single node - so N engines will all load from one node, capped at that node's network bandwidth.

(I'm trusting an agent's analysis here, so I could be wrong.)

docs/manifests/concepts/inference-cluster-vultr.yaml
haarchri: this PR was rebased, i will remove it from AKS example too, as we removed this traefik line in some other PRs https://github.com/modelplaneai/modelplane/pull/347 .. good catch

docs/content/platform/providers.md
haarchri: we have the provider-vultr in upbound org for now, we need to figure out which provider we will use on the long-run https://github.com/vultr/crossplane-provider-vultr (upjet based but no official release since years and behind the TF version), https://github.com/crossplane-contrib/provider-vultr (native, archived since 2025) and our published provider https://github.com/upbound/provider-vultr (native)

functions/compose-vultr-cluster/function/fn.py
haarchri: we implemented this via https://github.com/modelplaneai/modelplane/pull/366 - so folks can configure a ClusterProviderConfig or a ProviderConfig, we did this to support multiple authentifications if you want to run in multiple Regions/Subscriptions/Projects etc.

functions/compose-vultr-cluster/function/fn.py
haarchri: The 70% is _storage-reserved-percentage-for-default-disk_ (default 30% reserved). It is a setting, not a hard limit - and we using here bare metal nodes with lot of space

The share-manager funnel is real. Longhorn attaches the volume RWO to one node and re-exports it over NFS. All engines read through that one pod, capped at its NIC. It is per volume, so one PVC per model spreads it a bit...

The thing is, this is not really a Longhorn problem. AWS EFS for example caps at 500 MiB/s to 1.5 GiB/s per client, and in Bursting mode a 200 GB filesystem drops to 10 MiB/s once credits run out. Vultr does have native RWX (VFS), but on bare metal it goes through a single Storage Gateway, same shape as share-manager. Every option we have funnels somewhere....

So what we actually need is a bar: N engines, model size S, ready within T. Whatever backend cannot hit it does not ship, whether it is Longhorn, EFS or VFS. If none of them clear it, that tells us to stop using it and start thinking of something different instead? 

I would treat Longhorn as an implementation detail for now, not a commitment. It sits behind the ModelCache backend and I expect we replace it. If VFS covers our gpu node pools we can drop it ...

functions/compose-vultr-cluster/function/fn.py
haarchri: tested without any issues - the autoscaler is taking care - and the api backend is not changing something

functions/compose-vultr-cluster/function/fn.py
negz: > The thing is, this is not really a Longhorn problem.

Isn't being capped to a single node's NIC speed a problem?

We know with EFS that we hit approx 1.4GB/s _per engine_ load, with two engines loading simultaneously:

https://github.com/modelplaneai/modelplane/blob/74b893082bd1bbad00308c414cee2e2bc95fb00a/docs/manifests/examples/kimi-k2/model-deployment.yaml#L14-L15

I believe there's a 20GB/s limit at the sender end, so we could have ~14 engines loading simultaneously before we started to see that slow down.

We also know (per https://github.com/modelplaneai/modelplane/issues/204) that vLLM can pull directly from HuggingFace at around 500MB/s (per engine).

> So what we actually need is a bar: N engines, model size S, ready within T.

That's a good idea, and the kind of thing stepping back and thinking about design would help us align on.

For now, what I want to avoid is adding ModelCache support that actually makes start time slower in most cases. I can't find any published NIC speeds for Vultr, but assuming 10G NICs that's about about a 1GB/s ceiling for all engines to read from cache, right? So with 2 engines it's about as fast as loading directly from HF, and over 2 engines it's slower? Happy to be wrong here - I don't have access to test so I'm estimating.

I don't think ModelCache support necessarily needs to block Vultr support. If Longhorn is fast enough to be useful, great. If not, I'd rather ship Vultr without ModelCache support until we have a better cache solution.

functions/compose-vultr-cluster/function/fn.py
negz: Right, and in https://github.com/modelplaneai/modelplane/pull/366/changes#diff-f71d8713b29478f937d437e34ef5c69ec103e5f6ee0053d590238960f84a41a6 you've:

1. Added a `credentials` block under `aks`, `eks`, etc on the `InferenceCluster` XR
2. Updated `compose-inference-cluster` to propagate those creds to the underlying `AKSCluster`, etc XRs

I don't see a `credentials` block under the new `vultr` block on `InferenceCluster` though. Nor do I see code to plumb Vultr credentials in `compose-inference-cluster`. So these credentials can only be set if you create a `VultrCluster` XR directly? Is that intentional?

functions/compose-vultr-cluster/function/fn.py
haarchri: Agreed, and I am fine dropping Longhorn here. It was meant as an implementation detail...

I would rather split this into two things.
1. Vultr ships without ModelCache. Better to ship a provider with a documented gap. This connects to #331, if we write a contract for what every provider implementation must deliver, I think "cache StorageClass" should be optional and declared, not mandatory. Some clouds will not have a good RWX story for a while, and blocking the whole provider on it means we ship nothing. What matters is that the gap is visible in the docs and in the API, so users know what they get.

2. A separate design for shared model storage. Not "which RWX backend", the wider question.
- The bar, measured. N engines, model size S, ready within T. With real numbers from real clusters, not estimates. Right now we are all guessing.
- Is a shared filesystem the right shape at all? Every RWX option funnels through one point somewhere. So we would only be choosing between funnels. 
- HF in the scale-up path. If every scale-out pulls from HF, then HF rate limits or an HF outage means no scale-out. That should be a deliberate decision...
- Traffic cost. Without a cache we pull S per engine on every cold start and every scale-out. That is billed, per GB read on Storage, egress elsewhere. A cache can be slower on wall clock and still be right if it turns N pulls into 1

docs/content/platform/providers.md
haarchri: we using now: https://github.com/upbound/provider-upjet-vultr which is available via marketplace: https://marketplace.upbound.io/providers/upbound/provider-vultr/v1.0.0

functions/compose-vultr-cluster/function/fn.py
haarchri: opened https://github.com/modelplaneai/modelplane/issues/383

negz (approved):

PR #375, opened 2026-07-31
feat(nix): Finish control plane setup and surface failures in the nix run app

haarchri:
The nix run app started a control plane with `crossplane project run` and stopped there, which left it in a state that doesn't match what the getting-started guide produces. The prerequisites manifest, the modelplane-system namespace, composition RBAC, and provider-helm's DeploymentRuntimeConfig and ImageConfig, was never applied, so compositions that rely on that RBAC fail. Now the app applies `docs/manifests/getting-started/prerequisites.yaml` once the control plane is healthy.

Applying the manifest alone isn't enough for provider-helm, though: `crossplane project run` installs providers before the prerequisites exist, and Crossplane resolves ImageConfig runtime configs only at ProviderRevision creation. provider-helm therefore always comes up on the default runtime config, whose ServiceAccount lacks the RBAC granted above. The app now patches the provider's `spec.runtimeConfigRef` to point at `provider-helm-modelplane` explicitly, so it runs with the intended ServiceAccount regardless of install order.

Failures were also effectively undebuggable under nix.sh: installs that time out report only "context deadline exceeded", and the kind cluster lives inside the container's Docker daemon, so it's gone before anyone can inspect it. On failure the app now exports the kind kubeconfig and dumps `kubectl get pkgrev` (plus the full YAML) before dropping into the dev shell, and it propagates the failure as its exit code instead of swallowing it. The default `--timeout` is also raised from the CLI's 5m to 15m, since first-time image pulls through the Docker-in-Docker daemon regularly exceed 5m; an explicit `--timeout` on the command line still wins.

Fixes #

I have: 

### DIFF
--- a/nix/apps.nix
+++ b/nix/apps.nix
@@ -137,16 +137,58 @@
             rm -f _output/functions
             ln -s ${functionsPkg} _output/functions
 
-            crossplane project run "$@"
+            # The CLI's default --timeout of 5m covers waiting for the
+            # installed configuration to become healthy, which first-time
+            # image pulls through nix.sh's Docker-in-Docker daemon regularly
+            # exceed. Default to a longer wait; an explicit --timeout on the
+            # command line still wins.
+            timeout_args=(--timeout 15m)
+            for arg in "$@"; do
+              case "$arg" in
+                --timeout | --timeout=*) timeout_args=() ;;
+              esac
+            done
+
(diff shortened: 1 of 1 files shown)


negz (approved):
Thanks for fixing this! I'd been meaning to do it myself.

Issue #407, opened 2026-08-21
Load ModelCache weights by repo id, so an engine command is the same with or without a cache


negz:
### What problem are you facing?

An engine command in a ModelDeployment is coupled to whether its model is cached. A ModelCache hydrates its PVC with `hf download --local-dir`, which writes a flat model directory (`config.json` and `*.safetensors` at the root). vLLM can load that only by path, so a cached deployment has to set `--model=/mnt/models`, while an uncached deployment of the same model sets `--model=Qwen/Qwen2.5-7B-Instruct` and pulls from HuggingFace. Adding or removing a cache means rewriting the engine command, and one model spec can't move between the cached and uncached cases.

Pointing `HF_HUB_CACHE` at the flat dir doesn't bridge this: HuggingFace's repo-id resolver needs its own cache tree, so `--model=<repo>` + `HF_HUB_CACHE=/mnt/models` + `HF_HUB_OFFLINE=1` fails with `LocalEntryNotFoundError: Cannot find an appropriate cached snapshot folder`.

### How could Modelplane help solve your problem?

Hydrate into the HuggingFace cache layout rather than a flat dir, and inject the cache env when a cache is present, so `--model=<repo>` resolves from the PVC offline.

The hydration Job would run `hf download <repo>` into `HF_HUB_CACHE` (no `--local-dir`), producing `/mnt/models/models--<org>--<name>/snapshots/<sha>/`. compose-model-replica would set `HF_HUB_CACHE=/mnt/models` and `HF_HUB_OFFLINE=1` on any engine that references a cache — the ModelExpress path already injects both. The engine command is then identical whether or not a cache exists:

```yaml
containers:
- name: engine
  image: vllm/vllm-openai:v0.23.0
  args: [--model=Qwen/Qwen2.5-7B-Instruct]
```

With a cache, vLLM resolves the repo id to the local snapshot and loads offline; without one, it pulls from HuggingFace.

I validated the layout change on EKS. With the PVC hydrated in cache layout and the two env vars set, vLLM logs the resolution and serves:

```
HF_HUB_OFFLINE is True, replace model_id [Qwen/Qwen2.5-7B-Instruct] to model_path [/cache/models--Qwen--Qwen2.5-7B-Instruct/snapshots/a09a354...]
```

Both the default loader and `--load-format modelexpress` load from the cache this way. runai_streamer should too, since the repo-id→path swap happens in vLLM before the loader runs, but I didn't test it.

negz:
[...]

Issue #115, opened 2026-06-09
Build a Modelplane-owned ModelCache hydration image


dennis-upbound:
The ModelCache hydration Job currently runs `python:3.11-slim` and `pip install huggingface_hub` at runtime before `hf download`. A prebuilt, Modelplane-owned image with the tooling preinstalled would be faster, pinned/reproducible, and remove the network dependency on PyPI at hydration time. It also gives a place to preinstall other source tools (e.g. `aws` for a future S3 source).

Towards #66.

PR #411, opened 2026-08-26
WIP: Add an OCI source to ModelCache

dennis-upbound:

A `ModelCache` could only stage from HuggingFace, so a customer whose models already live in their own registry had no way to point Modelplane at them, and no way to serve from a registry they control.

This adds `spec.source: OCI` with an `oci` sibling carrying a registry `ref` and an optional `pullSecret`. The hydration Job pulls the artifact with `modctl` and extracts it onto the same cache volume a HuggingFace source stages to, so nothing downstream changes: `compose-model-replica` mounts the same PVC at the same path and the engine reads files the same way.

Everything source-specific reduces to four questions — which credential, how much capacity, which hydration variant, and what to name in an event — so the composition below them reads the same either way. A hydration variant is an image, an env list and a command, which is why the OCI path runs `modctl` out of `alpine` while HuggingFace keeps running `huggingface_hub` out of `python`.

```yaml
kind: ModelCache
metadata:
  name: qwen25-05b-oci
  namespace: ml-team
spec:
  source: OCI
  oci:
    ref: us-docker.pkg.dev/my-project/my-repo/qwen2.5-0.5b-instruct:modelpack-v1
    sizeGiB: 5
    pullSecret:
      name: registry-creds     # optional; a kubernetes.io/dockerconfigjson Secret
```

### What ran

Proven end to end on a real GKE cluster: a private registry through to a completion from the loaded model.

Qwen2.5-0.5B-Instruct (988MB of BF16 safetensors, 290 tensors) was packed with `modctl build` and pushed to Artifact Registry. A `ModelCache` with `source: OCI` and a `pullSecret` then hydrated it on the workload cluster:

```
IO summary: 953 MiB in 9.805s, 97.23 MB/s
Successfully pulled model artifact: us-docker.pkg.dev/.../qwen2.5-0.5b-instruct:modelpack-v1
```

The volume held `model.safetensors` at 988,097,824 bytes — byte-identical to the source — alongside the config and tokenizer files, at `/mnt/models`.

vLLM 0.23.0 then loaded from that volume on an NVIDIA L4 bound through DRA:

```
Starting to load model /mnt/models...
Filesystem type for checkpoints: NFS. Checkpoint size: 0.92 GiB.
Loading weights took 6.09 seconds
```

and served a real completion (`"content": "loaded from OCI"`, `finish_reason: "stop"`).

Two things this exercises that were previously only unit-tested. The `pullSecret` path ran against a private registry with real credentials, through the composition's Secret propagation to the workload cluster. And the engine's args came out as `["--served-model-name=qwen2.5-0.5b","--max-model-len=16384","--model=/mnt/models"]` — no `--model` was written by hand, so `apply_cache_args` injected it.

`test-compose-model-cache`, `ty-compose-model-cache` and the `python` check pass. Fifteen cases, two of them new. The local `e2e` harness also gained an `--oci` flag that stages a cache and asserts the artifact's files land on the volume; it is off by default and out of the CI gate, because it needs an artifact in a registry the workload cluster can pull from and so is not hermetic.

### What lands on the volume

An OCI cache is read by path, not by repo id. `modctl` records a filepath annotation per descriptor and extract restores those paths, subdirectories included, so an artifact built over a model directory yields that same directory rather than a HuggingFace hub cache tree. An OCI reference also carries no HuggingFace repo id to build a tree around. So the property #407 is after — one engine command whether or not a cache exists — holds for a HuggingFace source and not for this one.

That interacts with #407 in a way worth settling before either lands. #407 asks to hydrate into the hub layout and to set `HF_HUB_CACHE` and `HF_HUB_OFFLINE` on any engine referencing a cache. The second cannot be unconditional once an OCI source exists: an OCI volume is not a hub cache tree, so an engine reading it by path is fine, but one handed a repo id with `HF_HUB_OFFLINE=1` would fail to resolve. The env injection has to be conditioned on the cache's source, which means `compose-model-replica` needs to know something about the cache beyond its volume name — today it deliberately knows nothing.

`modctl` also still writes the pre-donation media types: `artifactType: application/vnd.cnai.model.manifest.v1+json`, layers `vnd.cnai.model.weight.v1.raw`, annotations `org.cnai.model.filepath`, where the ModelPack spec documents `vnd.cncf.model.*`. The version is pinned for that reason, and anything that later switches on `artifactType` has to accept both.

### `sizeGiB` is close to meaningless on GKE

Worth raising because the GKE run made it concrete. The composed `modelplane-rwx` StorageClass is Filestore at `tier: enterprise`, whose minimum share is 1TiB. `sizeGiB: 5` provisioned a **1024 GB** Filestore instance — no PVC error, roughly ten minutes, and a bill for a terabyte to cache a 1GB model.

So `sizeGiB` reads like it controls cost and does not, below 1TiB on GKE. That also undercuts deriving the size from the manifest by summing descriptor bytes: the platform rounds that up by two orders of magnitude, so tier choice is the lever, not sizing logic. The PVC is also `ReadWriteMany` unconditionally, which a Standalone single-node engine does not need.

I would rather settle this than keep `sizeGiB` required, but it is a separate change from adding the source.

### Drive-by fix

The `SOURCE` printer column read `.spec.source.huggingFace.repo` against a `source` that is a string, so it never resolved. It now reads `.spec.source`.

### Still open

`#115` is now a prerequisite rather than a follow-up. The image stops being one image, and neither source can hydrate air-gapped today, because both fetch a tool before they fetch the artifact.

I have:

- [x] Run the affected checks (`test-compose-model-cache`, `ty-compose-model-cache`, `python`) and made sure they pass.

### DIFF
--- a/apis/modelcaches/definition.yaml
+++ b/apis/modelcaches/definition.yaml
@@ -17,7 +17,7 @@ spec:
     additionalPrinterColumns:
     - name: SOURCE
       type: string
-      jsonPath: .spec.source.huggingFace.repo
+      jsonPath: .spec.source
     - name: READY
       type: string
       jsonPath: .status.summary.ready
@@ -32,10 +32,12 @@ spec:
             x-kubernetes-validations:
             - rule: "self.source != 'HuggingFace' || has(self.huggingFace)"
               message: spec.huggingFace is required when spec.source is HuggingFace.
+            - rule: "self.source != 'OCI' || has(self.oci)"
+              message: spec.oci is required when spec.source is OCI.
             properties:
               source:
                 type: string
-                enum: [HuggingFace]
+                enum: [HuggingFace, OCI]
                 description: >-
                   Which kind of artifact source to stage from. The
                   matching source object (e.g. spec.huggingFace) must
@@ -71,6 +73,43 @@ spec:
                       key:
                         type: string
                         default: HF_TOKEN
+                  sizeGiB:
+                    type: integer
+                    description: >-
+                      Capacity to allocate for the staged artifact on
+                      each matched cluster.
+                    minimum: 1
+                    maximum: 100000
+              oci:
+                type: object
+                description: >-
+                  OCI source. Required when source is OCI. The
+                  hydration Job pulls the artifact with modctl and
+                  extracts it onto the cache volume, restoring the
+                  file layout the artifact was built with.
+                required: [ref, sizeGiB]
+                properties:
+                  ref:
+                    type: string
+                    description: >-
+                      Registry reference to the model artifact, by tag
+                      or digest (e.g. registry.example.com/models/
+                      qwen2.5-0.5b:v1).
+                    minLength: 1
+                  pullSecret:
+                    type: object
+                    description: >-
+                      Optional Secret holding registry credentials for
+                      a private repository. Names a
+                      kubernetes.io/dockerconfigjson Secret in the
+                      ModelCache's own namespace; Modelplane propagates
+                      it to each matched cluster for the hydration Job
+                      to read.
+                    required: [name]
+                    properties:
+                      name:
+                        type: string
+                        minLength: 1
                   sizeGiB:
                     type: integer
                     description: >-
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -74,6 +74,9 @@ against. The control-plane cluster needs no DRA.
 nix run .#e2e              # bring up both clusters + deploy the mock model
 nix run .#e2e -- --verify  # same, then wait for readiness and assert a live 200
 nix run .#e2e -- --clean   # tear both clusters down
+
+# additionally stage a `source: OCI` ModelCache and assert the artifact lands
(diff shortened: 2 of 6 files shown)


dennis-upbound:
Closing this until the design in upbound/inference#7 is approved — no point reviewing an implementation while the direction it implements is still open, and adopting OCI upstream is exactly the kind of decision that should land in the design first.

The branch `dennis/oci-source-poc` stays. Nothing here is lost: the description above records what the POC proved on a real GKE cluster (a ModelPack artifact in Artifact Registry, hydrated by `modctl` onto a Filestore volume, loaded by vLLM 0.23.0 on an L4, serving a completion), and the two things it settled that the design was guessing at — an OCI cache is read by path rather than as a Hugging Face hub tree, so #407's unconditional `HF_HUB_OFFLINE` cannot hold once an OCI source exists; and `modctl` still writes the pre-donation `vnd.cnai.*` media types where the ModelPack spec documents `vnd.cncf.*`.

I will reopen or raise a fresh PR once the design is settled.

PR #413, opened 2026-08-26
Set the project on the composed ProjectIAMMember

dennis-upbound:

Provisioning a GKE `InferenceCluster` fails at the IAM binding:

```
Request `Create IAM Members roles/container.admin
serviceAccount:...@<project>.iam.gserviceaccount.com for project ""`
returned error: Error retrieving IAM policy for project "":
googleapi: got HTTP response code 404
```

`compose_iam_binding` builds `forProvider` from `role` and `member` only, on an assumption written into `resolve_project`'s docstring:

> The GCP provider uses projectID from the ProviderConfig/ClusterProviderConfig as the default for all managed resources, so we only need the value explicitly for the workloadPool string in the GKE cluster spec.

That holds for the `Network`, `Subnetwork` and `ServiceAccount` — all three come back with `project` late-initialized into their specs — but not for `ProjectIAMMember`, which creates with an empty project before any late-init happens. `project` is optional on the CRD, so nothing catches the omission before the call goes out.

`self.project` is already resolved for `workloadPool`, so the fix passes the same value.

The blast radius is the whole cluster, not just the binding: the XR stays at `Unready resources: gke-cluster`, so `BackendReady` never installs the serving stack and the `InferenceCluster` never reaches `Ready`.

The existing test encoded the broken shape — it asserted a `forProvider` of exactly `role` and `member` — so it is updated rather than added to.

### Two things found alongside it

**The GCP roles a credential needs were undocumented.** `build-the-platform` said only "permissions to create GKE clusters, VPCs, and IAM roles". Provisioning one cluster produced a separate 403 for each of seven roles, which are now listed. `roles/resourcemanager.projectIamAdmin` gets a callout: Modelplane grants the node service account `roles/container.admin`, so the provisioning credential must be able to set project IAM policy. That is a strong permission and a platform team should know it is required before handing a key over.

**The local e2e control plane cannot provision cloud clusters**, and `e2e/README.md` now says so. Both of `lean-control-plane.yaml`'s trims block it, and neither announces itself: unactivated managed resources report *no status conditions at all* (which reads as idle, not as an error), and the zero-replica `ImageConfig` is not undone by editing the `ImageConfig`, since it resolves only when a package revision reconciles.

### How this was found

Provisioning a real GKE cluster with an L4 pool to validate #411 and #412 against real vLLM. Patching `spec.forProvider.project` on the live managed resource unblocked provisioning, which confirmed the cause before this change was written.

After the fix the same path completed: cluster and node pools provisioned, serving stack installed, and a vLLM engine served a completion.

`test-compose-gke-cluster`, `ty-compose-gke-cluster` and the repo-wide `python` check pass.

### What is not covered

The composition grants the node service account `roles/container.admin` at project scope. That is broader than a node needs and is worth revisiting, but it is existing behaviour and out of scope here.

I have:

- [x] Run the affected checks (`test-compose-gke-cluster`, `ty-compose-gke-cluster`, `python`) and made sure they pass.

### DIFF
--- a/docs/content/getting-started/build-the-platform.md
+++ b/docs/content/getting-started/build-the-platform.md
@@ -24,14 +24,14 @@ against this capacity without knowing which cluster it runs on.
   | Role | Needed for |
   |---|---|
   | `roles/container.admin` | the cluster and its node pools |
-  | `roles/compute.admin` | the VPC network and subnetwork |
+  | `roles/compute.admin` | the VPC network and subnet |
   | `roles/serviceusage.serviceUsageAdmin` | enabling the APIs the cluster needs |
   | `roles/iam.serviceAccountAdmin` | the node service account |
   | `roles/iam.serviceAccountKeyAdmin` | the node service account's key |
   | `roles/iam.serviceAccountUser` | attaching that account to the nodes |
   | `roles/resourcemanager.projectIamAdmin` | granting the node account `container.admin` |
 
-  The last one is worth a look before you hand the key over: Modelplane grants
+  The last one is worth a look before you hand the key over. Modelplane grants
   the node service account `roles/container.admin`, so the credential doing the
   provisioning has to be able to set project IAM policy.
 {{< /tab >}}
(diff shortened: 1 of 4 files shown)


haarchri (approved):
good Catch, please check vale wording and then ready to go

PR #420, opened 2026-09-01
Assert what the served endpoints return, not just a 200

### LINKED ISSUE #368: Local e2e asserts only a 200; add behavioral test-suites against the running control plane

The local end-to-end test added in #365 brings the full path up on `kind` and asserts a live 200 through the control-plane gateway on both `/v1/chat/completions` and `/v1/messages`. That proves the path is wired and routing works — but it stops at "the endpoint answers." It doesn't assert *behavior*: response shapes, error paths, streaming, multi-turn, or the semantics the compositions are meant to guarantee. A regression that still returns 200 with a wrong body would pass unnoticed.

Suggested by @haarchri in the #365 review (https://github.com/modelplaneai/modelplane/pull/365#discussion_r3639339051): once the project is running in `kind`, run a Python test-suite against the published `ModelService` address to assert real behavior rather than just status codes. It would run in the same cloud-free harness and could grow to cover the OpenAI and Anthropic surfaces, error handling, and streaming.

This pairs naturally with the CPU-based-model idea (also raised on #365): the current mock returns canned responses, so meaningful behavioral assertions want either a richer mock or a small CPU model to exercise.

Follow-up to the local e2e added in #365. Related: #367 (KWOK nodes for scheduler-placement testing).

nicknikolakakis:
Fixes #368.

The local e2e brings the whole path up and then checks one thing about it: that `/v1/chat/completions` and `/v1/messages` return 200. A regression that answers with the wrong shape, loses the request body between the gateway and the engine, or buffers a stream into a single blob passes that check.

The mock engine is what limited it: one canned response to any POST, body ignored, so there was nothing else to assert. It now echoes the last user turn and the requested model, streams `chat.completion.chunk` frames terminated by `[DONE]` when a request sets `stream`, and returns 400 for an unparseable body or one with no `messages` and 404 for an unknown path. On top of that, the two curl pods become a stdlib `unittest` suite in `e2e/verify`, run from a pod on the control plane because the address is on the kind Docker subnet. It ships as a ConfigMap and runs under the same python image the mock uses, so nothing is built or installed.

The suite fails 5 of its 8 cases against the pre-change mock, which is what says it tests behaviour rather than reachability. I verified the pod wiring on a throwaway kind cluster: green against the new mock with the pod reaching `Succeeded`, and a deliberately wrong `MODELPLANE_MODEL` putting it in `Failed`, which is what makes `run.sh` exit non-zero. I have not run the full two-cluster e2e, so this needs the `test-e2e` label to be proven end to end.

The mock is about twice as long and still inline in `args`, since the ModelDeployment container template is a curated subset with no volume to load a file from. `env.valueFrom` from a ConfigMap is the way out if it grows further.

I have:

- [ ] ~Added or updated tests covering any composition function changes.~ This PR touches no composition functions; the tests it adds are the e2e serving suite in `e2e/verify`.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -92,7 +92,7 @@ against. The control-plane cluster needs no DRA.
 
 ```bash
 nix run .#e2e              # bring up both clusters + deploy the mock model
-nix run .#e2e -- --verify  # same, then wait for readiness and assert a live 200
+nix run .#e2e -- --verify  # same, then wait for readiness and run the behavioral suite
 nix run .#e2e -- --clean   # tear both clusters down
 ```
 
@@ -117,10 +117,16 @@ kubectl run curl -n ml-team --rm -it --image=curlimages/curl@sha256:7c12af72ceb3
   -d '{"model":"mock","max_tokens":16,"messages":[{"role":"user","content":"hi"}]}'
 ```
 
-`--verify` runs both of those (OpenAI then Anthropic) and exits non-zero on
-failure. It's the exact command the `E2E` CI workflow runs, so a green `--verify`
-locally and a green CI run mean the same thing; use the manual curls above to
-poke the endpoints interactively.
+`--verify` goes further than those curls: it runs [`verify/test_serving.py`](verify/test_serving.py)
+from a pod on the control plane and asserts what came back — the response shapes
+both surfaces promise, that the last user turn reaches the engine, that a
+streamed response arrives as `text/event-stream` frames rather than one buffered
+blob, and that malformed input and unrouted paths fail. The suite goes in as a
+ConfigMap and runs under a stock `python:3.12-alpine` pod; it imports only the
+stdlib, so nothing is built or installed. It exits non-zero on failure and is the
+exact command the `E2E` CI workflow runs, so a green `--verify` locally and a
+green CI run mean the same thing; use the manual curls above to poke the
+endpoints interactively.
 
 ## How it's structured
 
--- a/e2e/manifests/40-model-deployment.yaml
+++ b/e2e/manifests/40-model-deployment.yaml
@@ -37,6 +37,8 @@ spec:
                 - |
                   import json, http.server
                   class H(http.server.BaseHTTPRequestHandler):
+                      # HTTP/1.1 so the streaming path can chunk, the way a real engine does.
+                      protocol_version = "HTTP/1.1"
                       def _s(self, o, c=200):
                           b = json.dumps(o).encode()
                           self.send_response(c)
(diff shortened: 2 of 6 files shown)


haarchri:
can you please add the checklist, rebase and run the test one time locally that will help to review the PR after that we can add the `test-e2e` label - thanks

nicknikolakakis:
Checklist added to the description, and rebased onto main. Clean, no conflicts, and the diff is unchanged even though `4e3b603` touched `e2e/run.sh` and `798fcd1` touched `nix/checks.nix`.

On running it locally: `./nix.sh flake check` passes on the rebased branch, which covers the `ruff` and license checks now that `e2e/verify` is in their scope (it checks `aarch64-linux` in the container and skips `aarch64-darwin`). The two-cluster e2e itself I still cannot run here, since `crossplane project run` needs the Nix-built function images. What I did verify is the pod wiring on a throwaway kind cluster: the suite goes green against the new mock with the pod reaching `Succeeded`, a deliberately wrong `MODELPLANE_MODEL` puts it in `Failed`, and the suite fails 5 of its 8 cases against the old mock. So the `test-e2e` label is still what would prove the full path.

haarchri:
tests are failing with:

```
==> Running the behavioral suite (e2e/verify) against http://172.18.255.200/ml-team/mock
configmap/e2e-verify-suite created
pod/e2e-verify created
setUpModule (__main__) ... ERROR

======================================================================
ERROR: setUpModule (__main__)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/suite/test_serving.py", line 89, in setUpModule
    raise AssertionError(
AssertionError: http://172.18.255.200/ml-team/mock did not serve a chat completion within 300s (last status 400)

----------------------------------------------------------------------
Ran 0 tests in 300.376s

FAILED (errors=1)
verify: the behavioral suite did not pass (pod phase: Failed)
Error: Process completed with exit code 1.
```

nicknikolakakis:
Thanks, that log had everything needed. The 400 was the mock engine's own, and the bug is mine.

From the run's `e2e-diagnostics` artifact, every readiness attempt in the gateway's access log:

```
"response_code":400, "response_code_details":"via_upstream",
"bytes_received":70, "bytes_sent":33,
"x-envoy-origin-path":"/v1/chat/completions",
"upstream_host":"10.244.0.25:8000"
```

`via_upstream` puts the 400 at the engine rather than the proxy. `bytes_received: 70` is the readiness payload exactly, and `bytes_sent: 33` is exactly `{"error": "messages is required"}`, which only the mock emits. The EPP agreed: `"modelName":"mock","targetModelName":"mock"` on every request, so routing and model resolution were never the problem.

The cause is that the endpoint picker is an ext_proc filter, and Envoy strips `content-length` from a request it hands to one unless `allow_content_length_header` is set, which defaults to false. The body reaches the engine chunked. The mock read it by `content-length` only, so it saw zero bytes, parsed `{}`, found no `messages`, and answered 400 to everything. The old mock never noticed because it ignored the body.

The mock now decodes a chunked body too. Reproduced and checked locally by running the engine straight out of the manifest:

| request | before | after |
| --- | --- | --- |
| POST with content-length | 200 | 200 |
| POST chunked, the shape CI sends | 400 | 200 |
| chunked stream, Anthropic, `/v1/models` | - | 200 |
| chunked malformed body / no messages | 400 | 400 |
| unknown path | 404 | 404 |

The suite is 8/8 against the fixed mock, and `./nix.sh flake check` passes. It is folded into the mock commit rather than added on top, per CONTRIBUTING. Ready for another `test-e2e` run whenever you are.

PR #421, opened 2026-09-01
Bump the e2e workload node image past the DRA idle deadlock

### LINKED ISSUE #419: Pinned kindest/node v1.34.0 carries the kubelet DRA idle deadlock (kubernetes#133926, fixed in v1.34.2)
### What happened?

The pinned workload-cluster node image `kindest/node:v1.34.0` ([`e2e/run.sh#L20`](https://github.com/modelplaneai/modelplane/blob/main/e2e/run.sh#L20)) ships a kubelet with a DRA deadlock: kubernetes/kubernetes#133926, fixed in v1.34.2 via kubernetes/kubernetes#133934.

Once the kubelet-to-DRA-driver gRPC connection sits idle for ~30 minutes, the next pod that references a ResourceClaim hangs in `ContainerCreating` forever:

- pod has a single `Scheduled` event, nothing else, no kubelet errors
- the ResourceClaim is `allocated,reserved`
- the DRA driver only ever logs `numClaims=0` health probes, `NodePrepareResources` never arrives

A goroutine dump from the node (`kubectl get --raw /api/v1/nodes/<node>/proxy/debug/pprof/goroutine?debug=1`) shows the deadlock: the pod worker blocked in `dra.PrepareResources -> NodePrepareResources -> grpc idle.ExitIdleMode`, while `http2Client.Close -> monitoredPlugin.HandleConn` (`dra_plugin_manager.go:118`) `-> Connect -> ExitIdleMode` holds the other side.

The nasty part: CI never hits this (clusters never idle 30 minutes), while any long-lived dev cluster always does. Overnight cluster, first `ModelDeployment` of the morning, engine pods stuck with zero diagnostics.

### How can we reproduce it?

1. Create a kind cluster with the pinned image from `e2e/run.sh`
2. Install a DRA driver (I used [dra-example-driver](https://github.com/kubernetes-sigs/dra-example-driver); the deadlock is in the kubelet, not the driver)
3. Schedule one pod with a ResourceClaim so the kubelet opens the driver connection, then leave the cluster idle for 30+ minutes
4. Create a `ModelDeployment` (or any pod referencing a ResourceClaim): it sticks in `ContainerCreating` with only a `Scheduled` event

Workaround: `docker exec <node> systemctl restart kubelet`, pods run within ~75s.

Suggested fix: bump the pin to a v1.34.2+ image. The open question is the containerd side: the current pin was introduced by #315 because newer kind images' containerd "has a regression that fails Modelplane", but the PR does not record what actually breaks. `kindest/node:v1.34.8` ships containerd 2.3.1 and I have not tested it against Modelplane. If someone remembers the concrete failure from #315, that would tell us whether a straight bump is safe or whether the containerd issue needs solving first.

Full writeup with the goroutine evidence: https://srekubecraft.io/posts/modelplane-fleet-inference/ (demo: https://github.com/nicknikolakakis/srekubecraft-demo/tree/main/modelplane-fleet)

Modelplane version: v0.3.1
Crossplane: v2.4.0
Kubernetes: kind v0.32.0, `kindest/node:v1.34.0@sha256:7416a61b42b1662ca6ca89f02028ac133a309a2a30ba309614e8ec94d976dc5a` (containerd 2.1.3)
DRA driver: kubernetes-sigs/dra-example-driver (synthetic `gpu.example.com` devices)
Providers: provider-helm, provider-kubernetes (upbound org, as shipped with v0.3.1)
Platform: Docker Desktop on macOS (arm64)

nicknikolakakis:
Fixes #419.

The e2e workload cluster runs `kindest/node:v1.34.0`, whose kubelet deadlocks once its connection to a DRA driver has been idle for 30 minutes: the next pod referencing a ResourceClaim sits in `ContainerCreating` with one `Scheduled` event and nothing else, and `NodePrepareResources` never reaches the driver. The goroutine dump and the reproduction are in the issue. Fixed upstream by kubernetes/kubernetes#133934, which `CHANGELOG-1.34.md` lists under `v1.34.2`.

The pin's stated reason is the DRA APIs, GA in 1.34, so any 1.34 image satisfies it. The containerd constraint you may remember is a different pin: #315 changed only `docs/content/getting-started/installation.md`, where the comment says kind v0.31+ ships containerd 2.2.0. `e2e/run.sh` got its pin from the e2e PR instead and never carried that constraint. containerd 2.2.0 was also one release: v1.34.3 has 2.2.0, v1.34.8 has 2.3.1, v1.34.11 has 2.3.4, each checked with `docker run --rm --entrypoint containerd <image> --version`.

So this takes v1.34.8, the oldest published image clearing both the deadlock and 2.2.0. v1.34.11 is there if you would rather track kind's latest. I have not run the full e2e locally, so whether Modelplane is happy on containerd 2.3.x is what the `test-e2e` label would settle. The docs pin is untouched: that cluster runs no DRA driver, so the deadlock doesn't reach it.

The reuse guard rejects a v1.34.0 or v1.34.1 cluster now instead of reusing it, since long-lived dev clusters are the population that hits this and bumping only the image would leave them deadlocking.

Touches `e2e/run.sh` like #420 does, on different lines.

I have:

- [ ] ~Added or updated tests covering any composition function changes.~ This PR changes only `e2e/run.sh`, no composition functions.

### DIFF
--- a/e2e/run.sh
+++ b/e2e/run.sh
@@ -17,7 +17,8 @@ WL=modelplane-e2e-workload
 # Pinned so the workload cluster has the DRA APIs the serving stack's NVIDIA DRA
 # driver needs (resource.k8s.io, GA in k8s 1.34). The control-plane cluster that
 # project run creates needs no DRA, so its image doesn't matter here.
-WL_NODE_IMAGE=kindest/node:v1.34.0@sha256:7416a61b42b1662ca6ca89f02028ac133a309a2a30ba309614e8ec94d976dc5a
+# v1.34.2 or newer: older kubelets deadlock on an idle DRA connection (k/k#133934).
+WL_NODE_IMAGE=kindest/node:v1.34.8@sha256:02722c2dedddcfc00febf5d27fbeb9b7b2c14294c82109ff4a85d89ac9ba3256
 METALLB_URL=https://raw.githubusercontent.com/metallb/metallb/v0.14.8/config/manifests/metallb-native.yaml
 # Pinned by digest (a multi-arch manifest list) so a moving :latest can't flake
 # the verify curl pod.
@@ -50,6 +51,10 @@ if kind get clusters 2>/dev/null | grep -qx "$WL"; then
 	# older one lacks the DRA APIs and would fail the run confusingly later.
 	ver="$(kubectl --context "$WLCTX" get nodes -o jsonpath='{.items[0].status.nodeInfo.kubeletVersion}' 2>/dev/null || true)"
 	case "$ver" in
+	v1.34.0 | v1.34.1)
+		echo "workload cluster $WL is $ver, whose kubelet deadlocks on an idle DRA connection (fixed in v1.34.2); recreate it with: nix run .#e2e -- --clean" >&2
+		exit 1
+		;;
 	v1.34.*) log "Reusing workload cluster $WL ($ver)" ;;
 	*)
 		echo "workload cluster $WL is ${ver:-unreachable}, but v1.34 is required for the DRA APIs; recreate it with: nix run .#e2e -- --clean" >&2


haarchri:
thanks for the bump, can you please add the checklist and rebase ? i will add the `test-e2e` label after

nicknikolakakis:
Rebased onto main, no conflicts, the diff is unchanged.

The checklist is already at the bottom of the description: three checked, and the composition-function one struck through since this only touches `e2e/run.sh`. Let me know if you wanted it in a different form.

haarchri (approved):

Issue #279, opened 2026-06-23
Nvidia AICR as a Backend for Serving Stack Generation


haarchri:
### What problem are you facing?

Looking at the current compose-serving-stack function, Modelplane already contains logic for generating and assembling AI serving infrastructure inside Kubernetes Clusters.

As [NVIDIA AICR](https://github.com/nvidia/aicr) matures, it could become an interesting backend for parts of this implementation instead of maintaining custom manifest generation logic in Modelplane itself.

### How could Modelplane help solve your problem?

- Reuse AICR validated and continuously maintained AI infrastructure manifests.
- Reduce the amount of platform-specific logic that needs to live inside Modelplane.
- Stay aligned with recommended deployment patterns for GPU Operators, networking, storage, and related AI platform components.
- Leverage AICR growing ecosystem rather than reimplementing infrastructure assembly logic.

```
 XR
    ↓
function-aicr
    ↓
AICR Recipe
    ↓
Generated Resources (releases.helm.crossplane.io)
```

This would allow Modelplane to focus on platform abstractions and workflows while relying on as the source of truth for AI infrastructure blueprints.

Curious whether maintainers see AICR as something that could eventually replace or simplify parts of the current serving-stack composition logic.

A possible direction would be a Composition Function that consumes AICR recipes/manifests and translates them into the resources expected by crossplane compositions.

mchmarny:
Hi, I'm one of the maintainers of [NVIDIA AICR](https://github.com/NVIDIA/aicr) and wanted to add some context that might be useful here.

The integration pattern @haarchri describes is pretty much what AICR is designed for, so a few things worth knowing as you think this through:

**Maintenance is something we take seriously.** AICR's recipe matrix (service x accelerator x intent) is continuously validated with KWOK-tiered tests on every PR and a nightly UAT cadence against live managed Kubernetes services. When a new GPU Operator release or driver version lands, the validated configuration updates in AICR rather than in each downstream consumer. That's the ongoing work you wouldn't have to carry inside Modelplane.

**We do hardware-aware optimization, not just manifest packaging.** AICR captures cluster state via snapshots and evaluates constraints before generating a bundle, things like driver compatibility, node topology, and component version alignment. The output is tuned to the actual hardware profile, not a generic template. That logic would otherwise need to live somewhere in Modelplane's composition pipeline.

[...]

PR #422, opened 2026-09-02
Design:  generating the serving stack at build time

haarchri:

Towards #279.

This adds a design proposing to generate the per-cloud component lists at build time instead of maintaining them by hand. Generated and hand-written sources feed the same shape: NVIDIA AICR recipes cover the parts NVIDIA maintains, and Modelplane's own components resolve alongside them for everything else, including clouds and accelerators AICR doesn't cover. A release then installs one tested stack per cloud, the same on every cluster, with nothing to override. spec.versions, spec.standard, spec.dynamo and nvidiaDriverRoot go away, replaced by a single field naming the cloud.

A prototype lives beside the document as evidence the pipeline works: generate.py resolves recipes into the module shape compose-serving-stack iterates, and stacks/ carries the hand-written halves (Nebius, Vultr, AMD).

Fixes #

I have: 


design/serving-stack-generation/generate.py
```diff
@@ -0,0 +1,530 @@
+# Copyright 2026 The Modelplane Authors.
```
negz: Drop the POC files/code from this design doc PR please. I think there's enough context (code snippets etc) in the design doc.

haarchri: done

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+- **No GPU driver setting.** Which kernel driver a GPU node runs, and whether
+  the node image or the GPU Operator provides it, is a generated value per
+  cloud. No API field reads or writes it, and changing it means a Modelplane
+  release. One driver version per cluster, whatever its pools run.
+- **No overrides.** No version field, no values override, no way to add,
+  remove or substitute a component on one cluster. A platform team that needs
+  a different stack takes the break-glass `Composition` and owns the result.
+- **No registry override.** Every cluster pulls charts and images from the
+  upstream repositories and registries the generated lists name. A cluster
+  that must pull from a private mirror, has no field for that either.
```
jonasz-lasut, line 40: Coming from mostly insurance/banking companies this part would be a real struggle to work around for most platform teams I've worked with. We've usually had a requirement that all helm charts/oci artifacts needs to come from a internal registry mirror and be signed and cve scanned by us, which this effectively blocks. Also usually helm charts don't come with all the security requirements/settings enabled by default like strict non-root security contexts, PDBs etc. and lack of overrides would force us to build a `break-glass` Composition even if the only goal would be to set the pdb.enabled value or update the security contexts. Platforms in highly-regulated environments are also usually network-restricted which means that without ability to override the image registry, we would not be able to run the modelplane serving stack at all without building the `break-glass` Composition.

In my personal opinion it's not a reasonable requirement to ask users to build their own compositions just to ensure the correct security posture of the serving stack if they could easily do it just by tweaking a few values

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+- **Upgrades ride releases.** Updating Modelplane updates the stack on every
+  cluster its control plane manages, at once, with no way to stage it. A
+  cluster cannot hold a component back, and a component cannot move without a
+  release, a patch release for a fix, a minor for everything else. The one
+  exception is the GPU driver on clouds whose node image supplies it: that
+  version follows the image, so it can move as nodes and pools cycle, with no
+  release involved.
```
jonasz-lasut, line 47: In regulated environments that will mean a Modelplane deployment per environment as usually critical component updates (such as serving stack) need to be tested in dev/test environments before they can be officially promoted to the production, this should be explicitly documented as it's a driving cost and effort factor. With serving stacks being disconnected from Modelplane version we could gradually rollout new versions of the serving stack across the environments using only one Modelplane instance

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+Four consequences are worth stating as bluntly as the design means them,
+because they are the price:
+
+- **No GPU driver setting.** Which kernel driver a GPU node runs, and whether
+  the node image or the GPU Operator provides it, is a generated value per
+  cloud. No API field reads or writes it, and changing it means a Modelplane
+  release. One driver version per cluster, whatever its pools run.
+- **No overrides.** No version field, no values override, no way to add,
+  remove or substitute a component on one cluster. A platform team that needs
+  a different stack takes the break-glass `Composition` and owns the result.
```
jonasz-lasut, line 37: This part also has a side-effect on BYO clusters in regulated environments. They may already come pre-provisioned with all the required compliance components such as cert-manager, prometheus stack connected to central observability system (e.g. Grafana cloud) and the default serving stack would try to pull in cert-manager and prometheus-stack with default values and lead to CRD ownership conflict, again forcing BYO cluster users to use `break-glass` Composition and lose the benefit of using the upstream

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+a composition reference nor a selector, the XRD doesn't default one, and
+Crossplane's selector resolver lists every compatible `Composition` and picks
+one at random. The fix is the XRD defaulting to the stock `Composition`, so
+installing a second one is inert until it's selected. Selecting it needs no new
+Modelplane API: a mutating admission webhook can set the composition reference
+on the `ServingStack` as Modelplane composes it, or a Crossplane `ImageConfig`
+can rewrite the `compose-serving-stack` function image to one the team owns, so
+the stock `Composition` runs their function. Both are advanced mechanisms, and
+deliberately so, this is an escape hatch to be reached for knowingly, not a
+feature to be configured.
```
jonasz-lasut, line 471: compose-serving-stack function should probably be reusable as a python library if that's the case so that end-users can import it and only do some minor changes instead of having to write their own full function from scratch if they only want to change the registry url or security contexts

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+buys a stack Modelplane has tested as a whole. A `ModelReplica` composes
+resources against the controllers the stack installs.
+
+Four consequences are worth stating as bluntly as the design means them,
+because they are the price:
+
+- **No GPU driver setting.** Which kernel driver a GPU node runs, and whether
+  the node image or the GPU Operator provides it, is a generated value per
+  cloud. No API field reads or writes it, and changing it means a Modelplane
+  release. One driver version per cluster, whatever its pools run.
```
jonasz-lasut, line 34: This turns a CUDA-major or GPU-generation boundary into "day of the next Modelplane release plus a full fleet upgrade" event, which may block a day 0 support for new models if they require a vLLM (or similar) version, which only runs on a new CUDA-major/GPU-gen

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+before it ships, and the reconcile path should hold no decisions of its own.
+
+## Proposal
+
+Everything is fixed at build time. Every version, every values block, every
+component membership decision is resolved where a human reviews a diff, and the
+reconcile path only renders. This is the one commitment the design treats as
+settled. Everything downstream of it, the intermediate format, the
+one-driver-per-cluster choice, is a consequence that can be revisited without
+touching it.
```
jonasz-lasut, line 178: This blocks day-0 support for new hardware, which would require an AICR version update, which has to come together with Modelplane update and a full serving stack update rolled-out across a full fleet

design/serving-stack-generation.md
```diff
@@ -0,0 +1,842 @@
...
+Plain string contracts run beside those API contracts. A replica reaches
+ModelExpress at `modelexpress-server:8001`, sets `schedulerName: kai-scheduler`,
+and labels pods `kai.scheduler/queue: modelplane` for a `Queue` the serving
+stack creates.
+
+Modelplane composes a `grove.io/v1alpha1` resource against a chart pinned at
+`v0.1.0-alpha.12-rc2`. Neither an alpha API nor a pre-release chart promises
+compatibility across versions.
+
+The coupling is growing. Grove and ModelExpress arrived with the Dynamo serving
```
jonasz-lasut, line 99: NVIDIA's own installation guide says Grove must be upgraded in lockstep with Dynamo while Grove APIs are unstable. A model whose day-0 support lands in a new Dynamo release therefore implies a Grove bump, a Modelplane release, and a fleet-wide upgrade. Which again means issues with no-effort day-0 model support, which we've stated in a blog post for Nemotron "without a line of new Modelplane code, because day-zero model support is built into the design"

jonasz-lasut:
General opinion
I think that the build-time approach that comes with no abbility to override the serving stack without writing a custom Composition will lead to most platform teams at regulated companies maintaining their serving stack composition and never using our code due to its lack of flexibility in terms of regulatory requirements such as registry mirrors, security posture etc.
It also makes day-0 support for models without code changes really hard to do when new model may require a new vLLM (or similar) which targets a new major CUDA version, driver version, Dynamo release (which needs a newer version of Grove) or new hardware, which the embedded AICR may not support

design/serving-stack-generation.md
haarchri: - We should support this.... as long as you pull the same versions we tested, just from your registry, nothing we care about breaks. Probably an `ImageConfig` rewrite for the whole stack instead of a repository field per component. We should follow up after this design

- Values like PDBs and security contexts: if our defaults are missing a strict non-root context or a PDB, that is something we should fix in our/aicrs defaults, not a missing override. Otherwise every team fixes the same thing on their own and we never find out... 

I did prototype `spec.overrides` with version/repository/values per key and it works, so it's not a `cannot`. We just do not want to open it before we know which requirements the defaults cannot cover.

```
spec:
  overrides:                      
  - key: kube-prometheus-stack    
    version: "72.6.2"             
    repository: https://mirror.example.com/helm-charts
    values: {...}                 # strategic-merged over the entry's values

```

design/serving-stack-generation.md
haarchri: i can write that down. If you gate serving stack updates through dev/test/prod, you run one Modelplane per environment, and that cost belongs in the doc...

Decoupling serving stack version from Modelplane version is a bigger change and gives up `the one reviewed combination per release `property.

design/serving-stack-generation.md
haarchri: thinking here is about a mode where Modelplane installs nothing and you bring your own serving stack... That covers a fully pre-provisioned cluster, So: we could define what Modelplane needs from a cert-manager, prometheus, lws etc. it didn't install. Curious what you think?

design/serving-stack-generation.md
haarchri: Yes, a model that needs a newer runtime, CUDA major, driver or GPU generation means a Modelplane release, not a field. That is what we pay for resolving everything at build time. Grove makes it worse because an unstable API two levels down can drive our release cadence.

Where the line is, and this should be in the doc?! day-zero for a model that runs on the runtime we already ship works as described. Day-zero for a model that needs a new runtime, driver or hardware is a release... 

Per-pool instead of per-cluster driver version is the first thing I would revisit (some cloud providers give you new images and you have a new driver), mixed generation fleets are what break it. Not solving it in this PR https://github.com/NVIDIA/aicr/issues/2516

haarchri:
@jonasz-lasut thanks for your review - short version of all points:

- mirrors: agreed, we need to think about to support it, likely ImageConfig-style
- security defaults: our job to fix / make requirements in AICR / other generators
- pre-provisioned BYO: `install nothing` mode is the likely answer, partial case still open...
- full s`pec.overrides` prototyped, holding it back for now, want concrete requirements first from users
- release cadence for new hardware/runtimes: real cost, overrides don't fix it, we need to document this ...

design/serving-stack-generation.md
jonasz-lasut: I think that a detailed instructions on what you need to provision if you opt-out of prometheus stack, cert manager and other really common components should be enough, good point

haarchri: is this something we would check with a "pre-hook" ?

negz (approved):

PR #432, opened 2026-09-08
Design: multiple accelerator vendors in the serving stack

haarchri:
The serving stack we propose in https://github.com/modelplaneai/modelplane/pull/422  is NVIDIA-only by construction: every clouds component list installs the NVIDIA operator or DRA driver, the GPU taint and toleration are `nvidia.com/gpu`, and nothing checks that the devices an `InferenceClass` declares match the stack its cluster gets. 
AMD Instinct GPUs are reaching the clouds Modelplane targets, and TPU and Trainium sit behind the same gap: a cluster may carry devices from more than one vendor, and the stack must install each vendors components exactly where that vendor's devices are.

The design treats the vendor as data that already exists: a class `devices[].driver` names it (`gpu.amd.com`, `gpu.nvidia.com`), the cluster composition derives the vendor set onto the `ServingStack`, and the component join filters vendor-tagged components to the vendors present. Mismatches fail early in the composition functions, on the class, on the class/cluster pairing, and in the join. The aicr-generated cloud halves are deliberately untouched until a second vendor is installable on a managed cloud.

Design only, no implementation. Feedback wanted on the tag semantics ("part of one vendors device stack", untagged always installs), the per-cloud vendor table and its mirrors in the two functions that need it, and on deferring the generate.py vendor classification.

Fixes #

I have: 

### DIFF
--- a/design/multi-accelerator.md
+++ b/design/multi-accelerator.md
@@ -1,0 +1,176 @@
+# Multiple accelerator vendors in the serving stack
+
+**Status:** Proposed
+**Date:** September 2026
+**Author:** Christopher Haar
+
+## Summary
+
+Modelplane's serving stack is NVIDIA-only by construction: every cloud's
+component list installs the NVIDIA operator or DRA driver, the GPU taint and
+toleration are `nvidia.com/gpu`, and nothing checks that the GPUs a class
+declares match the stack the cluster gets. That assumption is expiring on the
+clouds Modelplane already provisions: Vultr sells AMD Instinct GPUs alongside
+NVIDIA, and other accelerator families (TPU, Trainium) sit behind the same
+architectural gap. One cluster may carry devices from more than one vendor,
+and the serving stack must install each vendor's device stack exactly where
+that vendor's devices are.
+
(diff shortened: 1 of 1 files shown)


dennis-upbound (approved):
The vendor tag needs to cover the metrics exporter, not just the driver.

DCGM ships as an NVIDIA-stack component (`nvidia-dcgm`, per the allowlist in #456), so tagging it `NVIDIA` is correct — and it means an AMD-only cluster installs no GPU exporter at all. The telemetry design in #363 names DCGM as *the* source for temperature, throttling, ECC and interconnect errors, which is the fault taxonomy behind a drain. Those go silent on AMD, and nothing says so.

AMD's GPU Operator ships a Device Metrics Exporter with Prometheus-format temperature, utilization, memory and power, so the fix looks like the driver case:

```python
Chart(key="amd-device-metrics-exporter", accelerator_vendor="AMD", ...)
```

Worth a line in the design saying a vendor's stack includes its exporter. Keeping `modelplane_gpu_*` meaning the same thing whichever exporter produced it is #363's problem, not this one — I'll take that side.

PR #433, opened 2026-09-08
Docs: Move Hugo engine out of this repo

tr0njavolta:

This PR closes the loop on moving the hugo theme to [modelplaneai/docs-site.](https://github.com/modelplaneai/docs-site). The docs-site build clones one branch of this repo per docs version and mounts docs/content, docs/data, docs/manifests, and apis/ into the site. The site build, htmltest, DocSearch crawl move to the docs-site repo.

Vale, example manifest validation, apigroups.yaml, recipes.yaml stay in this repo. 

versions.yaml removed because modelplaneai/docs-site/blob/main/themes/geekboot/data/docversions.json controls the available/latest versions. 

.github/workflows/docs.yml is all that is left of the site here. A docs pull request dispatches to docs-site, which builds that revision alone at the deployment root and aliases it to modelplane-docs-pr-<number>.vercel.app - a hostname that follows from the number, so this repo can post the link the moment it dispatches. A merge dispatches again and rebuilds every version as production. The dispatch uses a GitHub App installed on docs-site rather than a personal access token, so it belongs to no one's account.

Fixes #

I have: 


nix/apps.nix
```diff
@@ -333,47 +333,51 @@
...
     };
 
-  # Serve the docs site locally with live reload. Extra args pass through to
-  # hugo server, e.g.: nix run .#docs-serve -- --port 8080
-  docsServe = _: {
+  # Preview the docs from this repo, with live reload. The Hugo project that
+  # renders them - config, layouts, theme, asset pipelines - is the docs-site
+  # repo; the prose is here, under docs/content. This clones it into docs/site
+  # (gitignored, and the same path vercel.json builds from) and points it at
+  # this working tree, so what you see is the file under your cursor.
```
negz: Is there some way we can keep all docs interactions one-way, from this repo -> docs-site? iidc We currently have a situation where docs-site clones this repo and this repo also clones docs-site.

I'm imagining something like:

* Only docs-site has a Vercel project (not this repo)
* Any utils for previewing or otherwise working with docs are defined in docs-site
* A PR workflow from this repo triggers a preview workflow/build in docs-site

For the utils, if you define them as a Nix flake in docs-site, this repo would be able to import and reference them as a convenience.

tr0njavolta:
docs-site clones read-only, nothing here clones docs-site. A PR touching `docs/` or `apis/` makes docs-site build that sha and publish a preview when the PR opens. A merge to main or a release branch makes it rebuild every version into production. Forks get previews too, via pull_request_target with no checkout

.The docs-site has a Vercel project now, vercel.json is deleted here. The preview utils are a flake there, so nix run github:modelplaneai/docs-site#preview from a checkout here serves main with live reload. A PR here triggers the build there through repository_dispatch.

Created a GH App with permissions to read and write to the docs-site repo for the sync.

dennis-upbound (approved):

Issue #418, opened 2026-08-28
A multi-worker gang isn't portable across serving stacks until Grove exposes a gang-wide pod index


negz:
### What problem are you facing?

A `Leader`/`Worker` engine with more than one worker needs each worker pod to get a distinct rank, and the two serving stacks added in #406 expose that differently. On `Standard` Modelplane injects `$(MODELPLANE_RANK)`, 0 on the leader and 1..N on the workers, so one worker template covers every pod. Grove numbers pods within a clique rather than across the gang, so on `Dynamo` the command has to do the arithmetic itself:

```yaml
# Standard
- --node-rank=$(MODELPLANE_RANK)
# Dynamo
- --node-rank=$$((GROVE_PCLQ_POD_INDEX + 1))
```

So the same `ModelDeployment` can't move between stacks unmodified, which is the portability that PR's design aims for. The leader address already behaves that way on both.

Nothing stops the mismatch, either. `compose-model-deployment` doesn't read `spec.stack`, so a deployment authored for `Standard` can be placed on a `Dynamo` cluster, where `$(MODELPLANE_RANK)` is never injected and reaches the engine as a literal string. A single-worker gang is fine, since you can hardcode 0 on the leader and 1 on the worker.

### How could Modelplane help solve your problem?

ai-dynamo/grove#755 exposes a PodCliqueScalingGroup-wide pod index. Once that's in a release the Grove backend can alias `MODELPLANE_RANK` to it the same way it already aliases `MODELPLANE_LEADER_ADDRESS` to the PCSG name and index, and the gang's coordination env reads identically on both stacks.

I'd treat it as a known shortcoming until then. A shim would need either parsing the user's engine args or arithmetic that env expansion can't do, and a scheduler constraint rejecting multi-worker gangs on `Dynamo` would trade a working configuration for a clearer error.

Issue #111, opened 2026-06-06
Add a Dynamo backend


negz:
### What problem are you facing?

#65 planned three serving paths: native Kubernetes (single-pod), llm-d (multi-pod), and NVIDIA Dynamo (NVIDIA-optimized multi-pod). #99 landed native and llm-d but left Dynamo a dormant stub. A ModelReplica can't dispatch to Dynamo today.

Implementing the backend needs a concrete rule for when Modelplane picks Dynamo over llm-d. The two projects overlap on a lot, so the rule should rest on where they actually differ. From their own docs:

Both are orchestration layers over the same engines (vLLM, SGLang, TensorRT-LLM). Both do disaggregated prefill/decode, KV-/prefix-aware routing, and SLO-aware autoscaling, and both can run behind a Gateway API Inference Extension (GAIE) gateway. They also share plumbing: llm-d's disaggregation transfers KV over [NIXL](https://github.com/ai-dynamo/nixl), which is an NVIDIA/Dynamo-org library, and llm-d lists Dynamo's KVBM as a compatible out-of-tree KV connector.

Where Dynamo has capabilities llm-d doesn't ship today:

- **Grove** ([ai-dynamo/grove](https://github.com/ai-dynamo/grove)) — a Kubernetes scheduler for hierarchical gang scheduling, topology-aware placement within NVLink domains, and explicit startup ordering, with the pod group (not the pod) as the scaling unit. llm-d's multi-pod path is built on LeaderWorkerSet and has no equivalent gang scheduler or topology awareness.
- **ModelExpress** ([ai-dynamo/modelexpress](https://github.com/ai-dynamo/modelexpress)) — GPU-to-GPU weight streaming over NIXL/RDMA, so a new replica receives weights peer-to-peer from a seed GPU instead of loading from storage. llm-d has no equivalent; replicas load weights from storage on startup.
- **KVBM** ([docs](https://docs.nvidia.com/dynamo/latest/kvbm/)) — a built, unified GPU→CPU→SSD→remote-storage tiered KV hierarchy. llm-d's own native hierarchical offload is "under active development," with its CPU and storage tiers currently operating as independent options rather than one hierarchy ([llm-d KV offloading](https://llm-d.ai/docs/architecture/advanced/kv-management/kv-offloader)).

[...]

PR #448, opened 2026-09-10
Document the per-cluster serving stack choice

negz:

#406 added the Standard and Dynamo serving stacks but documented them only in passing: a scattering of asides that each restated the `spec.stack: Dynamo` opt-in, and nothing on the platform page where you actually register a cluster. This gives the per-cluster choice a home and points the asides at it.

A new Serving stack section on the cluster page explains Standard versus Dynamo and what each installs. The multi-node deployment docs gain a note on the two things that differ under Dynamo: a gang derives its rank from Grove's pod index because `MODELPLANE_RANK` isn't injected there yet (#418), and an engine opts into ModelExpress peer-to-peer loading. The glossary gains an entry, and the how-it-works, architecture, model-cache, and FAQ mentions now link the section instead of re-deriving the opt-in.

It also corrects the FAQ, which claimed Modelplane uses llm-d for multi-node serving. The stacks own multi-node now, a LeaderWorkerSet on Standard or a Grove PodCliqueSet on Dynamo; llm-d is the inference-aware routing layer.

A Dynamo how-to guide is deliberately left out, to follow up once it's validated end to end on real hardware.

Towards #111.

I have:

### DIFF
--- a/docs/content/architecture/_index.md
+++ b/docs/content/architecture/_index.md
@@ -129,9 +129,9 @@ provider resources and plain Kubernetes objects. A `ModelDeployment` is the
 clearest example. Its function schedules the replicas, then composes a
 `ModelReplica` for each, and a `ModelEndpoint` for each replica that's ready to
 serve. Each `ModelReplica` function composes the serving workload onto its target
-workload cluster through provider-kubernetes: a Deployment for single-node, or a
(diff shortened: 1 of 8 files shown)


haarchri (approved):
would be great to have some Recipes available too - fine in a followup

dennis-upbound (approved):

PR #452, opened 2026-09-15
Docs: Drop the per-cloud storage list from the ModelCache page

### LINKED ISSUE #439: docs: the cache storage list in model-cache.md reads as exhaustive but omits AKS and Nebius
### What happened?

`docs/content/models/model-cache.md` ("Storage prerequisites") lists two auto-provisioned sources:

> The cache PVC needs a `ReadWriteMany` (RWX) StorageClass on the workload cluster. What the platform admin must set up depends on the cloud:
>
> - **GKE** and **EKS:** auto-provisioned. Nothing for the admin to do.
> - **Existing:** the admin sets up a `ReadWriteMany` StorageClass on the cluster.

`docs/content/platform/inference-cluster.md` ("Cache storage") lists four, plus the Vultr caveat:

> - **`GKE`** (Filestore Enterprise), **`EKS`** (EFS), **`AKS`** (Azure Files), and **`Nebius`** (shared filesystem): auto-provisioned. Those classes are fixed; nothing for the admin to do.
> - **`Vultr`**: none. VKE's built-in RWX class (Vultr File System) isn't usable on GPU nodes [...]

To be fair to the page: `model-cache.md` does link onward, immediately after its list, with "How storage is provided on each cluster source, and how to bring your own backend, is covered in Register a Cluster". So this is a summary pointing at the canonical list, not a contradiction, and I nearly didn't file it.

The reason I think it's still worth a small fix: the short list is written as an enumeration rather than as an example, so it reads as exhaustive. Someone landing on the `ModelCache` page to answer "can I use a cache on AKS?" gets a list of supported clouds that doesn't include AKS, and the answer is yes. The Vultr caveat, which is the one case where the answer is genuinely no, is also only on the other page.

Suggestion, since you already link to the canonical section: drop the per-cloud list from `model-cache.md` entirely and keep the requirement plus the link. That leaves one source of truth and nothing to drift.

Happy to send that PR if you want it. Closing this as working-as-intended is a perfectly reasonable outcome too.

### How can we reproduce it?

Read the two sections side by side:

- https://github.com/modelplaneai/modelplane/blob/main/docs/content/models/model-cache.md (Storage prerequisites)
- https://github.com/modelplaneai/modelplane/blob/main/docs/content/platform/inference-cluster.md (Cache storage)

Docs only, `main` as of 2026-09-09.

haarchri:

Drop the per-cloud storage list from the ModelCache page

The Storage prerequisites section on the ModelCache page listed which cluster sources auto-provision an RWX StorageClass. The list named only GKE, EKS, and Existing, but Register a Cluster carries the canonical list, which has since grown AKS, Nebius, and the Vultr caveat. The short list reads as exhaustive, so a reader checking whether a cache works on AKS got the wrong answer, and the one source where a cache genuinely doesn't work (Vultr) wasn't mentioned at all.

Rather than sync the list, drop it. The page keeps the RWX requirement and links to Register a Cluster for the per-source details, leaving one list to maintain and nothing to drift.

Fixes #439

I have: 


tr0njavolta (approved):

dennis-upbound (approved):
Verified the anchor still resolves — `## Cache storage` is at inference-cluster.md L75. Right fix: the short list read as exhaustive and had already drifted past GKE/EKS/Existing.

PR #454, opened 2026-09-15
Allow scaling a ModelDeployment to zero replicas

### LINKED ISSUE #438: spec.replicas is bounded (minimum 1, maximum 10) but neither bound is documented
### What happened?

`docs/content/models/model-deployment.md` (lines 222-226) says:

> `spec.replicas` is the only scaling axis. Each replica is a complete, fixed-shape serving instance, so scaling adds or removes whole instances across the fleet. Because the deployment exposes the Kubernetes scale subresource, `kubectl scale` and KEDA work without anything extra. There's no in-cluster pod autoscaling.

`apis/modeldeployments/definition.yaml` (lines 41-44) bounds the field at both ends:

```yaml
replicas:
  type: integer
  minimum: 1
  maximum: 10
```

So scaling to zero is rejected:

```
$ kubectl -n ml-team scale modeldeployment mock-demo-v2 --replicas=0
The ModelDeployment "mock-demo-v2" is invalid: spec.replicas:
  Invalid value: 0: spec.replicas in body should be greater than or equal to 1
```

Neither bound appears in the docs. "Sizing a deployment" and "Scaling" in `model-deployment.md` both describe `spec.replicas` without mentioning a range, and `getting-started/scale-the-model.md` doesn't either.

Two separate surprises, and the ceiling is the bigger one:

- **`maximum: 10`** caps a deployment at ten replicas fleet-wide. For a project whose premise is spreading a model across many GPU clusters, that ceiling seems worth stating in the docs, or raising if it's just a placeholder guard.
- **`minimum: 1`** means scale-to-zero isn't reachable. Scale-to-zero is one of the main reasons people put KEDA in front of an expensive GPU workload, so naming KEDA and the scale subresource in the same paragraph sets an expectation the API doesn't meet.

I don't know whether either bound is deliberate. If they are, a sentence next to the KEDA mention would save people the discovery. If the minimum isn't, dropping it to 0 would make a parked-but-defined deployment expressible.

For context on why zero is useful: I wanted to withdraw one deployment's endpoints while leaving the object in place, to observe how a weighted `ModelService` behaves when an entry matches no healthy endpoints. I had to taint the hosting cluster instead, which is a heavier instrument and also moves replicas.

Note this is about the `ModelDeployment` replica count, not GPU node pools. #345 covers pools not scaling to zero on clouds using the vanilla cluster-autoscaler, which is a separate layer.

### How can we reproduce it?

1. Apply any `ModelDeployment`.
2. `kubectl -n <ns> scale modeldeployment <name> --replicas=0`
3. The API server rejects it with the message above.

Modelplane version: v0.3.1 (`xpkg.upbound.io/modelplane/modelplane:v0.3.1`)
Crossplane: 2.4.0
Kubernetes: kind, `kindest/node:v1.34.0`
Inference clusters: `source: Existing` (BYO, registered by kubeconfig)

haarchri:

Allow scaling a ModelDeployment to zero replicas

`spec.replicas` carried a minimum of 1, so `kubectl scale --replicas=0` was rejected at admission. The docs name KEDA and the scale subresource in the same paragraph, and scale-to-zero is a main reason to put KEDA in front of an expensive GPU workload, so the floor broke an expectation the API itself set. There was also no way to park a deployment: withdrawing its endpoints while keeping the object in place meant tainting the hosting cluster.

Dropping the floor alone would misreport the parked state. With zero desired, an empty schedule reads as `InsufficientCapacity` 0 of 0 replicas scheduled, and a control plane with no clusters reads as
NoClusters, both leaving the XR not ready, so a deliberately parked deployment would look permanently broken.

Zero desired now takes a dedicated path before input resolution: nothing is composed so the existing `ModelReplicas` and `ModelEndpoints` are pruned, `status.replicas` reports 0 for the scale subresource, and `ReplicasScheduled` and `ReplicasReady` read True with a new `ScaledToZero` reason, the way a Deployment at zero replicas reports Available. The docs' Scaling section now states the 0 to 10 range and the parked semantics.

Fixes #438

I have: 


dennis-upbound (approved):
Does parking actually release the GPUs?

The getting-started cast says `minNodeCount: 1   # keep >=1; the autoscaler can't scale a GPU pool up from 0 for DRA pods`. So on a pool sized that way, parking removes the pods but the floor keeps the nodes, and you keep paying for them. Set `minNodeCount: 0` and the pool can empty, but then nothing brings it back for DRA pods — parking becomes one-way.

Savings are real in between (8 nodes down to 1), so this isn't an objection to the feature. But the docs say "a KEDA scaler can idle an expensive deployment and later restore it", and restore only holds while the floor is 1 or more. Worth saying which of the two it is.

PR #456, opened 2026-09-15
Stop serving stack pods tolerating every taint

haarchri:

aicr's bundler stamps a wildcard toleration `{operator: Exists}` on every pod it renders, Deployments included. On the generated clouds that put cert-manager, the Prometheus stack and the rest on tainted GPU nodes, where they squat on accelerated capacity and their eviction delays autoscaler scale-down.

The generator now rewrites every wildcard the bundle ships: node agents that must reach GPU nodes tolerate exactly the GPU taint the cluster compositions apply, everything else tolerates nothing and schedules on the untainted system pool. Like ALLOW/DROP the tables fail closed, so an aicr bump can't put a new control-plane pod back on GPU nodes. A new unit test pins every cloud, chart values and manifests alike, to keyed tolerations.

The hand-written clouds never stated a wildcard, but one shipped at runtime anyway: the node-exporter chart's default toleration is keyless, and Nebius, Vultr and Existing didn't override it. The unit test can only see stated values, so catching this class needs a rendered cluster. Those three halves now scope node-exporter to the GPU taint, mirroring the generated clouds.

`nix run .#e2e -- --cloud eks|aks|gke|nebius|vultr|existing` installs that cloud's serving stack on the local kind workload cluster, with KWOK faking two nodes labelled and tainted exactly like that cloud's GPU pool while the real kind node plays the untainted system pool. A single values-driven Chainsaw test applies the fake nodes and a ServingStack XR directly (`source: Existing` hard-wires `cloud: Existing`, so the XR is applied rather than composed through an InferenceCluster), waits for it Ready, then checks placement from both sides: the values file's expected DaemonSets must fan out to the fake GPU pool (a toleration too narrow fails here e.g. AKS's `Equal` toleration with a drifted value), and an audit of every pod's tolerations rejects wildcards and unexpected GPU tolerations wherever the pod landed (a toleration too broad fails here, approximating client-side what AKS rejects at admission). Everything cloud-specific lives in `e2e/clouds/values-<cloud>.yaml`.

Validated with a local `--cloud aks`

```
PASS
Tests Summary...
- Passed  tests 1
- Failed  tests 0
- Skipped tests 0
Done.

==> Cloud serving stack OK: the aks stack installed and placed correctly 

```

Fixes #

I have: 


dennis-upbound (approved):
The unanchored patterns are the one fail-open path here.

A renamed upstream workload stops matching the allowlist and the audit goes red, which is the safe direction. But `k8s-ephemeral-storage-metrics` and `kubelet-plugin` match anywhere in a pod name, so a future workload that happens to contain either substring gets a wildcard toleration allowed silently — the exact regression this PR exists to catch.

Anchor the ones that can be? The comment says they're unanchored because the generated and hand-written halves name the same workload differently, so if that's only true for some of them, the rest could take `^`.

